PR #4, opened 2024-02-28
zarrify.py script update

sethmcg:
This is a pretty trivial pull request, but I've never done one before, so I figure it's a good way to test the workflow.

I added a shbang and a couple checks for input file existence / output file nonexistence, and updated the docstring.

### DIFF
--- a/scripts/zarrify.py
+++ b/scripts/zarrify.py
@@ -21,6 +21,7 @@ import glob
 from argparse import ArgumentParser
 
 
+
 parser = ArgumentParser(
     description="converts a directory of netcdf files into a zarr store"
 )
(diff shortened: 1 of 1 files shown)


scripts/zarrify.py
```diff
@@ -1,29 +1,53 @@
...
+# - os.path
+# - glob
+
 """
 import xarray
 import sys
+import os.path
+import glob
 
 if(len(sys.argv) != 3):
```
djgagne: Are you familiar with the argparse package? It handles a lot of processing and formatting of command line arguments in a slick way. See https://github.com/ai2es/miles-guess/blob/main/applications/train_mlp_SL.py for an exmaple.

sethmcg: Can do.  I wasn't sure whether the script was of sufficient general interest to warrant full-on argument handling.

I'm not familiar with argparse, but it looks easy enough.

sethmcg: @djgagne - PR updated

PR #22, opened 2024-03-14
Conus404 data loader

sethmcg:
I have a functioning data loader for the CONUS404 dataset!  It can load different subsets of the (currently-zarrified) variables.  It also can load a tiny test dataset I created from a different source, so I think it could probably be modified to load the ERA5 dataset as well if we want to unify the code.


credit/data.py
```diff
@@ -341,6 +345,154 @@ def __getitem__(self, index):
...
+        ## CONUS404 data is organized into directories by variable,
+        ## with a set of yearly zarr stores for each variable
+        if len(self.varnames) == 0:
+            self.varnames = os.listdir(self.zarrpath)
+        self.varnames = sorted(self.varnames)
+        
+        ## get file paths
+        zdict = {}
+        for v in self.varnames:
+            zdict[v] = sorted(glob(self.zarrpath+"/"+v+"/"+v+".*.zarr"))
```
djgagne: I suggest using `os.path.join` instead of manually adding slashes. It is more robust.

credit/data.py
```diff
@@ -341,6 +345,154 @@ def __getitem__(self, index):
...
+    corresponding zarr store and read only the data we want with an
+    isel() call.
+
+    For multiple variables, we necessariy end up reading from multiple
+    stores, but they're merged into a single xarray Dataset, so
+    hopefully that won't cause a big performance hit.
+
+    """
+    
+    zarrpath:     str = "/glade/campaign/ral/risc/DATA/conus404/zarr"
```
djgagne, line 395: Currently these are being defined as class variables, which means they will be the same for every instance of the class. They should be defined and set within a `__init__` method so that things like the seed and path to the data can be changed as needed.

credit/data.py
```diff
@@ -341,6 +345,154 @@ def __getitem__(self, index):
...
+        first = self.zindex[index]
+        last = first + self.sample_len
+        seg = self.segments[index]
+        subset = self.zarrs[seg].isel({time:slice(first, last)}).load()
+        sample=Sample(history=subset.isel({time:self.histmask}),
+                      target=subset.isel({time:self.foremask}),
+                      datetime=subset[time])
+        
+        if self.transform:
+            sample = self.transform(sample)
```
djgagne, line 475: Where is the transform method defined?

credit/data.py
sethmcg: Done!

credit/data.py
sethmcg: It's got the @dataclass decorator, which autoconstructs an init() method that takes those variables as args and sets them as instance variables.

credit/data.py
sethmcg: It's an optional argument to the constructor created by the dataclass decorator.  It'll be something like `Compose([NormalizeState(...), ToTensor(...)])`

PR #23, opened 2024-03-16
polar and laplacian diffusion filter module class

WillyChap:
This is a module to do diffusion/filtering operators into the model time step. 
This is done in the spectral harmonics space and will need some tuning of substeps ... 

Right now it works well, but might be a little agressive to do to the Q field. 

See: 
https://journals.sagepub.com/doi/full/10.1177/1094342011410088?casa_token=XU77aTGMkjEAAAAA%3AfVKn2V6INlocoCObdb-tyIByO6uYQ6hbn5cZCCvEusw88fJknGsX1aiK-jm_OuRb1sozzwX89Y2h5A 

for mathematical derivation

### DIFF
--- a/credit/pol_lapdiff_filt.py
+++ b/credit/pol_lapdiff_filt.py
@@ -222,14 +222,14 @@ class Diffusion_and_Pole_Filter:
         divspec2 = (self.lap*chispec)
     
         if idim == 1:
-            uchi, vchi = self.getuv(torch.stack((torch.zeros([divspec2.shape[0],divspec2.shape[1]]).to(device),divspec2)))
+            uchi, vchi = self.getuv(torch.stack((torch.zeros([divspec2.shape[0],divspec2.shape[1]]).to(self.device),divspec2)))
             return torch.squeeze(uchi), torch.squeeze(vchi)
         elif idim == 2:
-            uchi, vchi = self.getuv(torch.stack((torch.zeros([divspec2.shape[0],divspec2.shape[1]]).to(device),divspec2)))
+            uchi, vchi = self.getuv(torch.stack((torch.zeros([divspec2.shape[0],divspec2.shape[1]]).to(self.device),divspec2)))
             return uchi, vchi
         elif idim == 3:
             new_shape = (divspec2.shape[0], 2, *divspec2.shape[1:])
-            stacked_divspec = torch.zeros(new_shape,dtype=torch.complex64, device=device)
+            stacked_divspec = torch.zeros(new_shape,dtype=torch.complex64, device=self.device)
             # Copy the original data into the second slice of the new dimension
             stacked_divspec[:, 1, :, :] = divspec2
             backy = self.getuv(stacked_divspec)
--- a/credit/pol_lapdiff_filt.py
+++ b/credit/pol_lapdiff_filt.py
@@ -1,0 +1,426 @@
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.fft as fft
+# from scipy.ndimage import convolve
+#import windspharm
+#from windspharm.standard import VectorWind
+from torch_harmonics import *
+import torch_harmonics as harmonics
+import copy
+import gc
+
+def polfiltT(D,inddo):
+
+    if len(D.shape)==2:
+        for ii in torch.concatenate([torch.arange(-inddo,0),torch.arange(1,inddo+1)]):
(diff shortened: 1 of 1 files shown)


credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+    U = BB2[8,:,:]
+    V = BB2[8,:,:]
+    nlat = U.shape[0]
+    nlon = U.shape[1]
+    BB2_tensor=torch.tensor(BB2).clone().detach().to(device)
+    #instantiae this with the original model:
+    DPF = Diffusion_and_Pole_Filter(nlat=nlat,nlon=nlon,device=device)
+
+    #this fixes the divergent modes ... might eb doing too much to Q
+    BBfix=DPF.diff_lap2d_filt(BB2_tensor)
```
jsschreck, line 420: Just to confirm, the output of BBfix is still the same shape as BB1 and BB2 here, right? I clone the tensor initially and do not perform any correction on Z500

credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+        spectral domain operations.
+    
+        Parameters:
+        - T (Tensor): scalar-component of the velocity or vector field.
+        - ind_pol (list/int): Index/indices specifying poles for the filtering process.
+    
+        Returns:
+        - Tuple of Tensors: The modified T components of the scalar field.
+        """
+        T=T.clone()
```
jsschreck: I think the first clone is redundant?

credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+    
+    # Calculate the ramp down using a reversed sigmoid function
+    x_down = torch.linspace(-6, 6, ramp_length)
+    sigmoid_down = 1 / (1 + torch.exp(-x_down))
+    array[ramp_down_start:] = torch.flip(sigmoid_down, dims=(0,))
+    
+    # Return the modified array
+    return array
+
+def find_nearest(array, value):
```
jsschreck, line 93: Looks like some extra indentation here?

credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+    Creates an array of specified length with a sigmoid ramp up and down.
+    
+    Parameters:
+    - array_length: The length of the output array.
+    - ramp_length: The length of the ramp up and down.
+    
+    Returns:
+    - An array of the specified length with the described ramp up and down using a sigmoid function.
+    """
+    import numpy as np
```
jsschreck: Numpy imported already at the top.

credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+            var_spectrum = np.real(2 * s)  #multiply by two in order to conserve variance
+            A = Ck2/torch.sum(Ck2)
+            A[ind_1d:] = 0.
+            Zlow = torch.clone(Z)
+            Zlow[ind_1d:-ind_1d] = 0.0
+            X_filtered11 = torch.real(torch.fft.ifft(Zlow))
+            D[ii,:]=X_filtered11
+        return D
+
+    if len(D.shape)==3:
```
jsschreck, line 37: You could combine the two loops here into one, that might help reduce some of the code duplication (you could also put the stuff you use twice here into its own method).

jsschreck (approved):

credit/pol_lapdiff_filt.py
WillyChap: yeah that is the same shape going in. No correction to Z500 looks to be needed.

credit/pol_lapdiff_filt.py
```diff
@@ -0,0 +1,426 @@
...
+        if len(chispec.shape) == 1:
+            nt = 1
+            chispec = torch.reshape(chispec, ((ntrunc+1)*(ntrunc+2)//2,1))
+        else:
+            nt = chispec.shape[1]
+    
+        divspec2 = (self.lap*chispec)
+    
+        if idim == 1:
+            uchi, vchi = self.getuv(torch.stack((torch.zeros([divspec2.shape[0],divspec2.shape[1]]).to(device),divspec2)))
```
jsschreck: All the "device" calls in this method should be "self.device"

Issue #24, opened 2024-03-16
Q needs the quantizer scaling


WillyChap:
Q is pretty bad even after 1 time step (see figure below). It's my belief that this is the major contributor to the grid artifacts. (look at the pole where there is minimum moisture available. The loss is CLEARLY focusing on the tropics this is largely a result of the vast difference in scale between tropical and polar specific humidity. Having a single scalar Z score per level will exasperate this issue. I think we need to implement the quantizer scaling ASAP. 

This image shows the 2nd auto-recursive time step of the Q field, before and after the diffusion filtering. You can see that the grid artifact is already present, and the diffusion filtering has to do ALOT in order to minimize these errors. It will cause a lot of smoothing unless we can fix it. 

panel 1 (difference before and after filtering)
panel 2 (Q prior to filter) 
panel 3 (Q post filter)

<img width="1503" alt="image" src="https://github.com/NCAR/miles-credit/assets/14932329/d807aad3-2acd-426e-a7d6-401b5dac4c60">

PR #32, opened 2024-03-22
Quantile static

WillyChap:
This pull request provides access to the quantile transform scaler (via the bridgescaler package), It returns the exact tensors structure as our current scaling, so should cause no pipeline issues. I am optimistic it will help training alot. 

Here is the use: 

```
Dataset= ERA5Dataset(filenames=[FNS[nunu]],history_len=history_len,forecast_len=forecast_len,skip_periods=1,transform=transforms.Compose([
            NormalizeState_Quantile(scaler_file=conf["data"]["quant_path"]),
            ToTensor(history_len=history_len, forecast_len=forecast_len,static_variables=conf["data"]["static_vars"]),
        ]))
print(FNS[nunu])
BB_trancs_quant = Dataset.__getitem__(8784)

Dataset= ERA5Dataset(filenames=[FNS[nunu]],history_len=history_len,forecast_len=forecast_len,skip_periods=1,transform=transforms.Compose([
            NormalizeState(conf["data"]["mean_path"], conf["data"]["std_path"]),
            ToTensor(history_len=history_len, forecast_len=forecast_len,static_variables=conf["data"]["static_vars"]),
        ]))
print(FNS[nunu])
BB_trancs_std = Dataset.__getitem__(8784)
```
I am requesting two variables added to the data section of the config file, though i don't think it is static (I am happy to adjust once I know what needs to be added). They are currently integrated into crossformer.yml

The adjustment of the quantile scalar is apparent. Here is the difference between the Q at upper levels for a standard scaling (bottom) vs a quantile scaling (top).

image
 

Additionally, I have added a 'static variables' option to the to_tensor transform. This will now return a feild 'static' which provides the LandSea mask  (scaled 0-1)  and the topography (scaled 0-2). See Below: 

image

This addresses #24

### DIFF
--- a/credit/transforms.py
+++ b/credit/transforms.py
@@ -5,6 +5,8 @@ import xarray as xr
 import netCDF4 as nc
 from credit.data import Sample
 from typing import Dict
+import pandas as pd
+from bridgescaler import read_scaler
 
 
 logger = logging.getLogger(__name__)
--- a/credit/transforms.py
+++ b/credit/transforms.py
@@ -243,7 +243,7 @@ class ToTensor:
         self.for_len = int(conf["data"]["forecast_len"])
         self.variables = conf["data"]["variables"]
         self.surface_variables = conf["data"]["surface_variables"]
-        self.allvars = variables + surface_variables
+        self.allvars = self.variables + self.surface_variables
         self.static_variables = conf["data"]["static_variables"]
 
     def __call__(self, sample: Sample) -> Sample:
--- a/credit/transforms.py
+++ b/credit/transforms.py
@@ -12,7 +12,9 @@ from torchvision import transforms as tforms
 
 logger = logging.getLogger(__name__)
 
+
 def load_transforms(conf):
+
     if conf["data"]["scaler_type"] == 'quantile':
         transform_scaler = NormalizeState_Quantile(scaler_file=conf["data"]["quant_path"])
     elif conf["data"]["scaler_type"] == 'std':
@@ -130,7 +132,7 @@ class NormalizeState_Quantile:
         self.scaler_surf = self.scaler_surfs.sum()
         self.variables = variables 
         self.surface_variables = surface_variables 
-        self.levels = 15
+        self.levels = levels
 
     def __call__(self, sample: Sample, inverse: bool = False) -> Sample:
         if inverse:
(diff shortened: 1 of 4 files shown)


credit/transforms.py
```diff
@@ -92,6 +92,90 @@ def inverse_transform(self, x: torch.Tensor) -> torch.Tensor:
...
+
+    def __call__(self, sample: Sample, inverse: bool = False) -> Sample:
+        if inverse:
+            return self.inverse_transform(sample)
+        else:
+            return self.transform(sample)
+
+    def inverse_transform(self, x: torch.Tensor) -> torch.Tensor:
+        device = x.device
+        tensor = x[:, :(len(self.variables)*self.levels), :, :]  #B, Var, H, W
```
djgagne, line 135: I would store len(self.variables) as a variable in init somewhere. That should save you some time, especially given how many time it is called here.

credit/transforms.py
```diff
@@ -92,6 +92,90 @@ def inverse_transform(self, x: torch.Tensor) -> torch.Tensor:
...
+        surface_tensor = x[:, (len(self.variables)*self.levels):, :, :]  #B, Var, H, W
+        #beep
+        # Reverse quantile transform using bridge scaler:
+        transformed_tensor = tensor.clone()
+        transformed_surface_tensor = surface_tensor.clone()
+        #3dvars
+        rscal_3d = np.transpose(torch.Tensor.numpy(x[:,:(len(self.variables)*self.levels),:,:].values),(0, 2, 3, 1))
+        self.scaler_3d.inverse_transform(rscal_3d)
```
djgagne, line 143: There's a channels_last flag in the bridgescaler distributed transform and inverse_transform methods that can be set to False to do the transform in channels_first order even if the scaler was trained on channels_last data. You shouldn't have to reshape to channels_last anymore.

djgagne (changes requested):
I think some big speedups could be made if you use the channels_last=False transform functionality that I just added to bridgescaler this week. You shouldn't have to retrain the current scalers just yet.

dkimpara (approved):
quantile + static vars integrated and tested in train.py. pending speedups to quantile scaler in bridgescaler

@yingkaisha please see the new transforms and integrate into predict.py let me know if you'd like any help. Not sure if inverse transform for quantile exists yet. might need to do that in another PR

jsschreck (approved):

PR #37, opened 2024-03-30
TOA var and Quantile scaling, 

WillyChap:
This PR adds the TOA predictor variable. It comes in as another returned item to the to_tensor dict (return_dict['TOA']). Additionally it adds the new quantile scaling and the dictionary. 

I tried to speed it up... but to no end. Please adjust @djgagne if you see anything to boost performance. I know we don't have to do the transform 1 time chunk at a time with the bridge scaler, but changing that offers almost no speedup, and the concat is fast. 

Additionally I've added a transforms notebook for rapid testing/development of the transforms 

image


credit/transforms.py
```diff
@@ -139,13 +142,11 @@ def inverse_transform(self, x: torch.Tensor) -> torch.Tensor:
         transformed_tensor = tensor.clone()
         transformed_surface_tensor = surface_tensor.clone()
         #3dvars
-        rscal_3d = np.transpose(torch.Tensor.numpy(x[:,:(len(self.variables)*self.levels),:,:].values),(0, 2, 3, 1))
-        self.scaler_3d.inverse_transform(rscal_3d)
-        tensor[:, :, :, :] = torch.tensor(np.transpose(rscal_3d), (0, 3, 1, 2)).to(device)
+        rscal_3d = (np.array(x[:, :(len(self.variables)*self.levels), :, :]))
```
djgagne, line 145: Instead doing `np.array(x[...])` do `x.numpy()`, which returns a view instead of a copy.

credit/transforms.py
```diff
@@ -139,13 +142,11 @@ def inverse_transform(self, x: torch.Tensor) -> torch.Tensor:
...
+        transformed_tensor[:, :, :, :] = torch.tensor((self.scaler_3d.inverse_transform(rscal_3d))).to(device)
```
djgagne, line 146: If you want to try doing the transform in parallel, there is a `pool=` kwarg for QuantileScaler inverse_transform and transform that uses a multiprocessing Pool to perform parallel transforms across variables. However, I have found that it doesn't necessarily play nicely with MPI, which is being used under the hood for distributed training.

credit/transforms.py
```diff
@@ -167,20 +168,20 @@ def transform(self, sample: Dict[str, np.ndarray]) -> Dict[str, np.ndarray]:
...
-                    value['V'].sel(time=time)[:,:,:]=np.transpose(TTtrans[:,:,self.levels:(self.levels*2)].values, (2, 0, 1))
-                    value['T'].sel(time=time)[:,:,:]=np.transpose(TTtrans[:,:,(self.levels*2):(self.levels*3)].values, (2, 0, 1))
-                    value['Q'].sel(time=time)[:,:,:]=np.transpose(TTtrans[:,:,(self.levels*3):(self.levels*4)].values, (2, 0, 1))
-                    e_surf = xr.concat([value[v].sel(time=time) for v in self.surface_variables], pd.Index(self.surface_variables, name="variable")
-                                   ).transpose("latitude", "longitude", "variable")
+                    e3d = xr.concat(var_slices, pd.Index(var_levels, name="variable"))
+                    e3d = e3d.expand_dims(dim="time", axis=0)
+                    TTtrans = self.scaler_3d.transform(np.array(e3d))
+                    #this is bad and should be fixed:         
+                    value['U'].sel(time=time)[:,:,:]=(TTtrans[:,:self.levels,:,:].squeeze())
```
djgagne, line 175: .sel is generally much slower than .loc.

config/crossformer.yml
```diff
@@ -4,11 +4,12 @@ seed: 1000
 data:
     variables: ['U','V','T','Q']
     surface_variables: ['SP','t2m','V500','U500','T500','Z500','Q500']
-    static_variables: ['SP','LSM'] # []
+    static_variables: ['Z_GDS4_SFC','LSM','tsi'] # []
```
djgagne, line 7: John's staging on scratch is missing a few years of ERA5 data compared with Will's directory on campaign.

djgagne (changes requested):
The best way to speed up the quantile transform process for training is to pre-scale all ERA5 data, which should be much more straightforward to do. Another option would be to reserve some MPI ranks for scaling purposes and add MPI commands to send tensor data to those ranks for transforms. I could potentially store a scaled version of ERA5 but would need to expand my scratch quota first.

config/crossformer.yml
jsschreck: Yeah Will had to redo a few years after the fact; I'll copy the missing ones over now.

jsschreck (approved):

PR #38, opened 2024-04-12
Predict.py now supports FSDP

jsschreck:
The main update is enabling FSDP with predict.py

Other changes:

Added forecast.py utility
u-net deprecation on checkpointing (load_model does not work with Base)


WillyChap (approved):
This is great. thanks for the updates

PR #53, opened 2024-06-25
Added Swin transformer model + new rollout script

jsschreck:


credit/forecast.py
```diff
@@ -46,13 +48,12 @@ def generate_forecasts(start_date, days=10, duration=365):
     # start_date = datetime(2020, 1, 1)
     # forecasts = generate_forecasts(start_date)
 
-    config_file = "/glade/u/home/schreck/schreck/repos/global/miles-credit/results/crossformer/quarter/multi_step/model.yml"
+    config_file = "/glade/derecho/scratch/schreck/repos/miles-credit/results/wxformer/quarter/zscore_multi/model.yml"
```
WillyChap: Should this be hard coded?

yingkaisha: It doesn't matter, we will do

```python
from credit.forecast import load_forecasts
```

yingkaisha (approved):
The rollout part looks good. I haven't tested the swin model yet, if there's any problem we can fix that later.

credit/forecast.py
jsschreck: I can update it so that it will use a hard-coded dictionary to be able to run that example. The intention here is that we eventually build unit tests around whats in if name == main

PR #57, opened 2024-07-05
add 6hourly cached option

WillyChap:
This PR just adds the ability to run the 6 hourly cached and non-cached datasets. Everything should be smooth, just had to add options for the caching. I've include a new yml which should work with the file.

### DIFF
--- a/config/crossformer_6hrCached.yml
+++ b/config/crossformer_6hrCached.yml
@@ -5,12 +5,12 @@ data:
     variables: ['U','V','T','Q']
     surface_variables: ['SP','t2m','V500','U500','T500','Z500','Q500','Z_GDS4_SFC','LSM','tsi']
     static_variables: []
-    save_loc: '/glade/derecho/scratch/dgagne/STAGING/TOTAL_*'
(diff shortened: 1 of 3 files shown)


config/crossformer_6hrCached.yml
```diff
@@ -0,0 +1,202 @@
...
+  
+model:
+    type: "crossformer"
+    image_height: 640    # Image height (default: 640)
+    image_width: 1280    # Image width (default: 1280)
+    levels: 15           # Number of levels (default: 15)
+    frames: 2            # Number of frames (default: 2)
+    channels: 4          # Channels (default: 4)
+    surface_channels: 10  # Surface channels (default: 7)
+    static_channels: 3   # Static channels (default: 3)
```
yingkaisha, line 60: Double-check the number of channels.

yingkaisha (approved):
Fixed a few things and leave the (new) cached.yml as it is.

config/crossformer_6hrCached.yml
yingkaisha: will leave this part for future testing

PR #65, opened 2024-07-26
Initialization of credit.trainers

jsschreck:


yingkaisha (approved):
Looks good to me. We updated the CONUS 404 trainer, so we need to talk to them about the changes.

PR #66, opened 2024-07-27
new lat weighting

WillyChap:
fix latitude weighting in the loss


jsschreck (commented):
@WillyChap can you add the latitude_weights method call to the VariableTotalLoss2D? That one also just loads it from the file manually but it would be better to use the method.

credit/loss.py
```diff
@@ -169,10 +169,13 @@ def get_psd(self, f_x, device, dtype):
...
-    cos_lat = xr.open_dataset(conf["loss"]["latitude_weights"])["coslat"].values
-    # Normalize over lat
-    cos_lat_sum = cos_lat.sum(axis=0) / cos_lat.shape[0]
-    L = cos_lat / cos_lat_sum
+    _lat = xr.open_dataset(conf["loss"]["latitude_weights"])["latitude"].values
+    dims_lon = xr.open_dataset(conf["loss"]["latitude_weights"])["longitude"].shape[0]
+
+    weights = np.cos(np.deg2rad(_lat))
+    repeated_weights = np.repeat(weights[np.newaxis, :], dims_lon, axis=0)
+    L= repeated_weights.T
```
djgagne, line 178: The weights should be divided by the sum of all the weights so that the weights add up to 1. I think the repeat is a good idea but you should also still calculate the sum.

djgagne (dismissed):

credit/loss.py
WillyChap: @djgagne check it out to see if this works please!

yingkaisha (approved):

PR #105, opened 2024-09-20
Infrastructure for post-processing blocks (e.g. SKEBS, mass correction, laplacian filtering)

dkimpara:
`postBlock.py` will be the code to handle all post-processing blocks. please use it to develop your post blocks.

Usage:

1. Adding postBlock to an existing trained model:

    - copy save directory of existing model
    - modify configs to add post block, change save_loc etc

2. Training a fresh model with a postblock: same as regular model training

Developing in postBlock:
- add your code to postBlock.py
- modify configs and parser to reflect changes

Details:
- parser has been updated to include default configs for the postblock
- added tests for unet, crossformer and postblock

TODO:
- update unet to conform to new model schema

### DIFF
--- a/applications/rollout_metrics.py
+++ b/applications/rollout_metrics.py
@@ -21,7 +21,6 @@ import numpy as np
 
 # ---------- #
 import torch
-import torch.distributed as dist
 
 # ---------- #
 # credit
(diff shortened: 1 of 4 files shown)


credit/models/crossformer.py
```diff
@@ -401,8 +404,15 @@ def __init__(
             logger.info(f"Padding each longitudinal boundary with {self.pad_lon} pixels from the other side")
         if self.pad_lat > 0:
             logger.info(f"Padding each pole using a reflection with {self.pad_lat} pixels")
+        
+        
+        self.use_postBlock = post_conf["use_skebs"] # or post_conf["use_lap"] etc
```
djgagne: Can you use consistent snake case rather than mixing snake and camel case?

credit/postBlock.py
```diff
@@ -0,0 +1,58 @@
+import torch
```
djgagne: For module naming, please use `postblock.py` or `post.py`. Python does not use camel case.

credit/models/crossformer.py
dkimpara: done

credit/postBlock.py
dkimpara: done

PR #117, opened 2024-10-16
updating the periodic padding function to roll then reflect lats

WillyChap:


credit/boundary_padding.py
```diff
@@ -0,0 +1,139 @@
+
+import torch
+import torch.nn.functional as F
+
+class TensorPadding:
+    def __init__(self, conf_padding):
```
djgagne: Can you replace `conf_padding` with `mode="earth", pad_NS=40, pad_WE=40, **kwargs` That way the args would be consistent with the docstring, and the class could still take a dictionary as input.

credit/models/crossformer.py
```diff
@@ -321,9 +316,9 @@ def __init__(
         cross_embed_strides=(4, 2, 2, 2),
         attn_dropout=0.,
         ff_dropout=0.,
-        pad_lon=0,
-        pad_lat=0,
         use_spectral_norm=True,
+        interp=True,
+        padding_conf={'activate': False},
```
djgagne: The default value of a keyword argument should be an immutable type. Dictionaries are not immutable if used as a default argument, which can result in weird [side effects](https://docs.python-guide.org/writing/gotchas/) if the function or method is called multiple times with the default argument. Better practice is to set the default as None and then include the following if statement in the constructor:
```python
if padding_conf is None:
    padding_conf = {"activate": False}
if post_conf is None:
    padding_conf = {"activate": False}
```

credit/models/crossformer.py
```diff
@@ -368,28 +364,63 @@ def __init__(
...
+            # append everything
+            self.layers.append(
+                nn.ModuleList([
+                    cross_embed_layer,
+                    transformer_layer
+                ])
+            )
+        
+        if self.use_padding:
+            self.padding_opt = TensorPadding(padding_conf)
```
djgagne: Should be changed to **padding_conf if other changes are made.

credit/models/fuxi.py
```diff
@@ -288,15 +280,30 @@ def __init__(self,
                  num_heads=8,
                  depth=48,
                  window_size=7,
-                 pad_lon=80,
-                 pad_lat=80,
                  use_spectral_norm=True,
+                 interp=True,
+                 padding_conf={'activate': False},
```
djgagne: Need to fix mutable default arguments here too.

PR #134, opened 2024-12-08
Major update for the inference routine + QoL improvements

yingkaisha:
**New features**

* Now `grad_max_norm: 0` will turn off gradient clipping
* In `credit.parser`, `conf['trainer']['mode']` is verified to be one of the `fsdp`, `ddp`, or `none`.
* In `credit.parser`, the data availability checks now produce warnings instead of errors. 
    - If you require to predict 2024 but only have data up to 2023, you will receive warnings, and the inference routine will stop at the end of 2023.
* The following config options will now trigger warnings:
    - FuXi with `use_spectral_norm: False` and `mode: fsdp` --- this does not work. The reason is unknown.
  
* Datasets defined in `rollout_to_netcdf.py` and `rollout_metrics.py` are now combined as `Predict_Dataset_Iter` in `credit.data`

**Bugfix**
* Diagnostic variable issues in #120 is fixed in `rollout_metrics.py`.
* Bugfix the model weights loading in `rollout_to_netcdf.py` and `rollout_metrics.py` for ddp and single-GPU mode.


credit/data.py
```diff
@@ -1443,6 +1492,316 @@ def __iter__(self):
                     break
 
 
+# class Predict_Dataset(torch.utils.data.IterableDataset):
```
jsschreck: Before you merge, delete this old dataset and rename the new one Predict_Dataset.

jsschreck (approved):

PR #135, opened 2024-12-09
Addition of new multi-step dataset that allows batch size > 1

jsschreck:
The config handling for setting up variables for data is also refactored and moved to __init__.py. I will update the train.py and train_multistep.py shortly. There is some play with setting globals, though, this messes up linting, so its not quite in action yet.

@kanz76 @WillyChap since you guys are the targets of this update, you can test the script simply by pythoning it (note if name == main, the test is there for now). I will add a proper test soon. But the main difference between this and the old one is very small. 

```
dataset_multi = ERA5_MultiStep_Batcher(
        varname_upper_air=data_config['varname_upper_air'],
        varname_surface=data_config['varname_surface'],
        varname_dyn_forcing=data_config['varname_dyn_forcing'],
        varname_forcing=data_config['varname_forcing'],
        varname_static=data_config['varname_static'],
        varname_diagnostic=data_config['varname_diagnostic'],
        filenames=data_config['all_ERA_files'],
        filename_surface=data_config['surface_files'],
        filename_dyn_forcing=data_config['dyn_forcing_files'],
        filename_forcing=data_config['forcing_files'],
        filename_static=data_config['static_files'],
        filename_diagnostic=data_config['diagnostic_files'],
        history_len=data_config['history_len'],
        forecast_len=data_config['forecast_len'],
        skip_periods=data_config['skip_periods'],
        max_forecast_len=data_config['max_forecast_len'],
        transform=load_transforms(conf),
        batch_size=batch_size
    )

    dataloader = DataLoader(
        dataset_multi,
        batch_size=1,  # Adjust the batch size as needed
        shuffle=False,   # Shuffle is now handled in the dataset above, automatically for now. 
        num_workers=1,  # Still has to be one for multi-step datasets
        drop_last=True,  # Drop the last incomplete batch if not divisible by batch_size,
        prefetch_factor=4 # This will keep the Dataloader loading data even while data moves through model, etc.
    )
```

### DIFF
--- a/credit/datasets/__init__.py
+++ b/credit/datasets/__init__.py
@@ -1,0 +1,296 @@
+import os
+import sys
+import glob
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+def set_globals(data_config, namespace=None):
+    """
+    Sets global variables from the provided configuration dictionary in the specified namespace.
+
+    This method updates the global variables in either the given `namespace` or the 
+    caller's namespace (if `namespace` is not provided). If the `namespace` is not specified, 
+    it uses the global namespace of the caller (using `sys._getframe(1).f_globals`).
+
+    Parameters:
+    - data_config (dict): A dictionary where the keys are the global variable names 
+      and the values are the corresponding values to set.
+    - namespace (dict, optional): The namespace (or dictionary) where the global variables 
+      should be set. If not provided, the caller's global namespace is used.
+
+    The method logs each global variable being created and its name.
+    """
+
+    target = namespace or sys._getframe(1).f_globals
+    target.update(data_config)
+
+    # Identify if this is the __main__ namespace
+    name = target.get('__name__')
+
+    for key in data_config:
+        logger.info(f"Creating global variable in {name}: {key}")
+
+
+def setup_data_loading(conf):
+    """
+    Sets up the data loading configuration by reading and processing data paths, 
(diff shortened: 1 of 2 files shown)


jsschreck:
Two more Datasets are added which handle multiprocessing. The first one does not do pre-fetching, while the second one does. Note that the second one is experimental, there are some proper complexities with managing the workers in the background when getitem is not being called directly (in other words, its doing what Dataloader normally does when you use prefetch option). 

You may now test each of the three Datasets living in the script via 

`python era5_multistep_batcher.py (1|2|3)`

It will run a small test and print to screen how you need to use Dataloader.

jsschreck:
@kanz76 thanks for testing. I will fix the behavior you are seeing (I assume you just get the same batch for each device); forgot to set a seed in them, but following that it should fix the problem you are seeing.

jsschreck:
Options 1 and 2 should now work correctly with scaling across GPUs. shuffle = False has also been enabled.

dkimpara:
Is there a plan to integrate the new dataset into train_multistep.py?

jsschreck:
@dkimpara I will integrate it once we are done testing; Arnold has been doing this so far. I think its almost ready so soon I will add it to the train script. But feel free to pull it down and add it yourself if you wanted to test it. There are instructions on how to use it when you python ().py 1 (or 2) to test the two options.

kanz76 (approved):
Tested Option 1 & 2 on multiple GPUs. Both look good!

PR #137, opened 2024-12-20
Ensemble capability

dkimpara:
- supports batch_size > 1
- currently only KCRPS (unbiased CRPS) available
- train and validation are the same loss, for this mode

Compute limitations:
- single A100-80g GPU
- CESM 1dg 32 level data, small wxformer
- max effective batch size (batch_size * ensemble_size) = 8
- multistep, 1 backprop, using train_universal does not seem to affect this

new config field and options:
`int trainer.ensemble_size = 1` by default (also set by parser)
`str loss.training_loss = KCRPS`

high-level code design:
**randomness has to live inside of the model object**

- `era5trainer_v2` and `trainerERA5_multistep_grad_accum.py`: `torch.repeat_interleave` to copy samples to transform batches from (b, ...) to (b * ensemble_size, ...). Then this new tensor is passed into the model
- loss.py, KCRPSLoss: `torch.vmap` to vectorize and lift single obs modulus loss to handle ensembles with multiple target obs (batch_size > 1)
- metrics.py: compute ensemble mean to compute metrics on

Future features (did not put them in this PR so to not clutter the config)
- different train/validation losses (including ensemble/not-ensemble pairings)
- different ensemble sizes for train/validation
- ensembling features and postblock put in baseModel

Testing (on casper)

`applications/train.py -c config/test_cesm_ensemble.yml -l 1`

`applications/train_universal.py -c config/example-v2025.2.0-ensemble.yml`


jsschreck:
@dkimpara the batch_size > 1 ... is this meaning it will support the new datasets I added, or are you doing it yourself here?

dkimpara:
I haven't yet configured it for your datasets. I will do that once your PR
is settled. What I meant was that the ensemble features are compatible with
batches with more than one target
Dhamma Kimpara

PhD Candidate in Computer Science
National Center for Atmospheric Research, and University of Colorado Boulder
Pronouns in use: He/Him/His

jsschreck:
@dkimpara trainerERA5_v2.py will be deprecated in my PR; we will only need trainerERA5_multistep_grad_accum.py going forward, so you should add the ensemble change to that script as well. Keep what you have b/c I am not going to remove _v2 just yet since other people are still using it.

credit/trainers/trainerERA5_v2.py
```diff
@@ -174,6 +174,13 @@ def train_one_epoch(
                     x = torch.clamp(x, min=clamp_min, max=clamp_max)
                     y = torch.clamp(y, min=clamp_min, max=clamp_max)
                 
+                # --------------------------------------------- #
+                # ensemble
+                # copies each sample in the batch ensemble_size number of times. 
+                # if samples in the batch are ordered (x,y,z) then the result tensor is (x, x, ..., y, y, ..., z,z ...)
+                # WARNING: needs to be used with a loss that can handle x with b * ensemble_size samples and y with b samples
+                x = torch.repeat_interleave(x, conf["trainer"]["ensemble_size"], 0)
+
```
yingkaisha, line 186: This place can have an if condition, so the deterministic training routine will not touch this part.

credit/trainers/trainerERA5_v2.py
```diff
@@ -457,6 +464,13 @@ def validate(self, epoch, conf, valid_loader, criterion, metrics):
...
```
yingkaisha, line 482: if condition

credit/metrics.py
```diff
@@ -28,11 +28,17 @@ def __init__(self, conf, predict_mode=False):
...
+
     def __call__(self, pred, y, clim=None, transform=None, forecast_datetime=0):
         if transform is not None:
             pred = transform(pred)
             y = transform(y)
 
+        # calculate ensemble mean, if ensemble_size=1, does nothing
+        pred = pred.view(y.shape[0], self.ensemble_size, *y.shape[1:]) #b, ensemble, c, t, lat, lon
+        pred = pred.mean(dim=1)
+
```
yingkaisha, line 42: if condition

credit/parser.py
```diff
@@ -710,6 +710,9 @@ def credit_main_parser(
             "train_batch_size" in conf["trainer"]
         ), "Training set batch size ('train_batch_size') is missing from onf['trainer']"
 
+        if "ensemble_size" not in conf["trainer"]:
+            conf["trainer"]["ensemble_size"] = 1
```
yingkaisha: Can you have an option to skip the ensemble routine? For those who do not use ensemble training, going through the ensemble code can cause a slow-down.

yingkaisha (changes requested):
Finished review

credit/metrics.py
```diff
@@ -28,11 +28,17 @@ def __init__(self, conf, predict_mode=False):
         # DO NOT apply these weights during metrics computations, only on the loss during
         self.w_var = None
 
+        self.ensemble_size = conf["trainer"]["ensemble_size"]
```
jsschreck: If I do not use ensemble_size this will fail here. Prob better to have something like "if ensemble_size in conf["trainer"] conditional

credit/metrics.py
jsschreck: Yeah -- will also fail here.

credit/parser.py
dkimpara: I will make sure ensemble_size=1 (deterministic) doesn't trigger the ensemble code. This default will be set in parser

credit/metrics.py
dkimpara: I've set the default ensemble_size to be 1 (meaning deterministic) in the parser. IMO for code clarity it's easier to deal with having an argument always exist rather than not

dkimpara:
@jsschreck @yingkaisha 
looking at the new codebase with many new training options, would it be better to add ensembling code to dataloaders? what do you all think?

dkimpara:
> @jsschreck @yingkaisha looking at the new codebase with many new training options, would it be better to add ensembling code to dataloaders? what do you all think?

oof. nevermind. i took a look at the datasets code. looks pretty complex

credit/metrics.py
dkimpara: nevermind. I think i'll add it just to be safe. we might have different parsers going around

credit/parser.py
dkimpara: see parser and new if statements

yingkaisha:
I will do the review by Monday

yingkaisha:
Looks good to me

yingkaisha (approved):

jsschreck (approved):

PR #139, opened 2024-12-24
Multistep training with batch_size >=1 per GPU

jsschreck:


dkimpara:
can you add some details? is this going to replace all trainers/datasets or just the multistep ones?

jsschreck:
> can you add some details? is this going to replace all trainers/datasets or just the multistep ones?

I just added the single-step dataset. I am aiming for consolidation down to a single trainer with config options to specify what you need to do (single or multi). Working on that and should have it ready by end of week. So far the train_multistep.py in this PR has been tested with the new multi-step datasets. I will update documentation and what not once its all working.

credit/datasets/load_dataset_and_dataloader.py
```diff
@@ -0,0 +1,306 @@
...
+
+    Returns:
+        Dataset: The loaded dataset.
+    """
+    seed = conf["seed"]
+    conf = credit_main_parser(
+        conf, parse_training=True, parse_predict=False, print_summary=False
+    )
+    training_data_check(conf, print_summary=False)
+    data_config = setup_data_loading(conf)
```
kanz76: Seems redundant, since this is already performed in `train*.py`

jsschreck: Id rather not make downstream things like this depend on having to run the train* script. I am also going to be making changes to the parser and train_data_check after the new year ...

jsschreck: Maybe I can toss training_data_check, but if one does not run credit_main_parser the conf is incomplete (which I do not like).

kanz76: We could also just run the parser before using the downstream code (such as `load_dataset`) in whatever context it is being used, without having to run train* scripts. This can also avoid having multiple instances of configs in the code, which might differ (by accident)

jsschreck: @kanz76 it looks like in my latest version I already did this. so the calls within the class can come out now. This will get updated next PR when the parser gets restructured.

credit/datasets/load_dataset_and_dataloader.py
```diff
@@ -5,13 +5,11 @@
     MultiprocessingBatcher,
     MultiprocessingBatcherPrefetch
 )
-from credit.data import ERA5_and_Forcing_Dataset
 from torch.utils.data import DataLoader
 from torch.utils.data.distributed import DistributedSampler
 from credit.transforms import load_transforms
-from credit.parser import credit_main_parser, training_data_check
-from credit.datasets import setup_data_loading, set_globals
```
kanz76: import setup_data_loading should be restored since it is used in the load_dataset function.

jsschreck: Good catch! Thanks.

jsschreck:
@kanz76 I think the bug involving batch size and history len is now corrected.

kanz76 (approved):
I tested 

- `era5_multistep_batcher.py`  (except`MultiprocessingBatcherPrefetch`) 
- `load_dataset_and_dataloader.py` 
- `train_universal.py` 

They all look good!

kanz76 (approved):

PR #140, opened 2025-01-04
Fix  train/valid split read in dataset

kanz76:
This PR addresses an issue with the train/valid split in the dataset. The current implementation reads the entire dataset for both the training and validation sets, which can lead to data leakage.


jsschreck (approved):
Thanks for fixing!

PR #142, opened 2025-01-06
Predict scripts with a batch size

jsschreck:
Both `rollout_to_netcdf` and` rollout_metrics` should now support batch size >= 1. A config option `batch_size` was added to example-v2025.2.0.yml. 

The metrics scripts will also support calculation of ACC anamoly scores by supplying a climatology file. In the new yaml, this is specified by `predict: climatology: '/glade/campaign/cisl/aiml/ksha/CREDIT_arXiv/VERIF/verif_6h/ERA5_clim/ERA5_clim_1990_2019_6h_interp.nc'`

Finally, a new script, `rollout_ensemble.py` has been added which should create an ensemble using bred vectors.


applications/rollout_ensemble.py
```diff
@@ -0,0 +1,626 @@
...
+    seed_everything(seed)
+
+    # number of input time frames
+    history_len = conf["data"]["history_len"]
+
+    # length of forecast steps
+    lead_time_periods = conf["data"]["lead_time_periods"]
+
+    # batch size
+    batch_size = conf["predict"].get("batch_size", 1)
```
djgagne, line 123: I didn't realize `get` was a dict method until your use here just now. This is a really succinct way to handle missing defaults.

applications/rollout_ensemble.py
```diff
@@ -0,0 +1,626 @@
...
+                if "y_diag" in batch:
+                    y_diag_batch = batch["y_diag"].to(device).permute(0, 2, 1, 3, 4)
+                    y_batch = torch.cat((y_batch, y_diag_batch), dim=1).to(device).float()
+
+                # Clamp if needed
+                if flag_clamp:
+                    x_batch = torch.clamp(x_batch, min=clamp_min, max=clamp_max)
+
+                # Generate bred vectors
+                bred_vectors = []
```
djgagne: Can we move the bred vector functionality to a function in a credit module? It would be more reproducible and testable that way.

credit/metrics.py
```diff
@@ -100,3 +102,203 @@ def __call__(self, pred, y, clim=None, transform=None, forecast_datetime=0):
...
+                for i, var in enumerate(ordered_acc_vars):
+                    clim = self.get_climatology(forecast_datetime, var).to(
+                        dtype=pred.dtype, device=pred.device
+                    ).unsqueeze(0)
+                    anomalies_pred.append(acc_pred[:, i] - clim)
+                    anomalies_y.append(acc_y[:, i] - clim)
+
+                anomalies_pred = torch.stack(anomalies_pred, dim=1)
+                anomalies_y = torch.stack(anomalies_y, dim=1)
+
```
djgagne: I like the addition of climatology metrics. Any way to split the individual metrics into their own functions?

credit/metrics.py
```diff
@@ -100,3 +102,203 @@ def __call__(self, pred, y, clim=None, transform=None, forecast_datetime=0):
...
+                for i, var in enumerate(self.acc_vars):
+                    pred_prime = anomalies_pred[:, i] - torch.mean(anomalies_pred[:, i])
+                    y_prime = anomalies_y[:, i] - torch.mean(anomalies_y[:, i])
+
+                    # Add epsilon to avoid division by zero
```
djgagne: I recommend using `torch.maximum(denominator, epsilon)` instead of adding epsilon to the denominator. 1e-7 may also be too large.

credit/metrics.py
```diff
@@ -100,3 +102,203 @@ def __call__(self, pred, y, clim=None, transform=None, forecast_datetime=0):
...
+        "/glade/derecho/scratch/schreck/repos/miles-credit/production/multistep/wxformer_6h/model.yml"
+    ) as cf:
+        conf = yaml.load(cf, Loader=yaml.FullLoader)
+
+    conf = credit_main_parser(
+        conf, parse_training=True, parse_predict=False, print_summary=False
+    )
+
+    # Climatology file
+    fn = '/glade/campaign/cisl/aiml/ksha/CREDIT_arXiv/VERIF/verif_6h/ERA5_clim/ERA5_clim_1990_2019_6h_interp.nc'
```
djgagne: How big is the file? It would be good to include this either in the repo or a cloud bucket.

credit/metrics.py
```diff
@@ -100,3 +102,203 @@ def __call__(self, pred, y, clim=None, transform=None, forecast_datetime=0):
...
+    from credit.parser import credit_main_parser
+
+    logging.basicConfig(
+        level=logging.INFO,
+        format='%(asctime)s - %(name)s - %(levelname)s - %(message)s'
+    )
+    logger = logging.getLogger(__name__)
+
+    # Open an example config
+    with open(
```
djgagne: Please point example config to relative path in repo.

djgagne (changes requested):
I like the new predict scripts, especially adding ensemble support and fixing the metrics. I have requested some changes for modularity purposes.

applications/rollout_ensemble.py
jsschreck: Yeah super convenient!

applications/rollout_ensemble.py
jsschreck: I moved the code to a new directory called credit.ensemble.

credit/metrics.py
jsschreck: @djgagne can you give an example? You want me to split ACC off from the others? Should I add more than just RMSE, MSE, and MAE?

credit/metrics.py
jsschreck: Done -- I also changed epsilon to 1e-8

credit/metrics.py
jsschreck: Kyle did not add in level climatologies, so its not very large. Agreed it should be available to users so if they want to create their own, they know how it should be structured.

credit/metrics.py
jsschreck: It should now point to example-v2025.2.0.yml in /config

credit/metrics.py
dkimpara: i checked the filesize and its way too big. @jsschreck do you have a bucket link?

`63G     ./ERA5_clim_1990_2019_6h_interp.nc`

credit/metrics.py
dkimpara: resolved and tested in latest commit

credit/metrics.py
dkimpara: moved to an item on paper to-do list

dkimpara (approved):
modularized code and resolved DJs comments

PR #148, opened 2025-01-16
bugfix: removed deprecated ERA5Dataset and Bridgescaler_dataset from train.py

dkimpara:
- ERA5Dataset and Bridgescaler_datasets deprecated (removed from data.py in previous PR)

test with: `config/wxformer_1dg_test.yml`


jsschreck (approved):
Looks fine and minor so approving

PR #150, opened 2025-01-27
Bugfix on hybrid sigma-pressure to constant pressure level interpolation

yingkaisha:
**Bugfix**
* Minor fixes on `credit.interp`. Now it allows user-defined model-level coefficient names.
* Added some documentation to the script.
* Now `geopotential at surface` is loaded from `model_level_file` instead of the rollout files. This variable is static and typically not an output.
* A new file `test_interp_static.nc` is added to pair to the existing `test_interp.nc`.

### DIFF
--- a/tests/test_interp.py
+++ b/tests/test_interp.py
@@ -10,6 +10,9 @@ def test_full_state_pressure_interpolation():
     input_file = os.path.join(path_to_test, "data/test_interp.nc")
     ds = xr.open_dataset(input_file)
 
+    # load static file with interpolation info 
+    model_level_file = os.path.join(path_to_test, "data/test_interp_static.nc")
+    
     # pressure levels to interpolate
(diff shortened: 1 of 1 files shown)


yingkaisha:
Ready

djgagne:
I am ok with everything but adding the surface geopotential height to the model level netcdf file. I currently include the model level file in the credit metadata since it's a very small file, and adding surface geopotential would make the file a lot bigger. The ERA5 model levels and their coefficients are also constant across different spatial resolutions, whereas the surface geopotential field is not. When I call the pressure interpolation in credit/output.py, I link the surface geopotential variable to the model state xarray Dataset. 

https://github.com/NCAR/miles-credit/blob/b27e22f259370bb255c4dc2b3a182b80ff678b12/credit/output.py#L178-L192

yingkaisha:
> surface geopotential

surface geopotential is constant.

PR #153, opened 2025-02-06
Fixed trainer classes to actually use DDP

kanz76:
The current behavior of the trainers removes the DDP wrapper from the model, leading to independent models on different GPUs that do not share gradient updates. This caused error spikes when reloading weights to resume training.

To test this PR, use `config/model_test_new_dataset.yml`, which contains a small subset of ERA5, runnable in a couple of minutes. For instance, you can run for 2 epochs and then resume training by reloading the weights to confirm that the error does not spike. Note: Ensure that trainer:type: is set to `ddp` in the config file.

### DIFF
--- a/credit/distributed.py
+++ b/credit/distributed.py
@@ -232,8 +232,7 @@ def distributed_model_wrapper(conf, neural_network, device):
         )
 
     elif conf["trainer"]["mode"] == "ddp":
-        model = DDP(neural_network, device_ids=[device]) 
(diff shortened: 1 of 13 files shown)


credit/distributed.py
jsschreck, line 1: This is where I think we should do the check for unused parameters and set to False if there are None. Torch doc says we may take a performance hit when using True.

kanz76: At this stage, the model is not run yet. We can't tell which parameters won't be used until the forward method of the model is executed.

jsschreck (approved):

PR #157, opened 2025-02-10
FSDP to regular model or model weights

jsschreck:
Added new method to extract either non-wrapped model or model weights from FSDP module.

Tagging a bunch of you so you are aware it now exists. See the if name == main to see how to use it (though pretty easy)


djgagne (approved):
Helpful addition!

PR #158, opened 2025-02-11
Fixes to interpolation to pressure levels

djgagne:
I made some significant revisions to the pressure level interpolation module to accommodate the issues @yingkaisha 's PR raised and to make the pressure interpolation process for geopotential and temperature consistent with [CESM/ECMWF](https://doi.org/10.5065/D6HX19NH). New features:
* surface geopotential is now passed into `credit.interp.full_state_pressure_interpolation` as a numpy array rather than embedded in xarray Dataset with other state variables.
*  output.py updated to reflect changed API.
* temperature and geopotential are extrapolated below ground based on Trenberth et al. 1993. 
* mean sea level pressure reduction implemented based on Trenberth et al. 1993 and [CAM fortran code](https://github.com/ESCOMP/CAM/blob/cam_cesm2_2_rel/src/physics/cam/cpslec.F90). The formula adjusts the temperature and lapse rate correction based on the surface height. According to the paper you might still see some wonkiness in the Himalayas/Tibetan plateau, which would require some west-east interpolation over mountains.
* pressure to hybrid coordinate interpolation implemented. Interpolation routine ignores levels below ground but also requires surface pressure field to do so.


djgagne:
Everything should be ready for review now. Test passes and inference code runs on casper.

jsschreck (approved):

djgagne:
I have increased confidence that the [IFS postprocessing](https://confluence.ecmwf.int/display/OIFS/OpenIFS+Documentation?preview=%2F19661682%2F135572873%2FFULL+POS_ykfpos43.pdf) documentation is wrong about two things. Equation 3 has ps and pL flipped, as does equation 6. It also looks like the height-dependent temperature corrections are being used to extrapolate geopotential after all. My answers are now really close to those from ERA5 and are probably about as close as I can get without direct access to the FULL POS source code, which I am still working on getting. @jsschreck and @yingkaisha please look over because I want to get everything merged this afternoon.

djgagne:
Thanks John! Merging now.

PR #165, opened 2025-02-25
Renaming scripts

jsschreck:


jsschreck:
@djgagne @WillyChap @yingkaisha @kanz76 This is a big PR but not a lot changes for all of you. I am mainly deprecating the various trainer*.py scripts in credit.trainers, removing old models from credit.models and simplifying the main train script to "train.py" (it works for both single and multi-step training). You should not have to do anything else (other than potentially adding "type" to the trainer config, as there is "universal and conus404 options". It might make more sense if we rename to "era5" and a comparable (identical) call for "cam". But let me know your preference. If you do not currently use the trainer:type, it will complain at you and ask you to add it to your config file.

djgagne:
The refactorings all look reasonable to me. I agree with renaming universal to era5 and add other options for other datasets as I expect we will be using other datasets in the future. It looks like we also might be able to remove rotary-embedding-pytorch and vector-quantize-pytorch as dependencies unless there is somewhere else in the code where they are being called?

jsschreck:
@djgagne I updated according to your comments, including removing the two dependencies relates to old models.

djgagne (approved):
LGTM

PR #173, opened 2025-02-27
GFS initial conditions

charlie-becker:
First pass at creating some initial conditions from GDAS (GFS data assimilation). Includes a module to do all the work `nwp.py` and a demo script to run it.  It still needs some optimization, sanity checking, documentation, and some general clean up. Feedback appreciated.  It currently saves out to zarr and should match the zarr format for ERA5 exactly.


charlie-becker:
Fixed a couple of bugs. Data is output to Zarr in the correct format which includes surface and upper air variables. 

However, it is likely the geopotential (Z500) is not derived correctly and needs more investigation.

credit/nwp.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def load_gfs_data(full_file_path, variables):
+
+    ds = xr.open_dataset(fsspec.open(full_file_path).open())
+    available_vars = ds.data_vars
+    vars = [v for v in variables if v in available_vars]
+    ds = ds[vars].rename({'grid_xt': 'longitude', 'grid_yt': 'latitude'}).load()
+    if 'delz' in available_vars and 'delz' in variables:
+        ds['delz'].values = np.abs(ds['delz'].cumsum(dim='pfull').values) * 9.81
```
djgagne: The summation needs to be done in reverse since the model level closest to the surface has the last index.

applications/gfs_init.py
```diff
@@ -0,0 +1,24 @@
...
+
+date = "202503040600"
+model_levels = [ 10,  30,  40,  50,  60,  70,  80,  90,  95, 100, 105, 110, 120, 130, 136, 137]
+variables = config['data']['variables'] + config['data']['surface_variables']
+
+gfs_init = build_GFS_init(output_grid=credit_grid,
+                          variables=variables,
+                          model_level_indices=model_levels)
+
+gfs_init.to_zarr("/glade/derecho/scratch/cbecker/gfs_init_20250304_all.zarr")
```
djgagne: Please fix the hard coding of the date and model levels and use pandas Timestamp and other datetime functionality to handle time formatting.

credit/nwp.py
```diff
@@ -0,0 +1,134 @@
...
+import xarray as xr
+import fsspec
+import xesmf as xe
+from credit.interp import geopotential_from_model_vars, create_pressure_grid
+
+gfs_map = {'tmp': 'T', 'ugrd': 'U', 'vgrd': 'V', 'spfh': 'Q', 'pressfc': 'SP', 'tmp2m': 't2m'}
+level_map = {'T500': 'T', 'U500': 'U', 'V500': 'V', 'Q500': 'Q', 'Z500': 'Z'}
+upper_air = ['T', 'U', 'V', 'Q', 'Z']
+surface = ['SP', 't2m']
+STANDARD_GRAVITY = 9.80665
```
djgagne: Gravity is specified in credit/physics_constants.py. Please import from there.

applications/gfs_init.py
```diff
@@ -0,0 +1,24 @@
...
+parser = argparse.ArgumentParser()
+parser.add_argument("-c", "--config", help="Path to config file")
+args = parser.parse_args()
+with open(args.config) as config_file:
+    config = yaml.safe_load(config_file)
+
+credit_grid = xr.open_dataset('/glade/work/cbecker/miles-credit/credit/metadata/ERA5_Lev_Info.nc')
+
+date = "202503040600"
+model_levels = [ 10,  30,  40,  50,  60,  70,  80,  90,  95, 100, 105, 110, 120, 130, 136, 137]
```
djgagne: Model levels are in the ERA5 data, and we could store them in a config file or other smaller file.

credit/nwp.py
```diff
@@ -0,0 +1,134 @@
...
+    xp = regridded_nwp_data['P'].values
+    fp = regridded_nwp_data
+    output_pressure = (output_grid['a_half'] + output_grid['b_half'] * regridded_nwp_data['SP'])
+    sampled_output_pressure = output_pressure[model_level_indices].values
+    ny, nx = regridded_nwp_data.sizes['latitude'], regridded_nwp_data.sizes['longitude']
+    interpolated_data = {}
+    for var in upper_vars:
+        fp_data = fp[var].values
+        interpolated_data[var] = {'dims': ['latitude', 'longitude', 'level'],
+                                  'data': np.array([np.interp(sampled_output_pressure[:, j, i], xp[:, j, i], fp_data[:, j, i])
```
djgagne, line 174: I have a numba-accelerated version of this function in interp.py. I would recommend replacing this function with that.

applications/gfs_init.py
```diff
@@ -0,0 +1,24 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-c", "--config", help="Path to config file")
+args = parser.parse_args()
+with open(args.config) as config_file:
+    config = yaml.safe_load(config_file)
+
+credit_grid = xr.open_dataset('/glade/work/cbecker/miles-credit/credit/metadata/ERA5_Lev_Info.nc')
+
+date = "202503040600"
```
djgagne: Date should be passed as a config argument and use pandas Timestamp to handle more general parsing of the date format.

djgagne (changes requested):
I appreciate adding this functionality to CREDIT. In addition to the specific comments, please add docstrings in Google style to each function.

charlie-becker:
All comments have been addressed. A couple notes:

1. xesmf is still currently required for this. I think it would be good to save the weights of the regridder and load them directly. If it's not too big, we can put it in the repo, else just load it from glade. It will also speed up the script by about 2-3 minutes.
2. We should probably add a "most recent" flag somewhere to grab the most recent data from nomads when we run it in real time. 
3. We also need to include some tests for this

credit/nwp.py
charlie-becker: This is to constant pressure levels. Since this code is going from hybrid to hybrid, I left the code as is.

charlie-becker:
Ready for review.

applications/gfs_init.py
```diff
@@ -0,0 +1,36 @@
...
+import yaml
+import argparse
+import xarray as xr
+import os
+from os.path import join
+import pandas as pd
+
+username = os.environ.get('USER')
+base_path = '/glade/work/username/'.replace("username", username)
+base_out_path = '/glade/derecho/scratch/username/CREDIT/GFS_INIT/'.replace("username", username)
```
djgagne: Please do not hard code GLADE paths into the script. We would like to use this on laptops, cloud VMs, and other machines in general. Make these arguments in argparse.

applications/gfs_init.py
```diff
@@ -0,0 +1,36 @@
...
+os.makedirs(base_out_path, exist_ok=True)
+parser = argparse.ArgumentParser()
+parser.add_argument("-c", "--config", help="Path to config file")
+args = parser.parse_args()
+with open(args.config) as config_file:
+    config = yaml.safe_load(config_file)
+
+credit_grid = xr.open_dataset(join(base_path,'miles-credit/credit/metadata/ERA5_Lev_Info.nc'))
```
djgagne: If you need to automatically figure out a path relative to a given file, there is a way to do it with the os library. I use it [here in test_interp.py](https://github.com/NCAR/miles-credit/blob/507c9834a54e9daa252e6af4f559f3032fc9d170/tests/test_interp.py#L8C5-L8C62)

applications/gfs_init.py
charlie-becker: Do you prefer these be added as a separate arg flag or added to the base config?

djgagne: Put it in the config.

charlie-becker: Added tiny new conf file and cleaned paths. Ready for re-review

djgagne (approved):
Charlie, I appreciate you making all the fixes. Everything looks good to me.

PR #174, opened 2025-03-01
Detached and re-assigned y_pred to discard  computational graph

kanz76:
Detached and re-assigned y_pred to discard current computational graph, which still exists if `forecast_step` not in `backprop_on_timestep`


djgagne (approved):
Change seems reasonable and glad the test breaking is fixed.

PR #175, opened 2025-03-05
Small temporary update to pbs.py to fix newer cudnn settings on Derecho

jsschreck:
Also some minor updates to ensemble model and example config that fixes variable weights list.


djgagne (approved):
LGTM

PR #176, opened 2025-03-13
Realtime Rollout and Predict + Interpolation Updates

djgagne:
I have created a new realtime rollout dataset and script that will handle GFS as input or any other model processed into a single timestep file like Charlie's GFS data as input. It still needs to be tested, so please do not merge just yet. I also have not implemented the logic to handle 2 input times instead of 1 yet.


charlie-becker:
I've addressed a few bugs and have it pretty close. I'll continue working on it over the next couple of days.

jsschreck (approved):

PR #178, opened 2025-03-27
Updates to bred vector implementation and noisy WXFormer

jsschreck:


djgagne (approved):
Changes look good to me. I didn't see any clear issues in the bred vector math.

PR #181, opened 2025-04-03
Trainer documentation

jsschreck:


docs/source/training.md
```diff
@@ -0,0 +1,196 @@
...
+#PBS -A NAML0001
+#PBS -q casper
+#PBS -j oe
+#PBS -k eod
+source ~/.bashrc
+conda activate credit
+torchrun applications/train.py -c model.yml
+```
+
+and note that the ```torchrun``` command is used rather than MPIs. For now MPIs are not supported on Casper but that may change in a future release. But to get torch to run in distributed mode (either DDP or FSDP) we use torchrun to faciliate that (rather than setting that up manuanlly in train.py).
```
djgagne: Building PyTorch from source on casper would enable MPI support. torchrun works well-enough for single GPU applications.

docs/source/training.md
```diff
@@ -0,0 +1,196 @@
...
+#PBS -l select=1:ncpus=8:ngpus=1:mem=128g
+#PBS -l walltime=4:00:00
+#PBS -l gpu_type=a100
+#PBS -A NAML0001
+#PBS -q casper
+#PBS -j oe
+#PBS -k eod
+source ~/.bashrc
+conda activate credit
+torchrun applications/train.py -c model.yml
```
djgagne, line 180: Does torchrun require any extra arguments to point to the right port or whatnot? Stanley seemed to be having issues with this the other day.

docs/source/training.md
```diff
@@ -0,0 +1,196 @@
...
+
+### Key Differences
+
+| Feature          | Derecho          | Casper         |
+|-----------------|-----------------|---------------|
+| GPUs per node   | 4                | 1             |
+| Total GPUs      | 32 (8 nodes × 4) | 1             |
+| Memory          | 480GB            | 128GB         |
+| Walltime        | 12:00:00         | 4:00:00       |
+| GPU Type        | A100             | V100          |
```
djgagne: Casper has A100s and more memory per node than 128GB

djgagne (changes requested):
A few comments on the docs.

PR #186, opened 2025-04-14
Parallel version of CRPS

jsschreck:
This enables computation of the CDF needed in the CRPS loss across the GPUs. If ensemble_size > 1 and CRPS loss is used, the true ensemble size is ensemble_size * num_gpus.


kanz76:
I've reviewed `credit/datasets/load_dataset_and_dataloader.py`, as it's the bit of code I'm most familiar with. It looks good to me! Setting rank=0 and world_size=1 seems like a good approach to ensure consistency in data across GPUs (I'm using this method in my graph model, which is partitioned across GPUs, requiring data duplication across GPUs).

PR #190, opened 2025-04-21
Fixed seed issues with CRPS across GPUs

jsschreck:
Previously the noise in each model was replicated across the GPUs, so each GPU gave the same predictions. Its fixed in train.py and downstream. Also added extra conditional in load_* so that both KCRPS loss and era5-ensemble needs to be called for rank and world size to be set to 0 and 1, resp. to preserve (x,y) pairs across GPUs so we compute CDF correctly.


dkimpara (approved):
Mostly linting changes, only a couple lines actually changed. LGTM

PR #197, opened 2025-06-16
Credit losses refactor

jsschreck:
As per last discussion with Seth two Fridays ago, I said I would refactor loss.py. That is done here by creating a directory losses within credit, then mostly loading as before but some improvements.


djgagne (approved):
New losses look great and appreciate the documentation!

PR #200, opened 2025-06-30
Credit ptype

stanleyakor1:
This updated code incorporates the revised full_state_pressure_interpolation() routine. Notably, changes have been made to credit/output.py, credit/credit_ptype.py,  and the associated config files.

### DIFF
--- a/credit/credit_ptype.py
+++ b/credit/credit_ptype.py
@@ -1,5 +1,6 @@
 import numpy as np
 import os
+os.environ["KERAS_BACKEND"] = "torch"
 from numba import njit
 import xarray as xr
 from tqdm import tqdm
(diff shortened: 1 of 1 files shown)


credit/credit_ptype.py
```diff
@@ -0,0 +1,218 @@
...
+        
+        k = 0
+        for var_name in data.data_vars:
+            data_array[k:k+21, ...] = data[var_name].isel(time=0).values
+            k += 21  
+        data_array.reshape(n_vars, -1).T
+        return data_array.reshape(n_vars, -1).T
+        
+        
+    def load_scalar(self,scaler_path):
```
djgagne: Should be `load_scaler` rather than `load_scalar`

credit/credit_ptype.py
```diff
@@ -0,0 +1,218 @@
+import numpy as np
```
djgagne, line 1: Add 
```
import os
os.environ["KERAS_BACKEND"] = "torch"
```
at the top of the file before keras is imported to make sure it uses pytorch as the backend.

credit/credit_ptype.py
stanleyakor1: Noted!

credit/credit_ptype.py
stanleyakor1: Noted

stanleyakor1:
Thanks for the comments. All concerns have been addressed.

PR #201, opened 2025-07-04
Update docs admonitions and TOC to render properly

kafitzgerald:
Just a couple of minor things that caught my eye while reading through the docs.  Figured I'd throw in a PR in case it's helpful.

For the admonitions I could have just used the back tick syntax without updating the docs configuration, but the content in colon fences should render properly in standard Markdown editors as well which is nice.  See here: https://myst-parser.readthedocs.io/en/latest/syntax/optional.html#code-fences-using-colons

I also updated the table of contents to include both pages on ensembles.  Linking both pages to a single entry in the toc wasn't working quite right.


djgagne (approved):
Thank you for fixing the TOC and including the colon fences. That will definitely be useful in more docs.

PR #202, opened 2025-07-06
Code for the graph model

kanz76:
This PR contains the code for the graph model, including the model architecture, training, and inference.


kanz76:
The code requires the DGL package, which may not be installed in everyone's environment and could therefore cause runtime errors.

PR #206, opened 2025-07-09
Parallel GFS init and command line programs for train and rollout

djgagne:


djgagne:
@charlie-becker Can you review the PR and merge? I haven't added docstrings to gefs.py but am not going to get to them until after I return but would like the significantly updated codes in main.

charlie-becker (approved):

PR #212, opened 2025-08-08
Updates to IC perturbation schemes

jsschreck:
I still need to update the documentation page before this gets merged. Also I will add a notebook for easy testing.

Still we do not get spread from the bred-vector, or any other scheme except Greg and Trent's approach.


jsschreck:
You may currently test the bred vectors and what not by copying this repo,

credit/ensemble/color.py
```diff
@@ -0,0 +1,131 @@
...
+    - reddening = 0: White noise (uncorrelated, flat power spectrum)
+    - reddening = 1: Pink noise (1/f power spectrum)
+    - reddening = 2: Brown/Brownian/Red noise (1/f² power spectrum)
+    - reddening > 2: Higher-order red noise (1/f^n power spectrum)
+
+    Higher reddening values produce smoother, more spatially coherent patterns,
+    which are often more realistic for geophysical applications.
+
+    Parameters
+    ----------
```
djgagne: Can you convert the parameters section of the docstring from numpy to google style to be consistent with the rest of the CREDIT docs? Should be an easy AI task hopefully.

credit/ensemble/color.py
```diff
@@ -0,0 +1,131 @@
...
+    """
+
+    def __init__(self, amplitude: float = 0.05, reddening: int = 2):
+        self.amplitude = amplitude
+        self.reddening = reddening
+
+    def __call__(self, x: torch.Tensor) -> torch.Tensor:
+        """Generate spatially correlated noise matching input tensor dimensions.
+
+        Parameters
```
djgagne: Convert from numpy to google style docstring.

credit/ensemble/color.py
```diff
@@ -0,0 +1,131 @@
...
+        """Generate spatially correlated noise using frequency domain filtering.
+
+        Creates colored noise by:
+        1. Starting with white noise (uniform power spectrum)
+        2. Transforming to frequency domain
+        3. Applying power-law scaling (1/f^reddening)
+        4. Transforming back to spatial domain
+
+        Parameters
+        ----------
```
djgagne: Convert from numpy to google style docstring.

credit/ensemble/spherical.py
```diff
@@ -0,0 +1,302 @@
...
+
+    Warning
+    -------
+    For grids that don't naturally fit spherical harmonic constraints (N:2N or (N+1):2N),
+    this method generates noise at a larger valid grid size and symmetrically crops to
+    the target dimensions. This preserves most correlation properties but may introduce
+    minor boundary effects.
+
+    Parameters
+    ----------
```
djgagne: Convert from numpy to google style docstring.

applications/rollout_metrics_noisy_ics.py
```diff
@@ -279,151 +318,194 @@ def predict(rank, world_size, conf, backend=None, p=None):
...
+                        ensemble_members, delta_x_members = outputs
+                    else:
+                        ensemble_members = outputs
+
+                    for i, bv in enumerate(ensemble_members):
+                        if "x_forcing_static" in batch:
+                            bv = torch.cat((bv, x_forcing_batch), dim=1)
+                        ensemble_members[i] = bv
+
+                    # ensemble_members.append(x)
```
djgagne: Can you remove commented out code.

applications/rollout_metrics_noisy_ics.py
```diff
@@ -279,151 +318,194 @@ def predict(rank, world_size, conf, backend=None, p=None):
...
+                        ensemble_members, delta_x_members = zip(*[temporal_noise(x, None, 1) for _ in range(ensemble_size)])
+                        # Remove static and dynamic forcing channels from delta_x_members
+                        # Only keep perturbations for dynamic variables
+                        if static_dim_size > 0:
+                            delta_x_members = [delta_x[:, :-static_dim_size, ...] for delta_x in delta_x_members]
+
+                    else:
+                        ensemble_members = [x + noise(x) for _ in range(ensemble_size)]
+
+                    # ensemble_members = generate_bred_vectors_cycle(
```
djgagne: Remove commented out code.

applications/rollout_metrics_noisy_ics.py
```diff
@@ -3,6 +3,7 @@
 import os
```
djgagne, line 3: Do you have a version of rollout to netcdf or rollout realtime that supports the noisy ics? I only see rollout metrics with this.

credit/ensemble/bred_vector.py
```diff
@@ -1,9 +1,352 @@
...
+            # logger.info("using postblock")
+            self.postblock = PostBlock(self.post_conf)
+
+        if self.hemispheric_rescale is not None:
+            if not os.path.exists(terrain_file) or terrain_file is None:
+                raise FileNotFoundError(f"Terrain file {terrain_file} not found")
+            latlons = xr.open_dataset(terrain_file).load()
+            self.latitudes = torch.tensor(latlons.latitude.values)
+
+    def perturb(self, x_input: torch.Tensor, forecast_step: int = 1) -> torch.Tensor:
```
djgagne, line 93: It looks like your perturbation is being calculated by perturbing one field then iterating the model forward `forecast_step` number of steps. The problem is that you are then differencing the future perturbed model state with the original initial condition when you should be differencing with the analysis (new set of initial conditions derived from observations) valid at the same time as the perturbed forecast. The perturb function should have two sets of x_inputs, one for the initial conditions being perturbed and another that will be differenced against the forecast.

djgagne (changes requested):
I think I found the issue with your bred vector code. You're differencing against the original initial conditions being perturbed rather than new initial conditions valid at the same time as the perturbed forecast. This explains why the perturbations are so large. I also have some doc formatting fixes requested.

applications/rollout_metrics_noisy_ics.py
jsschreck: yeah its in the PR. I may not have fully updated the example config though. Will double check this morning

credit/ensemble/bred_vector.py
jsschreck: Ok, this is version 2 anyways for me already. Lets update again today (from what I could find from the HINS code their br looks like this, which did seem wrong).

credit/ensemble/color.py
jsschreck: Crap, will get them all updated.

jsschreck:
Ensemble files should now be updated with the right format docstring. The rollout script is also updated. I'm going to rename it since it implies noise but it supports several IC perturbation methods, not just adding noise. The last file still needing updated is bred_vectors.py

djgagne:
Changing the name of the `rollout_metrics_noisy_ic.py` file broke the link in `credit/applications`. I updated the link and am seeing if that passes the test.

djgagne (approved):

PR #213, opened 2025-08-21
Credit ptype

stanleyakor1:
- Use 'spawn' start method to avoid CUDA re-initialization in forked subprocesses
- Ensure predictions are moved to CPU before calling .numpy()


djgagne (approved):
Changes look good to me.  Merging.

PR #215, opened 2025-08-29
Update gfs_init.py paths for L137_model_level_indices.csv and ERA5_Lev_Info.nc

### LINKED ISSUE #214: gfs_init.py unable to find ERA5_Lev_Info.nc and L137_model_level_indices.csv
I'm not sure if the correct fix is to prefix the paths for these files with the `credit` directory.  I'll submit a PR to review.

sgpearse:
For me, gfs_init.py is unable to find L137_model_level_indices.csv and ERA5_Lev_Info.nc.  I believe the filepaths need to be prefixed with the `credit` directory.

Potentially fixes #214

### DIFF
--- a/applications/gfs_init.py
+++ b/applications/gfs_init.py
@@ -23,7 +23,7 @@ def main():
         os.path.join(base_path, "../credit/metadata/ERA5_Lev_Info.nc")
     )
     model_levels = pd.read_csv(
-        join(base_path, "../credit/metadata/L137_model_level_indices.csv")
+        os.path.join(metadata_path, "L137_model_level_indices.csv")
     )
     model_level_indices = model_levels["model_level_indices"].values
     variables = config["data"]["variables"] + config["data"]["surface_variables"]
--- a/applications/gfs_init.py
+++ b/applications/gfs_init.py
@@ -24,7 +24,7 @@ def main():
     else:
         metadata_path = os.path.join(base_path, os.pardir, "credit", "metadata")
     credit_grid = xr.open_dataset(
-        os.path.join(base_path, "../credit/metadata/ERA5_Lev_Info.nc")
+        os.path.join(metadata_path, "ERA5_Lev_Info.nc")
     )
     model_levels = pd.read_csv(
         os.path.join(metadata_path, "L137_model_level_indices.csv")
--- a/applications/gfs_init.py
+++ b/applications/gfs_init.py
@@ -19,6 +19,10 @@ def main():
     n_procs = args.proc
     os.makedirs(config["predict"]["initial_condition_path"], exist_ok=True)
     base_path = os.path.abspath(os.path.dirname(__file__))
+    if os.path.basename(os.path.abspath(os.path.join(base_path, os.pardir))) == "credit":
+        metadata_path = os.path.join(base_path, os.pardir, "metadata")
+    else:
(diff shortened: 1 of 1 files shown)


kafitzgerald (approved):
This looks correct to me.  

It seems like the updated paths might've been accidentally reverted in a recent PR (#206) causing the issue.  Would be good to confirm though to ensure that's the case and doesn't break something else.

kafitzgerald:
Something along these lines should probably do the trick for now:

```
    base_path = os.path.abspath(os.path.dirname(__file__))
    if os.path.basename(os.path.abspath(os.path.join(base_path, os.pardir))) == "credit":
        metadata_path = os.path.join(base_path, os.pardir, "metadata")
    else:
        metadata_path = os.path.join(base_path, os.pardir, "credit", "metadata")
    credit_grid = xr.open_dataset(
        os.path.join(metadata_path, "ERA5_Lev_Info.nc")
    )
    model_levels = pd.read_csv(
        join(metadata_path, "L137_model_level_indices.csv")
    )
```

Effectively, there needs to be a check to see where the file is called from.  

Let me know if you want to make the updates here or I can go ahead and make a new PR.

In the longer term, I think it might be worth exploring some different options.  I'm a little worried that some of these symlinks and the existing path handling might be problematic for packaging and portability.

applications/gfs_init.py
```diff
@@ -20,10 +20,10 @@ def main():
     os.makedirs(config["predict"]["initial_condition_path"], exist_ok=True)
     base_path = os.path.abspath(os.path.dirname(__file__))
     credit_grid = xr.open_dataset(
```
kafitzgerald, line 26:
```suggestion
    if os.path.basename(os.path.abspath(os.path.join(base_path, os.pardir))) == "credit":
        metadata_path = os.path.join(base_path, os.pardir, "metadata")
    else:
        metadata_path = os.path.join(base_path, os.pardir, "credit", "metadata")
    credit_grid = xr.open_dataset(
```

applications/gfs_init.py
```diff
@@ -20,10 +20,10 @@ def main():
...
-        os.path.join(base_path, "../metadata/ERA5_Lev_Info.nc")
+        os.path.join(base_path, "../credit/metadata/ERA5_Lev_Info.nc")
```
kafitzgerald:
```suggestion
        os.path.join(metadata_path, "ERA5_Lev_Info.nc")
```

applications/gfs_init.py
```diff
@@ -20,10 +20,10 @@ def main():
...
     )
     model_levels = pd.read_csv(
-        join(base_path, "../metadata/L137_model_level_indices.csv")
+        join(base_path, "../credit/metadata/L137_model_level_indices.csv")
```
kafitzgerald:
```suggestion
        os.path.join(metadata_path, "L137_model_level_indices.csv")
```

kafitzgerald (commented):
Alright, these changes should allow for both the CLI and script to work and make it a bit more robust.  I was trying to avoid editing your branch directly, but let me know if that's what you'd prefer as opposed to the suggested changes.  Open to a different approach as well if you have preferences.

And I did test this locally for both the CLI and script.  Ideally I think we'd expand the test suite to cover this case, but I think that can wait for the moment.

djgagne (approved):
That looks correct to me. I will make a similar update to gefs_init.py for an upcoming pull request.

PR #217, opened 2025-08-30
Compatibility with Zarr v3

### LINKED ISSUE #216: Lack of support for consolidated metadata in Zarr v3 is causing problems with file reads in CREDIT
Some of the file reads require consolidated metadata for Zarr files, which hasn't been fully implemented in Zarr v3 ([see here for more info](https://zarr.readthedocs.io/en/main/user-guide/consolidated_metadata.html)).

This causing problems reading files in `realtime_rollout.py`.

Downgrading to Zarr<3 should fix the issue for the time being and we may want to consider pinning this in the dependency specification file(s) for now.

We could also look into adapting the code so it's more robust.  I suspect consolidated metadata is something that's desirable in a lot of cases though.

kafitzgerald:
Removes the `consolidated=True` parameter from `open_zarr` for compatibility with Zarr v3 (this setting does not work with Zarr v3 files).

Since version 0.19 of Xarray ([released in 2021](https://docs.xarray.dev/en/v0.19.0/whats-new.html?highlight=release%20notes#new-features)) the default will try to read consolidated metadata first falling back to reading unconsolidated metadata if that fails.  This should be more robust overall and compatible with Zarr v3. 

Closes #216

### DIFF
--- a/credit/data.py
+++ b/credit/data.py
@@ -151,7 +151,7 @@ def get_forward_data(filename) -> xr.Dataset:
     if filename[-3:] == ".nc" or filename[-4:] == ".nc4":
         dataset = xr.open_dataset(filename)
     else:
-        dataset = xr.open_zarr(filename, consolidated=True)
+        dataset = xr.open_zarr(filename)
     return dataset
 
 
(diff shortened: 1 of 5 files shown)


djgagne (approved):
Change looks good to me.

djgagne:
@kafitzgerald Any idea why Github Actions doesn't run on your fork of miles-credit during the PR? I've had this happen before with external contributions from forks rather than internal branches and would love to know how to fix this.

PR #220, opened 2025-09-06
Updates to gefs_init and mean sea level pressure calculations

djgagne:
I have updated gefs_init and interp.py to address some issues that have been coming up in the AI Weather Quest. 
* gefs_init now checks if gefs files have been already downloaded before downloading them from google cloud. I still need to add a file size or checksum comparison to catch partially downloaded files, but this should prevent users from having to manually delete the directory of a gefs run when they want to run the script again. GEFS init also now regrids the GEFS height field by default, so you can access the surface height (not geopotential height, need to multiply by 9.81). 
* When regridding surface pressure, I have the model take the log of surface pressure, regrid, and then exp the regridded values. This should reduce artifacts from exponential variation of surface pressure across fine grid cells and make it a linear combination of values.
* I updated the original mean sea level pressure calculation. I think my formula was not written correctly, resulting in some way off values. 
* I also created a mean_sea_level_pressure_simple function that only uses surface pressure, lowest model level temperature, and surface geopotential. It's within 400 Pa of the corrected mslp function and is a lot simpler to run. I would recommend trying both ones out on some of our different datasets. It looked good for the raw GEFS data at least.
* `temp_height` the height of the temperature level being used for the MSLP reduction, is now adjustable from the config file with the argument `mslp_temp_height` under where you put your other pressure interpolation arguments.

### DIFF
--- a/credit/gefs.py
+++ b/credit/gefs.py
@@ -3,7 +3,7 @@ import numpy as np
 import gcsfs
 from tqdm import tqdm
 import pandas as pd
-from os.path import join
+from os.path import join, exists
 from scipy.sparse import csr_matrix
 import logging
 import yaml
@@ -24,7 +24,10 @@ def download_gefs_run(init_date_str, out_path, n_pert_members=30):
         member_path = join(ens_path, member)
         out_member_path = join(out_path, init_date_path, member)
         if fs.exists(member_path):
-            fs.get(member_path, out_member_path, recursive=True)
+            member_files = fs.ls(member_path)
+            for member_file in member_files:
+                if not exists(join(out_member_path, member_file.split("/")[-1])):
+                    fs.get(member_file, out_member_path)
     return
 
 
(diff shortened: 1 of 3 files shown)


credit/gefs.py
```diff
@@ -24,7 +24,10 @@ def download_gefs_run(init_date_str, out_path, n_pert_members=30):
         member_path = join(ens_path, member)
         out_member_path = join(out_path, init_date_path, member)
         if fs.exists(member_path):
-            fs.get(member_path, out_member_path, recursive=True)
+            member_files = fs.ls(member_path)
+            for member_file in member_files:
```
charlie-becker: When I tried running this, It does not create the `out_member_path` and downloads the files as the "directory" name which causes an error. The path either needs to be created before downloading or a "/" needs to be added to the path.

credit/gefs.py
```diff
@@ -343,6 +372,8 @@ def process_member(
     print(member + ": Regrid")
     regrid_ds = regrid_member(member_tiles, weight_file)
     regrid_ds = combine_microphysics_terms(regrid_ds)
+    out_file = f"gefs_cam_grid_native_vert_{member}.nc"
+    regrid_ds.to_netcdf(join(out_path, out_file))
     print(member + ": Interpolate vertical levels")
```
charlie-becker: Is there a specific reason to save out the native grid? The files are almost twice as large as the interpolated ones and ends up being quite a bit of storage for a single init and all 31 members. Maybe make it an option to save out the native grid?

djgagne: I was saving it out to debug. Removed now.

credit/gefs.py
djgagne: I fixed this issue. New one is more robust.

PR #224, opened 2025-09-17
Add regional model / LES model loss functions to `credit.losses` 

yingkaisha:
TBD

PR #226, opened 2025-09-18
Support for training Samudra (MOM6 dataset) within CREDIT

jsschreck:


jsschreck:
Updated crossformer to support multi-state prediction. New Samurda data loader. The remaining refactor of data + trainer will come in a later PR. If you python the new dataset it will run a simple test.

credit/ocean/samudra_data.py
```diff
@@ -0,0 +1,319 @@
...
+        if var.endswith("_anomalies"):
+            base_var = var.replace("_anomalies", "")
+            if var not in data.variables and base_var in data.variables:
+                logging.info(f"Computing anomalies for {base_var}")
+                data = compute_anomalies(data, base_var)
+
+    return data, data_mean, data_std
+
+
+class Normalize:
```
djgagne, line 178: We should move the Samudra Normalize class to transforms once Kyle's PR #224 is merged.

credit/trainers/trainer_om4_samudra.py
```diff
@@ -0,0 +1,682 @@
...
+                        # concat on var dimension
+                        y = torch.cat((y, y_diag_batch), dim=1)
+
+                    # --------------------------------------------- #
+                    # clamp
+                    if flag_clamp:
+                        y = torch.clamp(y, min=clamp_min, max=clamp_max)
+
+                    with torch.autocast(enabled=amp, device_type="cuda"):
+                        loss = criterion(y.to(y_pred.dtype) * wet_mask, y_pred).mean()
```
djgagne, line 275: I wonder if having a bunch of 0-0 pairs in the loss might be overly reducing the magnitude of the loss and thus the gradient by inflating N? How does the original Samudra loss handle this issue? If you wanted to still make the loss more generic, I would rewrite as 
```
# Assumes wet_mask > 0 is where the ocean points are
y_wet = y.ravel()[wet_mask > 0]
y_pred_wet = y_pred.ravel()[wet_mask > 0]
loss = criterion(y_wet, y_pred_wet).mean()
```

credit/ocean/samudra_data.py
jsschreck: yeah ill do it as soon as we merge his PR -- should be a quick one.

credit/trainers/trainer_om4_samudra.py
jsschreck: I do do this already just not right at the loss calculation step.

PR #230, opened 2025-09-19
updated era5_multistep_batcher.py to work with cftime

kjmayer:


djgagne (approved):
Changes look good to me. Thank you for adding these checks.

PR #231, opened 2025-09-19
Update create_derecho_env.sh to use conda for torch and remove erroneous NCCL

benkirk:


benkirk:
This also fixed `MASTER_ADDR` and `MASTER_PORT`  when running under MPI these do not need to be specified.  However, they must be broadcast and consistent across all the MPI ranks.

djgagne (approved):
Changes look good to me. Thank you for fixing the erroneous NCCL.

PR #239, opened 2025-10-01
Small bug fix

charlie-becker:
Small bug fix for saving output without interpolation. Bug caused failures when running without interpolating output (`conf["predict"]["interp"]`).


kafitzgerald (approved):

kafitzgerald:
FWIW, I was able to reproduce this issue and verify that the change does indeed fix the (in progress) test case.

PR #258, opened 2025-11-18
Bugfix on the application of `credit.transforms` to `rollout_realtime.py` 

yingkaisha:
Added upper air and surface variable checking logic in `ToTensor_ERA5_and_Forcing`, so `credit.transform` can run without upper air variables in its input xr.dataset.

### DIFF
--- a/credit/transforms/transforms_global.py
+++ b/credit/transforms/transforms_global.py
@@ -473,21 +473,11 @@ class ToTensor_ERA5_and_Forcing:
         self.for_len = int(conf["data"]["forecast_len"])
 
         # identify the existence of other variables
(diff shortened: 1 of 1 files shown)


sgpearse:
> @sgpearse I've never seen that during my use.

@yingkaisha Thank you again for your fast help.  My question is this: how does your configuration differ from mine?

I'm trying to deploy a simple CI/CD test system for miles-credit.  For that, I need to know how these tests are executed.

If you would kindly point me to one of your configuration files, I will try it as well.

sgpearse (approved):
Thanks for the fix @yingkaisha.  Memory leak appears to be a separate issue.  We can figure out what the CI/CD systems need to test on another thread.

PR #263, opened 2025-12-08
Crossformer refactor

jsschreck:
In short, we have a new directory inside credit.models called wxformer. All scripts with crossformer in models/ have been moved here, along with two new scripts to support the new ensemble model during inference. 

wxformer.py will also get Will's camulator additions using the new Upsampling blocks with new config option for using that for V2 models.


credit/models/wxformer/stochastic_decomposition_layer.py
```diff
@@ -0,0 +1,47 @@
...
+            noise (torch.Tensor): The latent noise tensor (batch, noise_dim), used for modulating the injected noise.
+
+        Returns:
+            torch.Tensor: The feature map with injected noise.
+        """
+
+        batch, channels, height, width = feature_map.shape
+
+        # Generate per-pixel, per-channel noise
+        pixel_noise = self.noise_factor * torch.randn(batch, channels, height, width, device=feature_map.device)
```
djgagne, line 40: This aspect of the noise looks irreproducible currently. How is the seeding done, or is there a way to enable passing in red noise for the pixel noise?

credit/models/wxformer/stochastic_decomposition_layer.py
```diff
@@ -0,0 +1,47 @@
+import torch.nn as nn
```
djgagne, line 1: Rename module to be consistent with the paper.

credit/models/wxformer/sdl_inference_wrapper.py
```diff
@@ -0,0 +1,1152 @@
...
+        del wrapper._stored_latents[scaled_name]
+    
+    logging.info("\n" + "="*70)
+    logging.info("✓ All experiments complete!")
+    logging.info("="*70 + "\n")
+    
+    return results
+
+
+if __name__ == "__main__":
```
djgagne, line 995: Can we move the actual application piece to the applications directory rather than keeping in a library file?

djgagne (changes requested):
I added an example test, and it would be good to build on that with more unit tests for the new wxformer ensemble stuff. Please also rename files and functions/methods to be consistent with the new paper.

credit/models/wxformer/stochastic_decomposition_layer.py
jsschreck: Yes, we have to use seed_everything method to get this part to replicate. But it does work. When we save the latent sutff in the wrapper, that is post-the random bit here.

credit/models/wxformer/stochastic_decomposition_layer.py
jsschreck: Done

djgagne (approved):
Thank you for adding the tests and name changes. I am approving and merging.

PR #265, opened 2025-12-10
New Distributed Sampler and ERA5 DataLoader

charlie-becker:
This adds a new custom distributed sampler that can handle single and multi-step training over any batch size and number of GPUs. Furthermore, the data loader can utilize multiple cpus for data loading through the torch data loader. 

The data loader is now self contained and only requires the specific `ERA5Dataset()`, the `DistributedMultiStepBatchSampler()`, and a configuration file.

There are some format changes necessary to the `data` key of the configuration file. The main change is that there is a specific `source` key which identifies the specific source torch dataset to utilize. This ERA5 dataset is the first of these to be implemented. The next addition to this will be a simple wrapper to combine the different source datasets.

The sampler and dataset require a `sampling_modes` list which contains the type of data to fetch. This is useful to discriminate between training and inference or single step and multi-step training. The current modes are the following:

- `init`: static data + prognostic data + dynamic forcing data (training x, or inference at time=0)
- `forcing`: dynamic forcing only (multi-step or autoregressive rollouts)
- `y`: target data (init data at t+1, without reloading static data)
- `stop`: flag to break the inference / training loop

A few other notes:

1. When using this distributed sampler, you must supply the sampler to the torch data loader as `batch_sampler=` and **not** `sampler=`
2.  The number of GPUs is in the sampler is controlled by `num_replicas=` and the ID as `rank=`
3. The number of CPUs to parallelize the Data Loader is `num_workers=`
4. The dataset returns a dictionary which includes relevant tensors: `static`, `prognostic`, `dynamic_forcing`
5. It also includes metadata: `time`, `stop_forecast`, and `mode`. The dataloader cannot return a pandas time stamp so it must be returned as a time integer which can easily and directly be converted over with `pd.Timestamp()`


credit/datasets/era5.py
```diff
@@ -0,0 +1,195 @@
...
+import pandas as pd
+import numpy as np
+import xarray as xr
+import torch
+import itertools
+
+logger = logging.getLogger(__name__)
+
+
+class ERA5Dataset(Dataset):
```
negin513, line 18: `Dataset` is referenced in class definition but never imported... 

Either add:
```
from torch.utils.data import Dataset, DataLoader
```

or update this to use `torch.utils.data.Dataset`....

credit/datasets/era5.py
```diff
@@ -0,0 +1,195 @@
...
+
+        self.init_times = self._timestamps()
+        self.years = [str(y) for y in self.init_times.year]
+
+        self.file_dict = {}
+        self.var_dict = {}
+
+        for field_type, d in config['source'][self.source_name].items():
+            if isinstance(d, dict):
+                files = sorted(glob(d.get("path", "")))
```
negin513, line 83: `glob` is not imported. adding `from glob import glob` should fix....

charlie-becker:
I updated the dataloader and sampler a bit to match the more traditional pyTorch Dataloader API.

1. Removed the sampling mode. The sampler now returns the index and uses it's position to determine if initial prognostic and static data need to be loaded
2. Added a `return_target` argument, which when True will also output target prognostics and `t + dt`. I still need to add diagnostics to this.
3. Can now handle cftime or pandas datetime
4. Returns metadata. This could use refinement, as it highly dependent on the data structure (level info, variable names, etc.). It's also worth considering whether we should return metadata for each batch, or only return unique data (really only time step) with the batch and make the other metadata an attribute of the dataset. 

Next steps:
1. add predicted diagnostics to target output data
2. Add transforms 
3. unit testing
4. clean metadata

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+
+class ERA5Dataset(Dataset):
+    """ Pytorch Dataset for processed ERA5 data. Relies on a configuration dictionary to define:
+
+            1) 2D / 3D variables
+            2) Start, End and Frequency of Datetimes
+            3) path to glob for the data
+
+            4) Example YAML Format:
+                data:
```
djgagne, line 29: Please use Markdown or restructuredtext syntax to separate out code or yaml blocks from the rest of the text so that sphinx can render the docstring properly.

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+        # print(return_data['metadata'].items())
+
+        return return_data
+
+    def _open_ds_extract_fields(self, field_type, t, return_data, is_target=False):
+
+        """ opens the dataset, reshapes and concats the variables into an np array, packs it into the return dict if the data exists. """
+
+        if self.file_dict[field_type]:
+            ds = xr.open_dataset(self.file_dict[field_type][t.year])
```
djgagne: Please please put the open_dataset operation within a with block. If you don't, memory leaks will follow, especially for accessing netCDF files. Also, unless things have changed recently with xarray, open_dataset doesn't work with zarr files unless you explicitly set the engine parameter to zarr. You should add some logic here to check for that.

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+        self._add_metadata(return_data, t, t_target)
+        # print(return_data['metadata'].items())
+
+        return return_data
+
+    def _open_ds_extract_fields(self, field_type, t, return_data, is_target=False):
+
+        """ opens the dataset, reshapes and concats the variables into an np array, packs it into the return dict if the data exists. """
+
+        if self.file_dict[field_type]:
```
djgagne, line 146: Can you raise some kind of exception if the field_type is not within file_dict?

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+            ds = xr.open_dataset(self.file_dict[field_type][t.year])
+            if isinstance(z.time[0].item(), cftime.datetime):
+                t = self._convert_cf_time(t)
+
+            if field_type != "static":
+                ds = ds.sel(time=t)
```
djgagne: Can you use a different variable name when you're subsetting the data rather than overwriting ds? It's easier to keep track of when the data scope changes if you do that even if it's a bit more verbose.

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+
+            ds = ds[self.var_dict[field_type]["vars_3D"] + self.var_dict[field_type]["vars_2D"]]
```
djgagne: This line seems unnecessary.

credit/datasets/era5.py
```diff
@@ -0,0 +1,223 @@
...
+        if self.return_target:
+            self._open_ds_extract_fields("prognostic", t_target, return_data, is_target=True)
+
+        self._add_metadata(return_data, t, t_target)
+        # print(return_data['metadata'].items())
+
+        return return_data
+
+    def _open_ds_extract_fields(self, field_type, t, return_data, is_target=False):
+
```
djgagne, line 143: There shouldn't be any empty lines between the function definition and the docstring.

credit/samplers.py
```diff
@@ -0,0 +1,82 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class MultiStepBatchSamplerSubset(Sampler):
+
+    def __init__(self, dataset: Dataset, batch_size: int) -> None:
+        """
+        The dataset is required to have attributes:
+         - init_times of valid init times, each time object compatible with:
```
djgagne, line 16: I would use a * rather than a minus for a bulleted list for formatting purposes.

credit/samplers.py
```diff
@@ -0,0 +1,82 @@
...
+         - timestep, the timestep to sample with
+        This sampler will draw multistep batches with init indices drawn from index_subset.
+        if index_subset=None, it will draw init indices from the entire dataset.
+        taking advantage of DistributedSampler class code with this dataset.
+        can be used on its own as a non-distributed sampler with index_subset=None
+        Args:
```
djgagne, line 23: There should be a blank line between the doc text and Args.

djgagne:
The ruff linter flagged more errors with undefined names and imports. @charlie-becker are you using an IDE like VSCode or PyCharm to write the code? It should be flagging these issues well before the linter catches them.

I also recommend making sure you have pre-commit hooks activated wherever you are committing code, both on your local machine and Casper/Derecho. After building the CREDIT environment and activating it, go to the miles-credit directory and run the following commands
```
pip install pre-commit
pre-commit install
```
Then, ruff will run on files you have edited for a given commit, fix any safe-to-fix formatting errors, and flag anything else you need to modify manually before you are allowed to commit code. I use it, and it has caught many formatting and variable issues for me. 

```
F821 Undefined name `z`
   --> credit/datasets/era5.py:145:27
    |
143 |         if self.file_dict[field_type]:
144 |             ds = xr.open_dataset(self.file_dict[field_type][t.year])
145 |             if isinstance(z.time[0].item(), cftime.datetime):
    |                           ^
146 |                 t = self._convert_cf_time(t)
    |

F821 Undefined name `cftime`
   --> credit/datasets/era5.py:145:45
    |
143 |         if self.file_dict[field_type]:
144 |             ds = xr.open_dataset(self.file_dict[field_type][t.year])
145 |             if isinstance(z.time[0].item(), cftime.datetime):
    |                                             ^^^^^^
146 |                 t = self._convert_cf_time(t)
    |

F821 Undefined name `cftime`
   --> credit/datasets/era5.py:201:16
    |
199 |         """ Convert pandas timestamp to cftime """
200 |
201 |         cf_t = cftime.datetime(ts.year, ts.month, ts.day, ts.hour, ts.minute, ts.second, calendar='noleap')
    |                ^^^^^^
202 |
203 |         return cf_t
    |

F821 Undefined name `ERA5Dataset`
   --> credit/datasets/era5.py:215:19
    |
213 |         data_config = config["data"]
214 |
215 |         dataset = ERA5Dataset(config['data'])
    |                   ^^^^^^^^^^^
216 |         sampler = DistributedMultiStepBatchSampler(dataset=dataset, batch_size=4, sampling_modes=['init'] + ['y'] * 1,
217 |                                                    rank=0, num_replicas=1, shuffle=True)
    |
```

credit/samplers.py
```diff
@@ -0,0 +1,82 @@
...
+        self.index_subset = torch.randperm(len(dataset))
+        self.batch_size = batch_size
+        self.num_start_batches = (len(self.index_subset) + self.batch_size - 1) // self.batch_size
+
+    def __len__(self):
+
+        # actual number of init batches
+        return self.num_start_batches
+
+    def __iter__(self):
```
negin513, line 40: IIUC, `DistributedMultiStepBatchSampler` ignores distributed indices here...
  The indices returned by `DistributedSampler.__iter__()` (which partitions data across GPUs by rank) are computed here but never passed to `MultiStepBatchSamplerSubset`, so I was not sure what the intion of calculating them were or if the `DistributedMultiStepBatchSampler` was supposed to be used in another context .
  
MultiStepBatchSamplerSubset  seems generates its own torch.randperm(len(dataset)) at line 29, so every GPU independently iterates over the entire dataset rather than its assigned partition, I think! We should double check!

negin513: I scheduled a quick chat to talk about this and easiest way to test this PR.

djgagne:
Since the main changes in this branch got put in a different one, going ahead and closing.

PR #269, opened 2026-01-09
Streamlines Ruff configuration, adds pre-commit.ci, and applies linting/formatting fixes

### LINKED ISSUE #268: pre-commit is failing
pre-commit (Ruff) is currently failing on main.

I suspect some of this has to do with the lack of visibility of this in CI that was mentioned in stand-up today.

It'd be great to get these failures addressed and hopefully make the results of the checks a bit more visible moving forward.

There are a number of different options for how to approach this last part.  [Pre-commit.ci](https://pre-commit.ci/) is pretty common though and might be worth a look.

kafitzgerald:
Closes #268 

This PR moves all Ruff configuration to `pyproject.toml`, adds pre-commit.ci (to make linting/formatting more visible in CI and streamlined w/ pre-commit), and applies a bunch of linting and formatting fixes.

I suspect we'll also want to remove the Ruff checks from the testing workflow in CI, but left those for now mostly for reference while we norm on configuration options (currently this varies between CI and the other configs).

Setting this as "draft" until we can find a convenient time to discuss.  I know there's some other stuff pending at the moment and this will create a bunch of conflicts.  Will be nice to address though and should help moving forward.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,3 @@
+# .git-blame-ignore-revs
+# Apply linting and formatting
+112443217fe73ab059dd5091e72c6b9ad327b14e


kafitzgerald:
> I am fine with the changes in general. One question I have is what the default number of characters per line should be? Currently it's set at 180, but if we're going to be reformatting a bunch of existing code, should we shrink this back down to 120 or 80 to be PEP 8 compliant or leave it longer and let people use line wrapping or scrolling in their preferred text editor to deal with longer lines?
> 
> More vertically oriented code is easier to stack side by side and to navigate through in command line text editors like vim, but it does sometimes result in some weird formatting choices by the linter. Thoughts?

Thanks for raising this.  I agree it'd be good to discuss.

The 180 comes from the Ruff configuration in the pyproject.toml (which doesn't seem to have been enforced previously, but was there already).

120 is a pretty common default for a lot of text editors, but a lot of (maybe most) Python projects that I'm aware of stick with the default of 88 from [Ruff](https://docs.astral.sh/ruff/rules/line-too-long/)/[Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) so slightly above PEP8's 79.  I've also seen 100 and 120, but less frequently and rarely over that.

I think my preference would be to move to the Ruff default, but I'd like to hear other's thoughts on this as well though (especially if folks have strong opinions).

PR #279, opened 2026-02-12
Fix MPI conda channel issue in create_derecho_env.sh

djgagne:
I made an edit to the create_derecho_env.sh script to add a conda pseudo channel for machine-specific MPI installations, as mentioned in this [article](https://conda-forge.org/docs/user/tipsandtricks/#using-external-message-passing-interface-mpi-libraries). I tested on Derecho and it ran successfully.


charlie-becker (approved):
LG

PR #280, opened 2026-02-12
Update era5.py

charlie-becker:
I updated the metadata scheme to better integrate with downstream tasks.

I ran into issues trying to preserve the static portions of the metadata when collating the batch. Either the metadata gets duplicated and returned as a list of individual tensors, or if I try to make it a tensor attribute, it vanishes once the batch is concatenated since it's technically a new tensor.

As an alternative, I created a simple method `_build_var_metadata` that builds the variable metadata directly from the config and stores it as an attribute of the torch dataset object. You can then either add this to the batch as as attribute, or pass it along separately to a pre-block layer. An additional bonus is that this same method can be used to create the variable names when fitting a bridgescaler object so we can later match them up during a transform.

### DIFF
--- a/config/era5_new_data_config.yaml
+++ b/config/era5_new_data_config.yaml
@@ -1,21 +1,23 @@
 data:
     source:
         ERA5:
-            prognostic:
-                # upper-air variables
-                vars_3D: ['U','V','T','Q']
-                vars_2D: ['SP','t2m', 'V500','U500','T500','Z500','Q500']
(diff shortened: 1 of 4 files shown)


charlie-becker:
I also expanded the config to include `level_coord:` and `levels:`

`level_coord` is the coordinate used to select the levels in xarray (not index) which will allow flexibility between pressure coordinates or model level coordinates.

`levels` are the actual vertical level values for above.

This is ready for review.

djgagne (approved):
Updates look good and pass all tests. I had the precommit bot fix some formatting in the PR, so now all tests pass and everything looks good.

PR #281, opened 2026-02-12
two fixes -- mpi + torch version for derecho

negin513:


djgagne (approved):
These changes look good to me. We can make the pbs.py changes in a separate PR.

PR #283, opened 2026-02-26
Preblocks

charlie-becker:
I've added two pre-blocks as a start to foster more discussion on exactly how we want to integrate the pre-processing (pre-block chain). The two I've added are:   

1. Bridgescaler pre-block: Pre-block that takes in a saved bridgescaler object and transforms any subset of variables that exists within the scaler object. It also supports inverse transforms which can be used as a post-block.
2. Regrid pre-block: Takes in a weights file that is generated from the ESMF library (not xESMF). Requires users to run `grid.py` on both a source grid and destination grid first to generate the SCRIP file format needed by `ESMF_RegridWeightGen`. This preblock utilizes the `torch.sparse_coo_tensor` for the matrix multiplication. It takes a short time (100 ms) to load the weights on the first pass, but GPU weight application is on the order of a couple of ms. 

I have independently tested these, but I will add unit tests soon. Comments welcome.

### DIFF
--- a/credit/preblock/regrid.py
+++ b/credit/preblock/regrid.py
@@ -23,7 +23,18 @@ class Regrid(nn.Module):
             weights = grid_weights["S"].values
             n_a = grid_weights.sizes["n_a"]
             n_b = grid_weights.sizes["n_b"]
-            dst_shape = grid_weights["dst_grid_dims"].values[::-1]  ## should probably chek to see if this is necessary
+            raw_dst = grid_weights["dst_grid_dims"].values
+            if len(raw_dst) == 2:
+                # Structured weight file: dst_grid_dims = [nlon, nlat]; reverse to [nlat, nlon]
+                dst_shape = raw_dst[::-1]
+            elif reshape_to_xy:
+                # Regular unstructured: infer 2D shape from unique destination center coords
+                n_lat = np.unique(grid_weights["yc_b"].values).size
+                n_lon = np.unique(grid_weights["xc_b"].values).size
+                dst_shape = np.array([n_lat, n_lon])
+            else:
+                # Irregular unstructured: output stays flat, no 2D shape needed
+                dst_shape = None
 
         self.reshape_to_xy = reshape_to_xy
         self.flip_axis = flip_axis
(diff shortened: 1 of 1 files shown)


credit/preblocks.py
```diff
@@ -0,0 +1,74 @@
...
+import numpy as np
+import torch
+import torch.nn as nn
+
+
+class Scaler(nn.Module):
+    """
+    Scaling layer using a bridgescaler object. Supports transform and its inverse.
+    """
+    def __init__(self, scaler, inverse=False):
```
djgagne: scaler should be a path to a scaler file rather than a scaler object. The file should be read in the init and then be applied in the forward method.

credit/preblocks.py
```diff
@@ -0,0 +1,74 @@
...
+
+        self._W = W
+        self._W_device = device
+        return W
+
+    def forward(self, x):
+
+        device = x.device
+        W = self._get_W(device)
+        x = torch.flip(x, dims=[-2])
```
djgagne: Whether or not to flip the grid should be a keyword arg in the init rather than automatically applied.

credit/preblocks.py
```diff
@@ -0,0 +1,74 @@
...
+        lead_shape = x.shape[:-2]
+        x_flat = x.reshape(-1, self.n_a).T
+        y_flat = torch.sparse.mm(W, x_flat).T
+
+        ny, nx = self.dst_shape
+
+        return y_flat.reshape(*lead_shape, ny, nx)
```
djgagne: Another generalization option is to support irregular grids by not reshaping at the end to a rectilinear grid. I would also make whether or not to reshape to rectilinear an option for the cell.

djgagne:
@charlie-becker 
1. I expect there to be a lot of different kinds of preblocks eventually, and I want to avoid the problem of 1 giant file. Similar preblocks can be grouped together into the same file, like different regridders or scaler layers can be in one file. 
2. The concatenation preblock needs to be at the end because modifying tensors in place breaks backpropagation. It's more memory efficient for doing transforms to have each variable in a separate tensor and to carry the metadata along because some preblocks will be applied only to a subset of variables. For example, you will want to use a conservative or bilinear regrid for continuous values but will want to use nearest neighbor for discrete fields like land-sea mask.

djgagne (approved):
Changes look good to me! Thank you for adding tests.

PR #286, opened 2026-03-05
CI check to run inference on each Pull Request

sgpearse:
This PR launches a self hosted runner on CIRRUS to run inference on each pull request.  6 hours are forecasted from the start of the 6 hour window of the submission time (IE if the PR is submitted on Mar 5 05:40:05 2026, inference will be run from March 5 00:00:00 to 06:00:00).

Results can be found in the "CI check to run inference on each Pull Request" test for each PR.

### DIFF
--- a/tests/CI_CIRRUS/Dockerfile
+++ b/tests/CI_CIRRUS/Dockerfile
@@ -1,0 +1,86 @@
+FROM nvidia/cuda:12.9.1-base-ubuntu24.04
+
+USER root
+
+ENV PYTHONBUFFERED=1
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CONDA_DIR=/opt/conda
+ENV PATH=$CONDA_DIR/bin:$PATH
+#ENV MINIFORGE_VERSION=23.3.1-0
+ENV MINIFORGE_VERSION=25.3.0-3
+
+# Install system dependencies
+RUN apt-get update && apt-get install -y \
+    curl git wget bzip2 ca-certificates vim \
+    libglib2.0-0 libxext6 libsm6 libxrender1 \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# Install Miniforge (Conda)
+RUN wget https://github.com/conda-forge/miniforge/releases/download/${MINIFORGE_VERSION}/Miniforge3-Linux-x86_64.sh -O /tmp/miniforge.sh && \
+    bash /tmp/miniforge.sh -b -p ${CONDA_DIR} && \
+    rm /tmp/miniforge.sh
+
+RUN echo ". ${CONDA_DIR}/etc/profile.d/conda.sh" >> /etc/bash.bashrc
+
+# Create Conda environment with Python + pysteps + build tools
+RUN conda create -n credit python=3.11 -c conda-forge -y && \
+    conda run -n credit conda install -c conda-forge pysteps pip setuptools wheel esmf esmpy yaml -y && \
+    conda clean -afy && \
+    conda run -n credit python -m pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu121 && \
+#    conda install -n credit -y -c conda-forge "hdf5=*=nompi_*" "libnetcdf=*=nompi_*" "netcdf4=*=nompi_*" && \
+    conda run -n credit conda install -n credit -y -c conda-forge "hdf5=*=nompi_*" "libnetcdf=*=nompi_*" "netcdf4=*=nompi_*" && \
+    conda run -n credit python -m pip install xesmf
+
+RUN conda init bash
+
+# Install PyTorch with CUDA support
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-cirrus.yml
```diff
@@ -0,0 +1,50 @@
+name: Run inference with Pull Request
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
kafitzgerald, line 5: I wonder if it might be worth setting this up to be triggered by adding a label to the PR.  That way we have a little more control over when it runs as opposed to all PRs by default.  Happy to point out some examples if that would be helpful (e.g. Xarray has this set up for their benchmarking workflow).

Branch constraints might also be worth discussing.

.github/workflows/pullRequest.yml
```diff
@@ -0,0 +1,50 @@
...
+     AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+     AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+     AWS_ENDPOINT_URL: ${{ secrets.AWS_ENDPOINT_URL }}
+     AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+    steps: 
+    - name: Debug permissions...
+      run: |
+        id
+        ls -ld /__w /__w/_temp /__w/_temp/_runner_file_commands || true
+        touch /__w/_temp/testfile || echo "Cannot write"
```
kafitzgerald: Is this something worth keeping in at this point?

.github/workflows/pullRequest.yml
```diff
@@ -0,0 +1,50 @@
...
+    - name: checkout code
+      uses: actions/checkout@v4
+    - name: Fetch and checkout PR head commit
+      run: |
+        git config --global --add safe.directory /__w/miles-credit/miles-credit
+        git fetch origin ${{ github.event.pull_request.head.sha }}
+        git checkout ${{ github.event.pull_request.head.sha }}
```
kafitzgerald: I might be missing something here, but I'm curious why we can't use the code from actions/checkout above.

kafitzgerald (commented):
Nice to see this working!

A few questions and suggestions inline.  

I'd also probably lean toward putting the model config and shell script elsewhere (as opposed to directly in the workflows directory), but am not sure offhand where the best place might be.  Maybe in a "ci" directory at the base of the repo?

.github/workflows/ci-cirrus.yml
sgpearse: That sounds good if the team prefers to do that.

.github/workflows/pullRequest.yml
sgpearse: Nope.  Removed!

.github/workflows/pullRequest.yml
sgpearse: Thanks for the catch.  Removed.

.github/workflows/ci-cirrus.yml
kafitzgerald: This is also something we could look into later as well if needed.

I don't think volume of PRs is particularly high here, but don't have a good feel for if we'd want to run on PRs to all branches vs. just main and/or if we might run into issues w/ resources e.g. with tests running on new commits to PRs, etc..

I'm not super familiar with the self hosted runners and/or what resources are available on CIRRUS, so was just trying to think ahead a bit.

sgpearse: We would run tests on all branches pushed to the VAPOR repository.  That way we would know if we broke something before submitting a PR.  I don't think this load would negatively impact CIRRUS.  Kevin mentioned that it's great to see the compute hours for its runners increasing yesterday, so it sounds like they would encourage this.

.github/workflows/pullRequest.yml
```diff
@@ -0,0 +1,41 @@
...
+        git config --global --add safe.directory /__w/miles-credit/miles-credit
+        echo "Running debug job for PR on self-hosted runner"
+        git log -1
+        chmod -R 777 ./.github/workflows/*
+        ./.github/workflows/dateChange.sh /__w/miles-credit/miles-credit/.github/workflows/model_predict_CI.yml
+        export PATH=/home/ubuntu/.local/bin:$PATH
+        conda run -n credit python -u /__w/miles-credit/miles-credit/applications/gfs_init.py -c /__w/miles-credit/miles-credit/.github/workflows/model_predict_CI.yml
+        conda run -n credit python -u /__w/miles-credit/miles-credit/applications/rollout_realtime.py -c /__w/miles-credit/miles-credit/.github/workflows/model_predict_CI.yml
+        du -h /output
+    #- name: Upload the gfs_init.py netCDF output files to CIRRUS storage
```
djgagne: Can you remove the commented out section from the file?

PR #287, opened 2026-03-06
Coupled camulator

WillyChap:
Adding coupling capability to CESM2.1.5 ocean + instructions via File I/O 

This is coupled with my CESM2.1.5 instructions here: https://github.com/WillyChap/CESM/tree/release-cesm2.1.5-camulator 
that plugs CAMulator into the DATM of CESM.

### DIFF
--- a/.github/workflows/ci-cirrus.yml
+++ b/.github/workflows/ci-cirrus.yml
@@ -1,0 +1,41 @@
+name: Run inference with Pull Request
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+jobs: 
+  build: 
+    runs-on: gha-runner-miles-credit1
+    container:
+      image: docker.io/sgpearse/cirrustest:1.3.20.root.1
+      options: --user 0
+    env:
+     GITHUB_TOKEN: ${{ github.token }}
+     AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+     AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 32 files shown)


climate/camulator_server.py
```diff
@@ -0,0 +1,1256 @@
...
+        TS_cam = accessor_output.get_state_var(prediction_cpu, "TS")[0, 0, 0].numpy()
+        # FLNSD = ε σ T⁴ + FLNS/DT  (FLNS accumulated is net outward, negative for typical conditions)
+        _sb_term = 0.99 * 5.670374419e-8 * TS_cam**4
+        _flns_term = FLNS_cam / DT_SEC
+        FLNSD_cam = _sb_term + _flns_term
+        # Optional FLNSD diagnostic: area-weighted SB-term comparison across TS/TREFHT/Tbot.
+        # Enable with --flnsd_diag. Validated 2026-03: TS gives ~331 W/m² vs CAM6 ref 336.9 W/m².
+        if args.flnsd_diag:
+            _cam_w = _cam_lat_w if not cam_flip else _cam_lat_w[::-1]
+            _wmean = lambda f: np.average(f, weights=np.broadcast_to(_cam_w, f.shape))
```
djgagne: Can you replace the lambda function with `def _wmean(f): np.average(f, weights=np.broadcast_to(_cam_w, f.shape))`. The linter is complaining about it.

djgagne (changes requested):
This new capability to couple Camulator with CESM2 is awesome! I'd love to see some results next time we meet up. 

I ran the ruff reformatter bot to bring the code in line with the rest of CREDIT. It fixed everything but flagged a lambda function that should be replaced with a regular Python function defined at that scope. See my comment there. It should be an easy fix. Once that is done, I am fine with merging.

PR #291, opened 2026-03-10
Dataset refactor

charlie-becker:
Significant refactor of ERA5 dataset, addition of MRMS dataset, and multi-source dataset (wrapper).

  - `credit/datasets/era5.py` — Refactored ERA5Dataset. Output changes from four concatenated tensors (prognostic, dynamic_forcing, static, diagnostic) to a nested dict of per-variable tensors keyed as era5/{field_type}/{dim}/{varname} (e.g. era5/prognostic/3d/T). All variables have a consistent 4D shape (n_levels, 1, lat, lon) with a singleton level dim for 2D variables, enabling default PyTorch collation without a custom collate_fn.
  
  - `credit/datasets/MRMS.py` — New MRMSDataset with the same output format (mrms/{field_type}/2d/{varname}), supporting local NetCDF/Zarr files and remote AWS S3 streaming. Uses ERA5-style field type semantics
  (prognostic/dynamic_forcing/diagnostic). Config "mode" controls streaming or loading from disk.
  
  - `credit/datasets/mrms_download.py` — Standalone utility for downloading MRMS GRIB2 data from S3 and saving to local NetCDF or Zarr, ready for use with MRMSDataset.
  
  - `credit/datasets/multi_source.py` — MultiSourceDataset wrapper that combines any number of registered source datasets into a single {source_name: sample_dict} output. Computes the timestamp intersection across all sources at init time. Exposes static_metadata aggregated from each source.
  
  - `credit/datasets/_file_utils.py` — Shared strftime-based filename parsing and binary-search file lookup, used by both ERA5Dataset and MRMSDataset (and future datasets). This allows support for any temporally aggregated data scheme when provided the strftime format in the config (e.g. %Y-%m-%d)

  - `credit/samplers.py` — DistributedMultiStepBatchSampler now takes num_forecast_steps as an explicit constructor argument (previously read from dataset.num_forecast_steps). dt is derived internally from dataset.datetimes. This decouples the sampler from dataset internals.
  
  - `credit/datasets/__init__.py` — Added exports for the three new dataset classes.

  **Config convention change**

  - `forecast_len` in conf_era_test.yaml is now base-1 (the value equals the number of forecast steps directly). Legacy configs are unchanged.
  
  - `config/multi_source_data.yaml`: An example configuration file with two datasets (just the data portion). We could consider splitting the configs into three different categories (data, training, inference) in a future refactor. 

**Additional details**

  - MultiSourceDataset is always the entry point, even for single-source configs, for a consistent API. Any additional datasets can be added to the registry which will be collected by the MultiSourceDataset as a nested dictionary. Regridding, scaling, merging, etc. will then be handled by pre-blocks on the GPU. 
  
  - Example output dictionary for the multi-source data (with ERA5 and MRMS):
```
    {
        "era5": {
            "input":    {"era5/prognostic/3d/T": tensor, ...},
            "target":   {"era5/prognostic/3d/T": tensor, ...},  # if return_target=True in dataset (@ t + dt)
            "metadata": {"input_datetime": int, "target_datetime": int},
        },
        "mrms": {
            "input":    {"mrms/prognostic/2d/MultiSensor_QPE_01H_Pass2_00.00": tensor, ...},
            "target":   {"mrms/prognostic/2d/MultiSensor_QPE_01H_Pass2_00.00": tensor, ...},
            "metadata": {"input_datetime": int, "target_datetime": int},
        },
    }
```
  - Static (across batch) metadata is repetitive and funky when collating a batch, so it is currently stored as an attribute of the multi-source dataset:
    
```
    {
       "era5": {"levels": [...,  110, 120, 130, 136, 137], "datetime_fmt": "unix_ns"},
                    "mrms": {"datetime_fmt": "unix_ns"},
    }
```
  - Static data is not supported by MRMS, and is currently only supported by the `ERA5Dataset` with it's own path. We could consider making a separate dataset for statics all together (which is non-trivial with the current code), so it's recommended at this point to merge all static data into a single file within the ERA5 portion of the config. This should be temporary, as future use cases may not need ERA5 data but still include static data. Currently, you could work around that by setting all of the other data types in ERA5 to `null`. 
  
  - `scripts/multi_source_example.py`: Simple example demonstrating multi-source usage.


credit/datasets/multi_source.py
```diff
@@ -0,0 +1,161 @@
...
+
+                {"input": {...}, "target": {...}, "metadata": {...}}
+        """
+        return {name: ds[args] for name, ds in self.datasets.items()}
+
+    # ------------------------------------------------------------------
+    # Private helpers
+    # ------------------------------------------------------------------
+
+    def _intersect_timestamps(self) -> pd.DatetimeIndex:
```
djgagne, line 144: For multi source datasets with different time steps, would it make more sense to handle the time matching logic in this class or as a helper method called in the `__getitem__` step? Handling it in __getitem__ would allow users to define how an arbitrary time maps to timesteps that actually have data and whether that is nearest timestep, last timestep, or some average of the two time steps.

djgagne (approved):
I am going ahead and approving since all tests and checks are now passing. era5 and MRMS work at least with default datasets. I expect we will make some additional adjustments, but it would be easier to do those in a future PR. Merging now.

PR #297, opened 2026-03-25
Preblocks

charlie-becker:
## Add config-driven preblock system

Introduces a modular preblock pipeline for applying variable-level transformations to batch dictionaries prior to model ingestion.

### Structure

```
preblocks/
    __init__.py          ← registry, build_preblocks(), apply_preblocks()
    base.py              ← BasePreblock (nn.Module)
    log.py     ← LogTransform
    sqrt.py       ← SqrtTransform
    regrid.py and scaler.py still need updating for config-driven usage
```

### Design

- Batch dicts follow the schema `batch[source][data_type]['source/var_type/var_shape/var_name']`
- Preblocks operate on `input` and `target` data types by default; `metadata` is never modified but always propagated
- `data_types` is configurable per-preblock to support cases where input and target transforms differ
- `target` is silently skipped if absent (inference mode)
- All transforms are out-of-place to preserve autograd compatibility

### Config

Preblocks are defined in the model config and instantiated via `**kwargs` unpacking — adding new args to a transform requires no changes to the registry or instantiation logic:

```yaml
preblocks:
    era5_log_transform:
        type: "log_transform"
        args:
            variables:
                - 'ERA5/prognostic/3d/Q'
            base: 'e'
            eps: 1.0e-8
    era5_sqrt_transform:
        type: "sqrt_transform"
        args:
            variables:
                - 'ERA5/prognostic/3d/U'
                - 'ERA5/prognostic/3d/V'
            data_types:
                - 'input'
```

### Usage

```python
from preblocks import build_preblocks, apply_preblocks

preblocks = build_preblocks(config['preblocks'])

for batch in dataloader:
    batch = apply_preblocks(preblocks, batch)
```

### Adding a new transform

1. Create `preblocks/new_transform.py` subclassing `BasePreblock`
2. Add one import and one registry entry to `__init__.py`

No other changes required.

### DIFF
--- a/credit/preblock/concat.py
+++ b/credit/preblock/concat.py
@@ -2,7 +2,7 @@ import torch
 from credit.preblock.base import BasePreblock
 
 
-class ConcatenateToTensor(BasePreblock):
+class ConcatToTensor(BasePreblock):
     """End-of-chain preblock that concatenates a nested batch dict of tensors
     into a single input tensor (and optionally a target tensor).
 
--- a/credit/preblock/concat.py
+++ b/credit/preblock/concat.py
@@ -1,0 +1,56 @@
+import torch
+from credit.preblock.base import BasePreblock
+
+
+class ConcatenateToTensor(BasePreblock):
+    """End-of-chain preblock that concatenates a nested batch dict of tensors
+    into a single input tensor (and optionally a target tensor).
+
+    Expects a batch dict of the form::
+
+        batch[source][data_type][var_name] -> torch.Tensor
+
+    where tensor shapes are (batch, channel, time, lon, lat) and concatenation
+    is performed along dim=1 (channel). Traversal order follows key insertion
+    order: for each source, all var_names under a data_type are concatenated,
+    then the next source, and so on.
+
+    ``metadata`` keys are passed through as-is (not concatenated).
+
+    Returns either::
+
+        (input_tensor, metadata)                    # if no "target" data_type present
+        (input_tensor, target_tensor, metadata)     # if "target" is present
+
+    Example config::
+
+        type: "concatenate_to_tensor"
+        args: {}
+    """
+
+    def forward(self, batch):
+        input_tensors = []
+        target_tensors = []
+        metadata = {}
+
+        for source, data_types in batch.items():
+            for data_type, variables in data_types.items():
+                if data_type == "metadata":
+                    metadata[source] = variables
+                elif data_type == "input":
+                    for tensor in variables.values():
+                        input_tensors.append(tensor)
+                elif data_type == "target":
+                    for tensor in variables.values():
+                        target_tensors.append(tensor)
+
+        if not input_tensors:
+            raise ValueError("No 'input' tensors found in batch.")
+
+        input_tensor = torch.cat(input_tensors, dim=1)
+
+        if target_tensors:
+            target_tensor = torch.cat(target_tensors, dim=1)
+            return input_tensor, target_tensor, metadata
+
+        return input_tensor, metadata
--- a/credit/preblock/log.py
+++ b/credit/preblock/log.py
@@ -36,11 +36,7 @@ class LogTransform(BasePreblock):
         # Validate data_types at init
         invalid = set(self.data_types) - set(self.VALID_DATA_TYPES)
         if invalid:
-            raise ValueError(
-                f"Invalid data_types {invalid}. "
-                f"Valid options are {self.VALID_DATA_TYPES}. "
-                f"Preblocks never operate on 'metadata'."
-            )
+            raise ValueError(f"Invalid data_types {invalid}. Valid options are {self.VALID_DATA_TYPES}. ")
 
         if base == "e":
             self._log_fn = torch.log
@@ -71,6 +67,6 @@ class LogTransform(BasePreblock):
 
                 x = batch[source][data_type][var_key]
                 # Out-of-place: safe for autograd
-                batch[source][data_type][var_key] = self._log_fn(x + self.eps)
+                batch[source][data_type][var_key] = self._log_fn(x + self.eps) - self._log_fn(self.eps)
 
         return batch
--- a/credit/preblock/log.py
+++ b/credit/preblock/log.py
@@ -1,0 +1,76 @@
+from credit.preblock.base import BasePreblock
+import torch
+
+
+class LogTransform(BasePreblock):
+    """
+    Applies a log transformation to specified variables in a batch dict.
+
+    Expected dict structure:
+        batch[source][data_type]['source/var_type/var_shape/var_name']
+
+    Config example:
+        type: "log_transform"
+        args:
+            variables:
+                - 'ERA5/prognostic/3D/Q'
+            data_types:     # optional, defaults to ['input', 'target']
+                - 'input'
+                - 'target'
+            base: 'e'       # optional, default 'e'. Options: 'e', '2', '10'
+            eps: 1.0e-8     # optional, default 1e-8
+    """
+
+    def __init__(
+        self,
+        variables: list[str],
+        data_types: list[str] = None,
+        base: str = "e",
+        eps: float = 1e-8,
+    ):
+        super().__init__()
+        self.variables = variables
+        self.data_types = data_types or ["input", "target"]
+        self.eps = eps
+
+        # Validate data_types at init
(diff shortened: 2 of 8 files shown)


credit/preblock/regrid.py
```diff
@@ -4,7 +4,7 @@
 import xarray as xr
 
 
-class Regrid(nn.Module):
+class Regridder(nn.Module):
```
djgagne: Need to update your unit test to import Regridder from regrid instead of Regrid.

credit/preblock/log.py
```diff
@@ -0,0 +1,76 @@
...
+
+                # Silent skip: variable not present in this data_type
+                if var_key not in batch[source][data_type]:
+                    print(batch[source][data_type].keys())
+                    print(f"LogTransform: data_type '{data_type}' not found in batch.")
+                    continue
+
+                x = batch[source][data_type][var_key]
+                # Out-of-place: safe for autograd
+                batch[source][data_type][var_key] = self._log_fn(x + self.eps)
```
djgagne: Can you edit the log transform to be `self._log_fn(x + self.eps) - self._log_fn(self.eps)` so that the correct values are preserved?

credit/preblock/sqrt.py
```diff
@@ -0,0 +1,58 @@
...
+
+                # Silent skip: variable not present in this data_type
+                if var_key not in batch[source][data_type]:
+                    print(batch[source][data_type].keys())
+                    print(f"SqrtTransform: data_type '{data_type}' not found in batch.")
+                    continue
+
+                x = batch[source][data_type][var_key]
+                # Out-of-place: safe for autograd
+                batch[source][data_type][var_key] = torch.sqrt(x)
```
djgagne, line 56: Something I've gone back and forth is on whether to rename a variable after the transform with `var_name_sqrt` or `var_name_log` to make it easier to track the transforms? The hard part is that it then makes downstream variable specific things harder to code in.

credit/preblock/scaler.py
```diff
@@ -2,7 +2,7 @@
 from bridgescaler import load_scaler
 
 
-class Scaler(nn.Module):
+class BridgeScaleTransformer(nn.Module):
```
djgagne: We will also need to use Kevin's new [scale_var_dict function](https://github.com/NCAR/bridgescaler/blob/054f10366b7d36b0d0fed0877a359a1f326c85e3/bridgescaler/backend.py#L265) here to apply Bridgescaler transforms on the CREDIT data structure.

credit/preblock/sqrt.py
charlie-becker: I do think it would be difficult to handle downstream in some cases with different variable names. What about adding properties to the meta data describing which variables were modified (and how)?

djgagne (changes requested):

credit/preblock/sqrt.py
djgagne: That's a good idea.

credit/preblock/scaler.py
kevinyang-cky: change to `self.scaler = load_scaler_dict(scaler_path)`

kevinyang-cky: change to `return scale_var_dict(x, self.scaler, "transform")`

kevinyang-cky: change to `return scale_var_dict(x, self.scaler, "transform")`

credit/preblock/scaler.py
```diff
@@ -1,19 +1,73 @@
...
+        self.scaler_path = scaler_path
+        self.scaler = load_scaler_dict(scaler_path)
+
+        # Validate data_types at init
+        invalid = set(self.data_types) - set(self.VALID_DATA_TYPES)
+        if invalid:
+            raise ValueError(f"Invalid data_types {invalid}. Valid options are {self.VALID_DATA_TYPES}. ")
+
+    def forward(self, batch: dict) -> dict:
+        for var_key in self.variables:
```
djgagne: The forward logic is not correct here. Kevin's new [scale_var_dict](https://github.com/NCAR/bridgescaler/blob/054f10366b7d36b0d0fed0877a359a1f326c85e3/bridgescaler/backend.py#L265-L345) function and the new `save_scaler_dict` and `load_scaler_dict` functions handle loading, saving, and applying bridgescaler scalers to nested dictionaries of variables recursively. You should not have to loop over anything.

credit/preblock/scaler.py
```diff
@@ -1,19 +1,73 @@
...
+                - "era5/prognostic/3d/T"
+                - "era5/prognostic/3d/U"
+            method: "transform"
+            data_types:       # optional, defaults to ['input', 'target']
+                - "input"
+                - "target"
     """
 
-    def __init__(self, scaler_path, inverse=False):
+    def __init__(self, scaler_path: str, variables: list[str], method: str, data_types: list[str] = None):
```
djgagne: You can go back to just passing scaler_path because `load_scaler_dict` handles loading in all the nested scaler objects for you since a collection of jsons is also a json.

djgagne (changes requested):
Scaler.py doesn't look like it is interacting with the new bridgescaler code correctly. I also didn't understand it right at first but I think you should be able to pass the full state dictionary data structure to bridgescaler and it will handle all the operations itself.

credit/preblock/scaler.py
charlie-becker: > The forward logic is not correct here. Kevin's new [scale_var_dict](https://github.com/NCAR/bridgescaler/blob/054f10366b7d36b0d0fed0877a359a1f326c85e3/bridgescaler/backend.py#L265-L345) function and the new `save_scaler_dict` and `load_scaler_dict` functions handle loading, saving, and applying bridgescaler scalers to nested dictionaries of variables recursively. You should not have to loop over anything.

@kevinyang-cky @djgagne 

I implemented the loop to potentially handle a use case for a subset of variables of the scaler, and not necessarily use all of them. That way we can can have one "main" scaler object but use it for various training cases. The current dictionary based scaler doesn't support specific sets of variables, correct? Thoughts?

kevinyang-cky: @charlie-becker correct. The current`scale_var_dict` requires the dictionary of scalers to share the same structure as the dictionary of tensor variables. One thought is that one passes the dictionary of scalers containing a subset of variables, and `scale_var_dict` can still take the full dictionary of tensor variables, but only scale the variables specified in the dictionary of scalers. There are definitely several ways to do it, and we could chat more about it.

kevinyang-cky: @charlie-becker, the modification is done, and please use Bridgescaler [v0.8.2](https://github.com/NCAR/bridgescaler/releases/tag/v0.8.2). Let me know if you have any questions. Thanks.

credit/preblock/concat.py
```diff
@@ -0,0 +1,56 @@
+import torch
+from credit.preblock.base import BasePreblock
+
+
+class ConcatenateToTensor(BasePreblock):
```
djgagne: Can you rename class to ConcatToTensor? Concatenate is a long word to write out repeatedly, and sometimes I mistype it.

credit/preblock/scaler.py
```diff
@@ -1,19 +1,35 @@
-import torch.nn as nn
-from bridgescaler import load_scaler
+from bridgescaler import load_scaler_dict, scale_var_dict
+from credit.preblock.base import BasePreblock
 
 
-class Scaler(nn.Module):
-    """
-    Scaling layer using a bridgescaler object. Supports transform and its inverse.
+class BridgeScaleTransformer(BasePreblock):
```
djgagne: Can you rename to `BridgeScalerTransformer` rather than `BridgeScaleTransformer` to be consistent with the package name?

credit/preblock/sqrt.py
djgagne: Has this metadata modification idea been implemented elsewhere in the code or still to be done?

djgagne (changes requested):
Some additional small changes.

charlie-becker:
> Has this metadata modification idea been implemented elsewhere in the code or still to be done?

Still to be done. Hoping to include that in a separate PR, but can address here if preferred.

djgagne (approved):
All changes look good to me. I approve.

PR #300, opened 2026-03-26
Updated example config and docs

djgagne:
I updated the example config to address some issues that Stas found when trying to train and test a model. The config will now train and validate a small model one 1 degree ERA5 data and then `credit_rollout_to_netcdf` can produce some small forecasts and perform pressure interpolation. I tested both training and rollout and got them to work. I do need to add rollout_realtime support to this example config but want to at least get this in first.


kafitzgerald:
I'll take a look at this in a bit.

It's odd that the check on CIRRUS isn't running here though.  @sgpearse, do you know what might be going on?

config/example-v2026.1.0.yml
```diff
@@ -26,7 +26,7 @@ data:
     # upper-air variables, must be YEARLY zarr or nc with (time, level, latitude, longitude) dims
     # files must have the listed variable names 
     # upper-air variables will be normalized by the dataloader, users do not need to normalize them
-    variables: ["u_component_of_wind", "v_component_of_wind", "temperature", "specific_total_water"]
+    variables: &prog_vars_3d ["u_component_of_wind", "v_component_of_wind", "temperature", "specific_total_water"]
```
kafitzgerald: I don't have a problem with this and it is more concise, but wonder if the anchor syntax used here might be less approachable than just repeating these (i.e. for folks not used to YAML).

docs/source/getting-started.md
```diff
@@ -8,23 +8,33 @@ the right versions of PyTorch and CUDA, for you. If you are running CREDIT on th
...
-conda create -n credit python=3.11
+conda create -n credit python=3.12
 conda activate credit
 ```
+
+:::{important}
+When installing PyTorch on Linux, the default option (v2.11) uses CUDA 13, which is currently not
+compatible with the CUDA driver on Casper. Add `--extra-index-url https://download.pytorch.org/whl/cu129` 
+to your pip install commands to install a PyTorch version compiled with CUDA 12.9. You can replace 
+`cu129` with `cu126` for CUDA 12.6 if necessary.
```
kafitzgerald: Do we have a feel for when it might be necessary to use 12.6?

kafitzgerald: Seems like I may have partially answered my question by running on a V100.

djgagne: I am modifying the instructions to point users to CUDA 12.6 so they can use all Casper GPUs. I doubt most people are going to notice execution speed differences between CUDA 12.6 and CUDA 12.9 although that might be useful to test at some point.

config/example-v2026.1.0.yml
djgagne: I removed the anchor tag for now.

djgagne:
@kafitzgerald I addressed every issue you raised except for the warnings. Those seem to be coming from CREDIT dependencies rather than CREDIT itself and will take a bit of effort to track down. The metadata config parser will now pull from the `credit.metadata` directory by default if no path is provided to the metadata file and will use era5.yaml if the argument is not provided at all. There's now a unit test file for output.py. 

I made the getting started instructions point to pytorch compiled with CUDA 12.6 since Casper GPU demand is high, and I want this to work out of the box for new users.

PR #305, opened 2026-03-27
[1/4] CREDIT v2: Trainer & Preblocks

### LINKED ISSUE #278: Preblock plans
Starting this thread to keep track of preblock implementation as part of the refactored datasets and preprocessing pipeline. The general idea with preblocks are to have PyTorch layers that can handle data transforms and scaling that can be combined together in one exportable workflow.

The original Dataset implementation in CREDIT folds loading from disk, combining different types of variables, and normalization into one class that has a confusing web of function calls to perform all of these tasks. See ERA5_Multistep_Batcher for an example. In the new data processing paradigm, we want to split the functionality into the following steps:

1. Source Datasets: each of these uses a PyTorch Dataset class to load data from a particular source (e.g., ERA5, GOES, MRMS, CAM, MOM6, etc) from disk and organize into a dictionary of tensors for each type of field (prognostic, diagnostic, dynamic forcing, static) and potentially for each variable.
2.  Transforms of individual variables: be able to do log transforms on fields like Q or surface pressure.
3. Calculate derived variables: e.g., calculating TOA radiation or solar zenith angle on the fly rather than loading from disk. Could be called from the dataset potentially. 
4. Regridding: if the input data are on a lat-lon grid and we want to convert to HEALPIX or regridding regional data to or from Lambert Conformal or Albers Equal Area. Also, if different data sources are on different grids and we want to regrid them to a common grid. The regridding weights would be calculated offline. 
5. Vertical interpolation: Aggregate 137 model levels to a smaller subset either through spline fitting or averaging.
6. Normalizing: Apply bridgescaler standard, minmax, or quantile scalers to model. Scaler values are calculated with bridgescaler offline.
7. Concatenating: Combine all the processed data into one datacube to go into the model.

@charlie-becker, @kevinyang-cky, and I had a meeting today to discuss implementation details for the preblocks. 

* Adding metadata to tensors: Since many preprocessing operations are conditional on what source and type of variable is being used, we need to include some metadata along with the tensors describing the source, variable, level, and potentially type of variable. We want to make the syntax as clear as possible but also avoid naming collisions across datasets. The metadata schema/syntax is going to be in flux while we figure out the best option here.
* When to concatenate into a combined tensor: Charlie was expressing preference for doing this early on and using metadata to distinguish parts of the tensor to grab for preblocks. One potential issue with this is that PyTorch autograd doesn't work with in-place operations on tensors and can throw errors or just be incorrect if you try to backprop through one of these operations. I think it would be easier to concatenate near the end rather than at the beginning, but we will need to test this in more detail.
* One bridgescaler scaler or support multiple ones?

jsschreck:
## Summary

First of four sub-PRs splitting up `trainer_refactor_v2` (#298).
**Merge order: [1/4] → [2/4] → [3/4] → [4/4].**

This PR contains the foundational trainer and data-pipeline changes. Please review carefully — [2/4] and [4/4] depend on it.

### What changed

**New gen2 trainer**
- `trainerERA5gen2.py` — new gen2 trainer so we could keep the v1 `trainerERA5` untouched rather than bolt gen2 behavior onto a shared class. Nested data schema, `apply_preblocks`-assembled batches, `forecast_len` 1-indexed, configurable `backprop_on_timestep`. Multi-step autoregressive loop: at `t > 1` ERA5Dataset returns only dynfrc channels, so the loop clones the previous `x`, splices in new forcing, and updates prognostic channels in-place.

**Preblock system**
- `credit/preblock/__init__.py` — `build_preblocks(conf)` and `apply_preblocks(preblocks, batch)` are the new public API. `ConcatPreblock` is gone; `ConcatToTensor` (in `concat.py`) assembles the tensor in ERA5Dataset insertion order: `[dynfrc | static | prog]`.
- `norm.py` — `ERA5Normalizer` z-score preblock.
- `log.py`, `sqrt.py`, `scaler.py`, `regrid.py`, `base.py` — supporting preblocks.

**All existing trainers updated (`self.conf` refactor) — the bulk of the file count**
`trainerERA5`, `trainerERA5_ensemble`, `trainerERA5_Diffusion`, `trainerWRF`, `trainerWRF_multi`, `trainerLES`, `trainer_downscaling`, `trainer_om4_samudra` — all updated to the new `BaseTrainer` API (`self.conf` keyword argument, updated scheduler/EMA hooks). This is why the PR touches so many files. **V1 trainer behavior is unchanged;** every change is structural (constructor signature), no logic was altered.

**Base trainer**
- `base_trainer.py` — `EMATracker`, `LinearWarmupCosineScheduler`, unified `BaseTrainer` init supporting DDP/FSDP/none modes.

**Preflight**
- `preflight.py` — dataloader startup check run before PBS job submission.

**Note on trainer registry string**
The class is now `TrainerERA5Gen1` but the registry key in `credit/trainers/__init__.py` is still `"era5"` (configs use `trainer.type: era5`). Updating that string to `"era5-gen1"` requires touching all v1 configs — deferred to [3/4] which already sweeps all config files.

**Application renames**
- `train.py` → `train_gen1.py` (v1 entry point kept as `credit_train` alias in `pyproject.toml`)
- New `train_gen2.py` — gen2 CLI entry point (`credit_train_gen2`)
- `credit/applications/` symlinks updated to match.

**Tests — backfilling missing coverage**
`tests/test_preblock_preflight_scheduler.py` and `tests/test_trainer_components.py` are new — there were no tests for these components before. This also inflates the file count. `test_trainer_components.py` technically lives in [4/4] but is runnable on this branch now.

**Smoke configs**
- `config/smoke/smoke_gen2_casper.yml` — single-step smoke test
- `config/smoke/smoke_gen2_multistep_casper.yml` — `forecast_len: 2` multi-step smoke test

---

## Test plan

- [x] `python -m pytest tests/test_trainer_components.py` (lives in [4/4] but runnable on this branch now — covers gen2 trainer init, multi-step rollout channel correctness, scheduler, EMA)
- [x] `python -m pytest tests/test_preblock.py tests/test_preblock_preflight_scheduler.py` — 50 passed, 3 skipped (bridgescaler/numba skipped on NumPy 2.4)
- [x] Spot-check `trainerERA5gen2.py` multi-step autoregressive loop — `static_dim_size = n_dynfrc + n_static` (line 91); at t>1: `x_new[:, :n_dynfrc] = x_dynfrc`, `x_new[:, static_dim_size:] = y_pred_prog`, static unchanged. Correct.
- [x] Confirm `apply_preblocks` channel order — ERA5Dataset inserts dynfrc first (always), then static+prog only at i==0. `ConcatToTensor` concatenates in insertion order → `[dynfrc | static | prog]`. Matches `static_dim_size` slice math. Correct.
- [x] Confirm existing v1 trainer behavior unchanged — diff shows only: (1) class renamed `Trainer`→`TrainerERA5` (alias restored), (2) `conf` added to `__init__` signature, (3) cosmetic import reorder. No train/validation loop logic changed. `Trainer = TrainerERA5` alias added for backward compat.
- [x] Smoke test on Casper: `config/smoke/smoke_gen2_casper.yml` (single-step) — passed
- [x] Smoke test on Casper: `config/smoke/smoke_gen2_multistep_casper.yml` (`forecast_len: 2`) — passed (2 epochs, validation loss converging, checkpoint saved)
- [x] `pip install .` builds cleanly (symlink fix pushed; was failing CI due to broken `train_gen1.py` symlink)


jsschreck:
@WillyChap feel free to add any new pre-blocks stuff you might have written for CAMulator here if its not already up and posted. I know you have post-blocks stuff. That refactor is coming when @charlie-becker is done with pre-blocks.

credit/preblock/concat.py
```diff
@@ -0,0 +1,93 @@
...
+    for key, tensor in tensor_dict.items():
+        sk = _sort_key(key, field_order)
+        groups.setdefault(sk, []).append(tensor)
+    parts = []
+    for sk in sorted(groups.keys()):
+        parts.extend(groups[sk])
+    return torch.cat(parts, dim=1)
+
+
+class ConcatPreblock(nn.Module):
```
djgagne: @charlie-becker also has a concat preblock. Is this meant to only replicate the Gen 1 dataset concat process? If so we might want to name it ConcatGen1 or something like that.

credit/trainers/preflight.py
```diff
@@ -0,0 +1,212 @@
...
+
+        bytes_per_sample = H * W * total_ch * 4  # float32
+        bytes_per_sample *= 2  # input + target
+
+        workers = trainer_conf.get("thread_workers", 4)
+        prefetch = trainer_conf.get("prefetch_factor", 4)
+        batch_size = trainer_conf.get("train_batch_size", 1)
+
+        total_bytes = workers * prefetch * batch_size * bytes_per_sample
+        return total_bytes / 1e9
```
djgagne: Should divide by 2e30 rather than 1e9 since RAM uses the base-2 definition of the gigabyte (GiB) rather than the base-10 definition.

djgagne: Might also want to replace uses of GB with GiB to reduce ambiguity about what definition is being used.

credit/trainers/preflight.py
```diff
@@ -0,0 +1,212 @@
...
+
+        return psutil.virtual_memory().available / 1e9
+    except ImportError:
+        return 0.0
+
+
+# ---------------------------------------------------------------------------
+# First-batch timeout check
+# ---------------------------------------------------------------------------
+
```
djgagne, line 101: Can you also add a check that creates a synthetic batch of data and sends it on a forward and backward pass through the model on GPU and then does an optimizer step? That way we can get a max vram usage for a given model configuration.

credit/trainers/base_trainer.py
```diff
@@ -2,53 +2,122 @@
...
 
+try:
+    from torch.utils.tensorboard import SummaryWriter as _SummaryWriter
+except ImportError:
+    _SummaryWriter = None
+
 logger = logging.getLogger(__name__)
 
 
+class EMATracker:
```
djgagne, line 42: To me, EMATracker would make more sense in `scheduler.py` rather than in `base_trainer`.

credit/trainers/__init__.py
```diff
@@ -3,10 +3,18 @@
 
 # Import trainer classes
```
djgagne, line 4: Since we're doing a big rewrite, can we rename all the Trainer classes within their individual modules rather than renaming in `__init__.py`?

credit/trainers/base_trainer.py
```diff
@@ -57,94 +126,180 @@ def __init__(self, model: torch.nn.Module, rank: int):
...
+        # Store full conf for subclass method bodies that still need it
+        self.conf = conf
+
+        # ---- Extract all trainer settings ----
+        trainer_conf = conf["trainer"]
+        self.save_loc = os.path.expandvars(conf["save_loc"])
+        self.mode = trainer_conf.get("mode", "none")
+        self.distributed = self.mode in ("fsdp", "ddp")
+        self.start_epoch = trainer_conf.get("start_epoch", 0)
+        self.epochs = trainer_conf.get("epochs", 70)
```
djgagne: 70 is a ton of default epochs. Maybe default to 1 or 10?

credit/trainers/base_trainer.py
```diff
@@ -57,94 +126,180 @@ def __init__(self, model: torch.nn.Module, rank: int):
...
+        self.grad_max_norm = trainer_conf.get("grad_max_norm", 0.0)
+        self.batches_per_epoch = trainer_conf.get("batches_per_epoch", 0)
+        self.valid_batches_per_epoch = trainer_conf.get("valid_batches_per_epoch", 0)
+        self.ensemble_size = trainer_conf.get("ensemble_size", 1)
+        self.save_best_weights = trainer_conf.get("save_best_weights", True)
+        self.save_backup_weights = trainer_conf.get("save_backup_weights", True)
+        self.stopping_patience = trainer_conf.get("stopping_patience", 100)
+        self.save_every_epoch = trainer_conf.get("save_every_epoch", False)
+        self.stop_after_epoch = trainer_conf.get("stop_after_epoch", False)
+        self.num_epoch = trainer_conf.get("num_epoch", int(1e8))
```
djgagne, line 156: Why is the default here 1e8?

credit/trainers/base_trainer.py
djgagne: That said, if EMATracker is only used in BaseTrainer, then it would be fine to leave here.

credit/trainers/trainerERA5v2.py
```diff
@@ -0,0 +1,421 @@
...
+            if self.distributed:
+                dist.all_reduce(batch_loss, dist.ReduceOp.AVG, async_op=False)
+            results_dict["train_loss"].append(batch_loss[0].item())
+            results_dict["train_forecast_len"].append(self.forecast_len)
+
+            if not np.isfinite(np.mean(results_dict["train_loss"])):
+                print(results_dict["train_loss"])
+                raise optuna.TrialPruned()
+
+            to_print = "Epoch: {} train_loss: {:.6f} train_acc: {:.6f} train_mae: {:.6f} forecast_len: {:.0f}".format(
```
djgagne: It would be good to break the logging bit out into a separate method in base_trainer that is not as hard coded to certain metrics.

djgagne (changes requested):
Overall the PR looks really good. I have specific questions about some of the different parts, including a few more modifications. As a matter of naming convention, let's also try to convert v2 mentions to Gen 2 so that we separate the API changes from CREDIT package versioning. The biggest change here would be renaming `trainerERA5v2` to `trainerERA5gen2`.

credit/preblock/concat.py
jsschreck: ConcatToTensor is the Gen 2 schema concat (Charlie's implementation from PR #297). The old ConcatPreblock it replaced was the Gen1 placeholder. Name is appropriate as-is.

credit/trainers/preflight.py
jsschreck: Fixed. Changed `/ 1e9` → `/ 2**30` throughout and renamed to `estimate_dataloader_memory_gib` / `_available_ram_gib`.

jsschreck: Updated all log messages and error strings to use GiB as well.

credit/trainers/preflight.py
jsschreck: Added `check_model_gpu_memory(conf, model, optimizer, rank)` to `preflight.py`. Creates a zero-filled batch of the inferred input shape, runs a full forward/backward/optimizer step, and logs peak VRAM (MiB and % of total GPU memory). Called from `base_trainer.fit()` before the epoch loop. Input channel count is inferred from `model.frames × (channels × levels + surface_channels + input_only_channels)`; skips silently if that product is zero.

credit/trainers/base_trainer.py
jsschreck: EMATracker is only used in BaseTrainer, so per your followup comment we left it here.

jsschreck: Confirmed — left in base_trainer since it's only used there.

credit/trainers/__init__.py
jsschreck: Done for all trainers: `TrainerERA5`, `TrainerERA5Gen2`, `TrainerERA5Diffusion`, `TrainerERA5Ensemble`, `TrainerDownscaling`, `TrainerIC`, `TrainerSamudra`, `TrainerLES`, `TrainerWRF`, `TrainerWRFMulti`. The `as` aliasing in `__init__.py` is removed. Also renamed `trainerERA5v2.py` → `trainerERA5gen2.py`, class to `TrainerERA5Gen2`, and config key `era5-v2` → `era5-gen2`.

credit/trainers/base_trainer.py
jsschreck: Changed to 1.

credit/trainers/base_trainer.py
jsschreck: `num_epoch` caps how many additional epochs run from `start_epoch`. The `1e8` sentinel means 'don't limit' — `epochs` is the effective cap when `num_epoch` is not set in config. Added a comment to that effect.

credit/trainers/trainerERA5v2.py
jsschreck: Added `_log_batch_progress(epoch, results_dict, optimizer, pbar, phase)` to `BaseTrainer`. Iterates over standard metric keys (`{phase}_loss`, `{phase}_acc`, `{phase}_mae`) and appends LR for the train phase. `TrainerERA5Gen2` now calls it for both the train and validate loops.

jsschreck:
@djgagne @charlie-becker — quick design question before [2/4] moves forward.

Right now [2/4] bundles three things:
1. The `credit migrate` / `credit ask` / `credit plot` CLI commands
2. The gen2 rollout scripts (`rollout_realtime_gen2.py`, `rollout_to_netcdf_gen2.py`)
3. The `api.py` FastAPI server

The rollout scripts are fairly standalone and touch the same channel-ordering logic we fixed in [1/4]. Would it be cleaner to pull them out into their own PR that sits between [1/4] and the CLI work? The CLI commands themselves (`credit migrate`, `credit ask`) do not depend on the rollout scripts at all.

Arguments for splitting:
- Rollout scripts are the most review-sensitive piece (channel math, denorm, multi-step state update)
- Reviewers could sign off on inference correctness independently of the CLI ergonomics discussion
- Smaller surface area per PR

Arguments against:
- One more PR to manage in the chain
- The rollout scripts are already written and tested; splitting is mostly overhead

Happy either way — wanted your read before we add more commits on top.

applications/train_gen2.py
```diff
@@ -0,0 +1,342 @@
...
+    """Build a MultiSourceDataset for train or validation."""
+    if is_train:
+        data_conf = conf["data"]
+    else:
+        # Merge data_valid overrides on top of data
+        data_conf = {**conf["data"], **conf.get("data_valid", {})}
+        # source config always comes from conf["data"]
+        data_conf["source"] = conf["data"]["source"]
+
+    return MultiSourceDataset(data_conf, return_target=True)
```
charlie-becker: Should only `return_target=True` if `is_train=True`, otherwise `False`

applications/train_gen2.py
```diff
@@ -0,0 +1,342 @@
...
+    if conf["trainer"].get("update_learning_rate", False):
+        for param_group in optimizer.param_groups:
+            param_group["lr"] = learning_rate
+
+    return conf, model, optimizer, scheduler, scaler
+
+
+def main(rank, world_size, conf, backend=None, trial=False):
+    """Set up and run training."""
+    conf["save_loc"] = os.path.expandvars(conf["save_loc"])
```
charlie-becker, line 105: Can we potentially move most of the other helper functions in this scrip to `credit/trainers/utils.py` (or elsewhere) to simplify this script for users? Ideally, only leaving `main()` and `main_cli()`

charlie-becker:
> @djgagne @charlie-becker — quick design question before [2/4] moves forward.
> 
> Right now [2/4] bundles three things:
> 
> 1. The `credit migrate` / `credit ask` / `credit plot` CLI commands
> 2. The gen2 rollout scripts (`rollout_realtime_gen2.py`, `rollout_to_netcdf_gen2.py`)
> 3. The `api.py` FastAPI server

I believe the rollout scripts will need quite a bit more clean-up for Gen2 and V2. My preference would be to leave it out of this PR (1/4) which focuses mostly on the trainer and preblocks, which is substantial in itself.

applications/train_gen2.py
```diff
@@ -0,0 +1,342 @@
...
+        seed = conf.get("seed", 42) + rank
+    else:
+        batch_size = conf["trainer"]["valid_batch_size"]
+        shuffle = False
+        seed = conf.get("seed", 42)
+
+    forecast_len = conf["data"]["forecast_len"]
+    num_workers = conf["trainer"].get("thread_workers" if is_train else "valid_thread_workers", 4)
+    prefetch_factor = conf["trainer"].get("prefetch_factor", 2) if num_workers > 0 else None
+
```
charlie-becker: I'm not sure this is needed. I think you can still benefit from prefetch with a serial / single worker process (although with limited benefit).

config/smoke/smoke_gen2_casper.yml
```diff
@@ -0,0 +1,151 @@
...
+    valid_thread_workers: 2
+    prefetch_factor: 2
+
+preblocks:
+  norm:
+    type: era5_normalizer
+    args:
+      mean_path: "/glade/campaign/cisl/aiml/ksha/CREDIT_data/ERA5_mlevel_1deg/mean_std/mean_6h_1979_2019_conserve_1deg.nc"
+      std_path: "/glade/campaign/cisl/aiml/ksha/CREDIT_data/ERA5_mlevel_1deg/mean_std/std_6h_1979_2019_conserve_1deg.nc"
+
```
charlie-becker: Since most of pre-block stuff is data relevant can you move it up directly below the data section?

config/smoke/smoke_gen2_casper.yml
```diff
@@ -0,0 +1,151 @@
...
+          path: "/glade/campaign/cisl/aiml/ksha/CREDIT_data/ERA5_mlevel_1deg/static/ERA5_mlevel_1deg_static_subset.zarr"
+        diagnostic: null
+
+  timestep: "6h"
+  forecast_len: 1
+  start_datetime: "2000-01-01"
+  end_datetime: "2003-12-31"
+
+data_valid:
+  start_datetime: "2014-01-01"
```
charlie-becker: can you change `data_valid` to `validation_data` or something slightly more descriptive? Although obvious to us, I think `data_valid` isn't super clear.

credit/trainers/preflight.py
```diff
@@ -0,0 +1,296 @@
...
+        rank:      Global rank. Checks only run on rank 0.
+        timeout_s: Seconds to wait for the first batch before failing.
+    """
+    if rank != 0:
+        return
+
+    trainer_conf = conf.get("trainer", {})
+    workers = trainer_conf.get("thread_workers", 4)
+    prefetch = trainer_conf.get("prefetch_factor", 4)
+    batch = trainer_conf.get("train_batch_size", 1)
```
charlie-becker, line 133: Defualt of `prefetch_factor=4` seems high. Maybe we should change to 2?

applications/train_gen2.py
```diff
@@ -0,0 +1,342 @@
...
+    """Inject Gen1-compatible variable keys into conf["data"] for metrics/loss.
+
+    ``LatWeightedMetrics`` and ``VariableTotalLoss2D`` expect flat lists at:
+    - ``conf["data"]["variables"]``       (upper-air / 3D vars)
+    - ``conf["data"]["surface_variables"]`` (2D prognostic vars)
+    - ``conf["data"]["diagnostic_variables"]``
+
+    These are derived from the nested Gen2 source config so that metrics/loss
+    classes can be used without modification.
+    """
```
charlie-becker: Hopefully, we can eventually get rid of this all together with V2 and use the metadata from the preblocks / datasets directly. But should be fine until then.

charlie-becker (changes requested):
Looking great. I focused the review on `train_gen2.py`. 

I have successfully tested `train_gen2.py` with the smoke configs (and other datasets) on both Casper and Derecho with no changes. Nice work!

I left a few general comments regarding clarity, default settings and thoughts for downstream changes.

jsschreck:
> I believe the rollout scripts will need quite a bit more clean-up for Gen2 and V2. My preference would be to leave it out of this PR (1/4) which focuses mostly on the trainer and preblocks, which is substantial in itself.

Do you want me to make it a stand-alone PR? (like maybe 3/5?, and update the others to */5?). Its no problem, you are right we are going to need to make more changes (but at the same time its the same script so a lot of changes but just a handful of scripts).

applications/train_gen2.py
```diff
@@ -0,0 +1,342 @@
...
+        if conf["pbs"]["queue"] == "casper":
+            logging.info("Launching to PBS on Casper")
+            launch_script(config, script_path)
+        else:
+            logging.info("Launching to PBS on Derecho")
+            launch_script_mpi(config, script_path)
+        sys.exit()
+
+    local_rank, world_rank, world_size = get_rank_info(conf["trainer"]["mode"])
+    main(world_rank, world_size, conf, backend)
```
djgagne: We can combine main_cli and main into one function. I didn't do this at the time due to just trying to get stuff to work, but we could refactor now without any issues.

charlie-becker:
> > I believe the rollout scripts will need quite a bit more clean-up for Gen2 and V2. My preference would be to leave it out of this PR (1/4) which focuses mostly on the trainer and preblocks, which is substantial in itself.
> 
> Do you want me to make it a stand-alone PR? (like maybe 3/5?, and update the others to */5?). Its no problem, you are right we are going to need to make more changes (but at the same time its the same script so a lot of changes but just a handful of scripts).

I think a separate PR for rollout makes sense and should be small enough for efficient review, but others can chime in if they other prefernces.

djgagne:
> > > I believe the rollout scripts will need quite a bit more clean-up for Gen2 and V2. My preference would be to leave it out of this PR (1/4) which focuses mostly on the trainer and preblocks, which is substantial in itself.
> > 
> > 
> > Do you want me to make it a stand-alone PR? (like maybe 3/5?, and update the others to */5?). Its no problem, you are right we are going to need to make more changes (but at the same time its the same script so a lot of changes but just a handful of scripts).
> 
> I think a separate PR for rollout makes sense and should be small enough for efficient review, but others can chime in if they other prefernces.

Yeah, rollout should definitely be a separate PR. It doesn't have to be part of this series, can stand on its own.

credit/preblock/norm.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # mean/std may be scalar (2D var) or 1-D levels vector.
+        # tensor shape: (B, levels_or_1, T, H, W)
+        if mean.dim() == 1 and mean.shape[0] > 1:
+            # Per-level stats: reshape to (1, levels, 1, 1, 1)
+            mean = mean.view(1, -1, 1, 1, 1)
+            std = std.view(1, -1, 1, 1, 1)
+        # else: scalar — broadcasts naturally
+
+        return (tensor - mean) / (std + 1e-7)
```
djgagne: Adding an fixed epsilon value to the standard deviation is problematic for normalizing really small numbers  or ones with large dynamic ranges like microphysical quantities.

credit/preblock/__init__.py
```diff
@@ -1,18 +1,34 @@
...
 from credit.preblock.log import LogTransform
 from credit.preblock.sqrt import SqrtTransform
-from credit.preblock.scaler import BridgeScalerTransformer
 from credit.preblock.regrid import Regridder
 from credit.preblock.concat import ConcatToTensor
+from credit.preblock.norm import ERA5Normalizer
+
+# bridgescaler depends on numba which requires NumPy ≤ 2.2
+try:
+    from credit.preblock.scaler import BridgeScalerTransformer
```
djgagne, line 13: This protective block should no longer be necessary with the updates to the environment build. Need to test.

credit/transforms/deprecated/_transforms.py
```diff
@@ -13,7 +13,11 @@
 import torch
 
 from credit.data import Sample, device_compatible_to
-from bridgescaler import read_scaler
+
+try:
+    from bridgescaler import read_scaler
```
djgagne: Shouldn't be necessary.

credit/transforms/transforms_quantile.py
```diff
@@ -15,7 +15,12 @@
 import xarray as xr
 
 import torch
-from bridgescaler import read_scaler
+
+try:
+    from bridgescaler import read_scaler
```
djgagne: Shouldn't be necessary with new environment.

tests/test_preblock.py
```diff
@@ -141,6 +148,12 @@ def test_regrid_flip_axis(weight_file):
     )
 
 
+_skip_bridgescaler = pytest.mark.skipif(
```
djgagne, line 151: Again should hopefully be removable with new environment.

djgagne (changes requested):
Overall most of the changes look good and put us on the path to the Gen 2 API. I just merged with main and only had to resolve a couple of small conflicts. I think the new environment from v2026.1.1 should fix the numba issues currently being encountered, especially getting rid of miles-guess as a required dependency.

applications/train_gen2.py
jsschreck: We found validation needs targets to compute the loss in the validate loop (`y_pred` vs `y`). Changed to always `return_target=True`. If there's a use case for returning no target during validation, we can add a `skip_validation` shortcut path, but the standard path requires it.

applications/train_gen2.py
jsschreck: Done — `inject_flat_var_keys`, `load_dataset`, `load_dataloader`, and `load_model_states_and_optimizer` all moved to `credit/trainers/utils.py`. `train_gen2.py` now only contains the `main_cli` entry point.

applications/train_gen2.py
jsschreck: Agreed. Added a guard: `prefetch_factor = conf["trainer"].get("prefetch_factor", 2) if num_workers > 0 else None` so it's only passed when `num_workers > 0`.

config/smoke/smoke_gen2_casper.yml
jsschreck: Done in all smoke configs — `preblocks:` now sits directly below `validation_data:`.

config/smoke/smoke_gen2_casper.yml
jsschreck: Done throughout — config key, trainer code, all smoke configs, and docstrings updated to `validation_data`.

credit/trainers/preflight.py
jsschreck: Changed default to 2 in `load_dataloader`.

applications/train_gen2.py
jsschreck: Agreed — added a TODO comment in the function noting this should be removed once metrics/loss are updated to read directly from the Gen2 source config.

applications/train_gen2.py
jsschreck: Done — merged into a single `main_cli()` that handles argument parsing and the full training loop directly.

credit/preblock/norm.py
jsschreck: Fixed — changed `(std + 1e-7)` to `std.clamp(min=1e-12)`, which avoids inflating near-zero stds without adding a fixed offset.

credit/preblock/__init__.py
jsschreck: Kept a minimal guard for now since not all downstream environments are updated yet. Will remove in the follow-up env standardization PR.

credit/transforms/deprecated/_transforms.py
jsschreck: Kept for now pending env standardization. Will remove in cleanup PR once all environments are confirmed updated.

credit/transforms/transforms_quantile.py
jsschreck: Same as above.

tests/test_preblock.py
jsschreck: Agreed — skip guard kept for now pending env standardization. Will remove once all CI environments have bridgescaler confirmed.

applications/train_gen2.py
charlie-becker: I already see a `skip_validation` argument in the config...

I would prefer this to be an option, as it requires twice the memory and I/O when a target is loaded.

jsschreck: The validation loop uses the target to compute loss and metrics — omitting it causes the `(input, target, metadata)` unpack to fail with a 2-value tuple. So `return_target=True` is required whenever validation actually runs.

That said, your point is valid: if `skip_validation=True`, we shouldn't even build the validation dataset. We can gate the validation dataset construction on `not conf["trainer"].get("skip_validation", False)`, which avoids the memory/IO cost entirely in that case. Would that work for you?

charlie-becker: Yep, that should work. Thanks.

jsschreck: Done — validation dataset and loader are now only built when `skip_validation=False`. If `skip_validation=True`, `valid_loader=None` is passed straight through to `trainer.fit()`, which already guards all validation paths behind `self.skip_validation`.

djgagne (approved):
I am happy with the changes. Approved.

charlie-becker (approved):
Thanks for addressing my concerns. LGTM

PR #320, opened 2026-04-01
Domain parallelism for V2: spatial sharding across GPUs for high-resolution training

negin513:
# Short Summary

Adds a `credit/domain_parallel` package that enables training on  high-resolution spatial inputs that exceed single-GPU memory by sharding the spatial (latitude) dimension across multiple GPUs.                                                          
                                                                                                                  
**Inspired by [PhysicsNeMo's `ShardTensor`](https://docs.nvidia.com/physicsnemo/latest/user-guide/domain_parallelism/fsdp_and_shard_tensor.html).**

Here is what is included in the new module: `credit/domain_parallel/`

- **`manager.py`** — `DomainParallelManager`: builds a 2D logical process mesh of `(data_parallel × domain_parallel)` groups from a flat world of GPUs. Domain-parallel ranks share the same data sample but hold different spatial shards; data-parallel ranks hold the same shard but process different samples.

- **`sharding.py`** — `shard_tensor`, `gather_tensor`, `shard_batch`: split tensors along the H dimension for input distribution and all-gather them back for metrics/output.

- **`halo_exchange.py`** — `HaloExchange`: a differentiable boundary communication primitive that exchanges halo rows between neighboring ranks before each convolution so boundary pixels see correct neighbors.

- **`layers.py`** — drop-in domain-parallel replacements for `nn.Conv2d`, `nn.Conv3d`, `nn.ConvTranspose2d`, `nn.GroupNorm`, and `F.interpolate`; each wrapper handles halo exchange and/or distributed reductions automatically.

- **`convert.py`** — `convert_to_domain_parallel`: walks an existing model and replaces supported layers in-place, plus `validate_sharding_constraints` to catch config errors early.  

## Tests                                                                                                          
  - `tests/test_domain_parallel.py`: unit tests for sharding math, halo width  calculations, and layer forward/backward correctness (single-process, CPU).
  - `tests/test_domain_parallel_multigpu.py`: multi-GPU integration tests using  `torchrun`; validates end-to-end forward + backward with halo exchange across 2 GPUs.   
  
  no `DTensor` until FSDP is updated to FSDP2....


jsschreck (commented):
.

jsschreck:
Hey Negin, a few things you might consider before or after merge — happy to take these on myself once your initial code is in if it's easier:

- Add a GPU count check to skip the multi-GPU tests when fewer than 2 GPUs are available (addresses DJ's CI issue)
- Add explicit error when `mode: domain_parallel` is set but `domain_parallel_size` is missing/1 in config — currently silently runs single-GPU
- **Gradient sync after `loss.backward()` in the trainer is still missing** — without an `all_reduce` across domain ranks, boundary gradients will silently diverge and parameters will drift out of sync during training. In `trainerERA5.py`, after the backward step and before `optimizer.step()`:
  ```python
  if use_domain_parallel:
      for p in model.parameters():
          if p.grad is not None:
              dist.all_reduce(p.grad, op=dist.ReduceOp.AVG,
                              group=domain_manager.domain_group)
  ```
- Clarify behavior of pure `domain_parallel` mode when `world_size > domain_parallel_size` (no DDP/FSDP wrapping means data-parallel replicas won't sync)
- Consider passing the domain manager through conf or the Trainer rather than storing it as `model._domain_parallel_manager`

Want me to just handle these as a follow-up PR after this merges?

jsschreck (approved):
Thank you so much! Huge improvement :D :D

Issue #185, opened 2025-04-09
Thoughts on config file readability


charlie-becker:
There are multiple points in the configuration files that I found unclear as I navigated the code base for the first time. Here's a list of some pain points and suggestions for clarity:

1. `save_loc` is the location of the model that you want to load for use. This is not intuitive to me -- I feel like `save_loc` is where the output of my run should be saved. I would recommend changing this to `model_location` or something similar. I also feel that the "location where we do want to save models / data / etc" should be listed up top (here). 
2. There are a number of perhaps outdated config parameters that bloat the config. Examples: 
`  type: "custom"       # keep it as "custom"`, 
` skip_periods: null  # do not use skip_period`
3. Inconsistent use of in-line comments in the configuration file.  It would be good if in-line comments were present for all parameters (or None) which can be a "quick-guide" and then can go to full config readme for detailed clarification. Especially for non-intuitive values like: `history_len: 1`, `validation_history_len: 1` (at least provide units)
4. The `predict['forecasts']:` could be much more intuitive:  
    
```
type: "custom"       # keep it as "custom"
start_year: 2020     # year of the first initialization (where rollout will start)
start_month: 1       # month of the first initialization
start_day: 1         # day of the first initialization
start_hours: [0, 12] # hour-of-day for each initialization, 0 for 00Z, 12 for 12Z
duration: 30         # number of days to initialize, starting from the (year, mon, day) above
                             # duration should be divisible by the number of GPUs 
                             # (e.g., duration: 384 for 365-day rollout using 32 GPUs)
days: 2              # forecast lead time as days (1 means 24-hour forecast)

```
I think date strings parsed with pandas is much more readable. The same functionality of above could be re-written as:
 
```
start_date: "2020-01-01" 
end_date: "2020-01-30"
init_frequency: "12h"
forecast_length: "48h"
```
5.  There are too many configs, and test configs (unless meant to used for a simple documented test case for users) should be deleted. There is room for improvement on naming convention / consistency. Is a separate config needed for all model combinations if only a couple of config parameters are changed (1h vs 6h, for example)? The more configs there are with different data paths, for example, the more chances for downstream error. I think it might make sense to have sub-directories: 

```miles-credit/config/
--> wxformer/
     --> train_1h.yml
     --> train_6h.yml
--> fuxi/
     ...config
--> inference/
 ```
6. I believe the training data is still scattered across the filesystem. The data should all be moved to a centralized location and the configs updated and verifies to reflect.

PR #327, opened 2026-04-08
Configuration File Docs Text Update

danineamati:
### Description

There were a few places in the "What's in the Configuration File?" docs that seemed to include outdated text. Correspondingly, the following have been changed:
1. Outdated conversational text after variable types (i.e., Periodic & Static Forcing section)
2. Many tables were listed as "Summary of Key Recommendations" and are now more direct (e.g., "Summary of Key Training Strategy Recommendations")
3. There were two "Model Configuration" sections, with one seemingly having been a placeholder.

Note: It may be worth considering the Markdown header number. Right now, we have `# What's in the Configuration FIle` and `# CREDIT Configuration Guide`. I wonder if we need the latter at all. We can also consider multiple top level `#` to better split the sections in the table of contents.

### Related Issues and PRs

This does not close any current issues. Larger structural changes are still retained in issue #185.

### Checklist
- [X] The documentation is up to date.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -171,3 +171,6 @@ climate/Look_at_Tens.ipynb
 
 # local dev / HPC-specific test scripts
 test_energy_fixer_updown.py
+
+# MAC DS_Store files
+.DS_Store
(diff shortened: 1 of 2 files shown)


djgagne (approved):
Thank you for catching these, especially the LLM text. Looks good to me.

PR #328, opened 2026-04-08
models: guard legacy imports against NumPy >=2.3 / numba conflict

jsschreck:
## Summary

`bridgescaler` pulls in `numba`, which breaks on NumPy >=2.3 in CI — `import credit.models` hard-fails and no tests run.

This wraps the legacy model imports in a try/except and exposes `_LEGACY_MODELS_AVAILABLE` so test suites can skip gracefully rather than error out.

This is a CI compatibility fix, not a gen2 architectural change. Splitting it out of the [1/4] trainer PR so it can merge independently.

**@djgagne** — `bridgescaler` is yours. Ideally this guard lives upstream (bridgescaler pins or drops the numba dependency for NumPy >=2.3) so CREDIT does not need to own it. Flagging here so it is on your radar.

**@charlie-becker** — tagging you as well since this touches model loading.

## Files
- `credit/models/__init__.py` — try/except around legacy imports, `_LEGACY_MODELS_AVAILABLE` flag
- `tests/test_models.py` — skip guards using the flag
- `tests/test_model_loading.py` — new tests for `load_model` / `load_fsdp_or_checkpoint_policy` error paths

## Test plan
- [ ] CI passes on Python 3.11/3.12/3.13
- [ ] `python -m pytest tests/test_models.py tests/test_model_loading.py`

### DIFF
--- a/credit/models/__init__.py
+++ b/credit/models/__init__.py
@@ -3,67 +3,66 @@ import sys
 import copy
 import logging
 
-# Import model classes
-from credit.models.crossformer import CrossFormer
-from credit.models.camulator import Camulator
-from credit.models.unet import SegmentationModel
-from credit.models.fuxi import Fuxi
-from credit.models.swin import SwinTransformerV2Cr
-from credit.models.graph import GraphResTransfGRU
-from credit.models.debugger_model import DebuggerModel
-from credit.models.wxformer.crossformer import CrossFormer as WXFormer
-from credit.models.wxformer.crossformer_ensemble import CrossFormerWithNoise
-from credit.models.wxformer.crossformer_downscaling import DownscalingCrossFormer
-from credit.models.unet_downscaling import DownscalingSegmentationModel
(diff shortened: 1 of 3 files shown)


jsschreck:
@djgagne is this PR now redundant?

djgagne:
@jsschreck I think so. I do need to revisit bridgescaler to remove its numpy < 2.4 and pandas < 3 now that bug fix releases for pandas 3.0 have come out and numba supports numpy 2.4. Removing miles-guess from the main dependency tree also fixed a bunch of issues. I would like to propose as a separate PR redoing the credit models registry to use lazy imports rather than importing every single model module every time CREDIT is loaded. Doing so should significantly speed up the initialization of credit train and credit rollout and allow us to include custom modules in the registry not just those in base credit.

jsschreck:
> I would like to propose as a separate PR redoing the credit models registry to use lazy imports rather than importing every single model module every time CREDIT is loaded. 

Lets do that -- the model-zoo branch will bring in significantly many more models.

PR #332, opened 2026-04-16
add HRRR Dataset

charlie-becker:
### Description
Addition of a HRRR Dataset with new gen2 data scheme. Also includes a separate `hrrr_download.py` script to download the grib files (including idx files) directly which can be read by the Dataset. 

Features:
- Supports reading locally or streaming directly from AWS in grib format using pygrib. 
- HRRR versions 2 - 4
- Supports Pressure level data, Model level data (50 levels hybrid-sigma), and sub-hourly (15 minute 2D fields)
- Uses multiple threads to parallelize grib byte-range reads on top of CREDIT torch multi-processing
- Current variable registry covers quite a few variables, but not all, and adding more variables to the registry at the top of `hrrr.py` should be straight forward.
- Subsetting extent

Example config:

    HRRR_NAT: # model levels
      mode: "remote" 
      base_path: "/Users/cbecker/PycharmProjects/CREDIT/tests/data/" # only used for `mode: "local"`
      forecast_hour: 10          # 0 = analysis; >0 = forecast lead hour
      levels: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16] # first 16 model levels
      variables:
        prognostic:
          vars_3D: [ T, U, V, Q]
          vars_2D: [ t2m ]
        dynamic_forcing: null
        diagnostic: null
      extent: [ -130, -60, 20, 55 ]   # [min_lon, max_lon, min_lat, max_lat]

    HRRR: # Pressure levels
      mode: "remote" 
      forecast_hour: 10          # 0 = analysis; >0 = forecast lead hour
      levels: [ 250, 500, 700, 850, 925, 1000 ]   # hPa; omit for all available
      variables:
        prognostic:
          vars_3D: [ T, U, V, Q]
          vars_2D: [ t2m ]
        dynamic_forcing: null
        diagnostic: null
      extent: [ -130, -60, 20, 55 ]   # [min_lon, max_lon, min_lat, max_lat]

    HRRR_SUBH: # Sub-hourly (15 minute)
      mode: "remote" 
      forecast_hour: 10          # 0 = analysis; >0 = forecast lead hour
      variables:
        prognostic:
          vars_2D: [ t2m ]
        dynamic_forcing: null
        diagnostic: null
      extent: [ -130, -60, 20, 55 ]   # [min_lon, max_lon, min_lat, max_lat]

One other change slipped into this PR on the concat preblock which now simply adds more variable metadata that will be used downstream for postblocks to reconstruct dictionaries from model output tensors.


tests/test_hrrr.py
```diff
@@ -0,0 +1,325 @@
...
+
+def test_parse_idx_basic():
+    entries = _parse_idx(_IDX_TEXT)
+    assert len(entries) == 4
+    assert entries[0]["var"] == "TMP"
+    assert entries[0]["level"] == "500 mb"
+    assert entries[0]["step"] == "anl"
+    assert entries[0]["byte_start"] == 0
+    assert entries[0]["byte_end"] == 12344  # next entry start - 1
+    assert entries[-1]["byte_end"] is None  # last entry
```
danineamati, line 136: Maybe this is a bit unnecessary. But how do you feel about another test to check the last index as well. Rationale: 
1. guard against not all entries loaded. Although this is implicitly there with the length check, there hypothetically could be a case where all entries have incorrectly repeated values.
2. multiple spaces in "2 m above ground".

credit/datasets/hrrr_download.py
```diff
@@ -0,0 +1,240 @@
...
+After downloading, switch ``mode`` to ``"local"`` in the config.
+
+Usage::
+
+    python -m credit.datasets.hrrr_download -c config/my_conf.yaml --num-workers 8
+
+Or programmatically::
+
+    from credit.datasets.hrrr_download import download_hrrr
+    download_hrrr(config, num_workers=8, overwrite=False)
```
danineamati: Based on trying this out, I think it may be more clear to put this as `download_hrrr(config["data"], ...)` (i.e., see bottom of file).

(Update: please first see the comment below)

credit/datasets/hrrr_download.py
```diff
@@ -0,0 +1,240 @@
...
+    """Download HRRR grib2 + .idx files from AWS S3 to local disk.
+
+    Each timestamp is downloaded in parallel using a ``multiprocessing.Pool``.
+    Both the grib2 file and its ``.idx`` sidecar are fetched so that
+    ``HRRRDataset`` in ``mode: "local"`` can use fast byte-range reads.
+
+    Args:
+        config: Top-level ``data`` config dict (same object passed to
+            ``HRRRDataset``).
+        product: HRRR product to download — ``"wrfprsf"`` (pressure-level,
```
danineamati: (Updated)
If I may ask, why not pull the product from the configuration file?

Currently, in the configuration file, we have something like:
```shell
data:
  source:
    HRRR:
      mode: "remote"
      ...
    HRRR_NAT:
      mode: "remote"
      ...
```

So, the product information is _already_ in the key for the data source (e.g., `HRRR_NAT` --> `wrfnatf`). The issue seems to be that we are passing in the top-level `data` config rather than the `data/source/HRRR` or `data/source/HRRR_NAT`. But, even then, we could loop through the keys in `source` to find and download all the options. 

I do wonder if we want to break this function into sub functions:
1. HRRR common data field extraction
2. subhf specific handling
3. download pulling.

credit/datasets/hrrr.py
```diff
@@ -0,0 +1,957 @@
...
+
+VALID_FIELD_TYPES = {"prognostic", "diagnostic", "dynamic_forcing"}
+
+# V3+ S3 path includes a 'conus/' subdirectory; v1/v2 does not
+_HRRR_V3_CUTOFF = pd.Timestamp("2018-07-12")
+_S3_BUCKET = "noaa-hrrr-bdp-pds"
+# Public HTTPS base — used for Range requests (faster than s3fs seek+read)
+_HRRR_HTTPS_BASE = f"https://{_S3_BUCKET}.s3.amazonaws.com"
+
+#: Variable registry mapping user-facing names to HRRR ``.idx`` lookup keys.
```
kevinyang-cky, line 140: Personal request for additional variables.
- Land/Water Mask
- Simulated Brightness Temperatures

credit/datasets/hrrr.py
```diff
@@ -0,0 +1,957 @@
...
+        """
+        if product not in VALID_PRODUCTS:
+            raise ValueError(f"Unknown HRRR product '{product}'. Valid: {sorted(VALID_PRODUCTS)}")
+        source_cfg = config["source"][config_key]
+
+        self.product: str = product
+        self.source_name: str = _PRODUCT_SOURCE_NAMES[product]
+        self.return_target: bool = return_target
+        self.mode: str = source_cfg.get("mode", "local")
+        self.base_path: str | None = source_cfg.get("base_path", None)
```
kevinyang-cky, line 621: Just curious about the design philosophy here. In ERA5 and MRMS, the `path` is under `[source][variable][field_type]` in the config, whereas in the HRRR Dataset here, the `base_path` is just under `[source]`.

credit/datasets/multi_source.py
danineamati: Thought process from meeting is to pass in the config portion specifically for this source as `cls(source_cfg)` or similar to avoid additional arguments passed (e.g., for `HRRR_nat`).

credit/datasets/hrrr.py
charlie-becker: Thanks for flagging this. We should be consistent, but I'm not entirely sure how to handle it. I don't love having a separate path for every one of the data types for every dataset, and would prefer a single path for each dataset. 

I think the crux of this comes down to how we should handle loading things like dynamic forcings and statics that often aren't a part of the "dataset of interest", for example: solar radiation, land type, masking, etc. A simple scenario we can't currently handle: MRMS + terrain. The only way we could currently handle this is by using MRMS + "ERA5" static fields.

I don't necessarily think every dataset should need to handle other file paths for forcings / statics in the way we originally wrote ERA5 (similar to gen 1). 

Maybe this is where we incorporate a Base Dataset (as also proposed) that can handle "misc" data?

@djgagne

credit/datasets/hrrr.py
charlie-becker: done!

tests/test_hrrr.py
charlie-becker: Sure for 1. , I don't see 2. as a problem

credit/datasets/hrrr_download.py
charlie-becker: Yeah, I see what you mean. I'd rather not supply the product directly as an argument and derive it from the config. I'll make the change.

danineamati: In the new version, it just pulls directly from the key. No need for `product` or `config_key`.

credit/datasets/hrrr.py
```diff
@@ -0,0 +1,988 @@
...
+
+def _build_prs_entry_map(idx_entries: list[dict], idx_name: str) -> dict[float, dict]:
+    """Return a ``{pressure_level_hPa: idx_entry}`` dict for a pressure-level variable."""
+    result: dict[float, dict] = {}
+    for e in idx_entries:
+        if e["var"] == idx_name and e["level"].endswith(" mb"):
+            try:
+                lv_f = float(e["level"].replace(" mb", ""))
+            except ValueError:
+                continue
```
danineamati, line 377: How do we feel about adding a logging debug here or similar since this would otherwise be completely silent?

charlie-becker: Good idea

credit/datasets/hrrr.py
```diff
@@ -0,0 +1,988 @@
...
+    range_header = f"bytes={byte_start}-{byte_end}" if byte_end is not None else f"bytes={byte_start}-"
+    getter = session.get if session is not None else requests.get
+    try:
+        resp = getter(https_url, headers={"Range": range_header}, timeout=_HTTP_TIMEOUT)
+    except requests.exceptions.ConnectionError:
+        # AWS S3 closes idle keep-alive connections after ~20 s.  On the next
+        # request the session tries to reuse the stale socket and gets
+        # RemoteDisconnected.  One retry is enough — the second attempt opens
+        # a fresh connection.
+        resp = getter(https_url, headers={"Range": range_header}, timeout=_HTTP_TIMEOUT)
```
danineamati, line 363: Not sure if we want to have this in the coverage (identical to the line above) and not sure how to implement such a test.

danineamati:
Added tests for HRRR (Download and Dataset). Overall, we now have improved coverage:

```shell
> pytest --cov=credit/datasets/ --cov-report=term-missing
...
Name                                             Stmts   Miss  Cover   Missing
------------------------------------------------------------------------------
...
credit/datasets/hrrr.py                            325     25    92%   358-363, 505-509, 521-526, 539, 864-870, 902-903, 914, 945, 968
credit/datasets/hrrr_download.py                    92     32    65%   74-87, 111, 152-153, 210-213, 225-259
...
credit/datasets/multi_source.py                     36      2    94%   179, 184
```

The lower coverage for the HRRR Dataset file is due to not covering the CLI at the end of the file in the `__name__ == "__main__"` block. We also did not included integration tests for download and then load into the HRRR Dataset.

Additionally, `pygrib` will be a new dependency for CREDIT. 

----
Blocking issue: The Python 3.11 tests seem to be stuck after completing `tests/multi_source_dataset_test.py` and before completing the first test in `tests/output_test.py`. This issue is unclear to me.

danineamati (approved):
Pending Charlie and DJ's check, should now be all set to merge!

PR #333, opened 2026-04-17
Add Postblock Structure

charlie-becker:
### Description
This provides a postblock structure in the same fashion as preblocks that are called from the configuration file. 

Each postblock exists as its own module and are sequentially called in the order specified in the configuration file. The `Reconstruct` postblock does not need to be specified and is automatically applied as the first postblock. `Reconstruct` takes the output tensor from the model, and turns it into a self-describing dictionary similar to that that comes out of the data loader for ease of use in downstream post-blocks. 

In order to achieve this, metadata must be propagated from model input. To address this, the preblock `ConcatToTensor` (also always invoked) now also outputs dictionary metadata for later reconstruction. This includes the dictionary keys as well as the relevant slice indexes for each variable. 

The automatic invocation of `ConcatToTensor` and `Reconstruct` do not require specification in the config, and will also be invoked if either `preblocks:` or `postblocks:` (or both) are missing from the config (train / inference without pre/post blocks), further simplifying the configuration.

Postblocks also have the option to be applied once after a multistep rollout, or independently for each step within the rollout period. This is defaulted to being applied to every step along the way. To only apply that postblock to the end of the rollout, set `per_step: False` in the config under the postblock args.  

# Example Usage
### Data Portion of Config
```
data:
  source:
    ERA5:
      level_coord: "level"
      levels: [ 10,  30,  40,  50,  60,  70,  80,  90,  95, 100, 105, 110, 120, 130, 136, 137]
      variables:
        prognostic:
          vars_3D: ['U','V','T','Q']
          vars_2D: ['SP','t2m']
          path: '<path_to_files>*'
          filename_time_format: "%Y"

  timestep: "6h"
  forecast_len: 1
  start_datetime: "2021-01-01"
  end_datetime: "2021-12-31"
  
preblocks:
  log_transform:
    type: "log_transform"
    args:
      base: 'e'
      variables:
        - "era5/prognostic/3d/U"

postblocks: 
  inverse_scaling:
    type: "bridgescaler_transform"
    args:
      method: "inverse_transform"
      variables:
        - era5/prognostic/3d/U
```
### General Pipeline
```
dataset = MultiSourceDataset(config['data'], return_target=False)
sampler = DistributedMultiStepBatchSampler(dataset, batch_size=1, num_forecast_steps=1, shuffle=True, num_replicas=1, rank=0)
loader = iter(DataLoader(dataset, batch_sampler=sampler, num_workers=0, pin_memory=False, prefetch_factor=None))

preblocks = build_preblocks(config.get('preblocks'))
postblocks = build_postblocks(config.get('postblocks')

batch = next(loader)
batch, meta = apply_preblocks(preblocks, batch)
preds = Model(batch)
preds = apply_postblocks(postblocks, preds, meta)
```
Output will now be self-describing tensors:

`preds['predictions'].keys() --> ['era5/prognostic/3d/T', 'era5/prognostic/3d/U', ...]`

### Related Issues and PRs

Closes #XXX

### Checklist


credit/postblock/__init__.py
```diff
@@ -1,19 +1,72 @@
-from credit.postblock._postblock import (
```
djgagne, line 1: How are we going to handle the legacy postblocks? This PR currently ignores them, and removing this import appears to be breaking some downstream tests.

djgagne (changes requested):
Gen 1 Postblocks still need to work in Gen 2 but potentially with some demarcation. Currently Gen 1 postblocks are put in the model part of the config, so they could stay there and gen 2 postblocks would be part of the postblock section of the config.  I do suggest implementing similar importlib tricks that I used in #358 for the postblocks so we don't have too many required imports.

djgagne (approved):
All tests passing now. Good to merge.

Issue #335, opened 2026-04-18
Overly large metadata files in credit.metadata


djgagne:
I discovered that there are two very large netcdf files that have been added to credit.metadata at some point and appear to be the cause of why the miles-credit build tarball was not uploaded to PyPI (PyPI limits the size of file uploads to 100 MB). The culprits:
* CESM_Lev_Info.nc: 50 MB
* conus404.3d.dims.nc: 11 MB
* conus404.2d.dims.nc: 3.7 MB 

We will need to not only remove these files from the metadata directory but also edit the git history of the main branch to purge them. Doing so should speed up git clones and installs more generally.

kafitzgerald:
It should also be possible to exclude some of the files from the packaging for PyPI if that's helpful in the near term.  Relevant docs here: https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html 

Using something like [git-filter-repo](https://github.com/newren/git-filter-repo/) is probably the way to go for removing the files, but it's still non-trivial.  There's some helpful guidance from GitHub on this topic (including how to and potential challenges / side effects):
* https://docs.github.com/en/repositories/working-with-files/managing-large-files/about-large-files-on-github#removing-a-file-that-was-added-in-an-earlier-commit
* https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository

Sorry if this is old news.

kafitzgerald:
I'm fairly sure this (i.e. [`nbstripout`](https://github.com/kynan/nbstripout)) is the tool I was thinking of for stripping outputs from notebooks.  It can be configured a number of different ways including as a GitHub action. 

[Jupytext](https://jupytext.readthedocs.io/) text based notebooks are another option.  The text-based nature also makes them easier/cleaner to review and diff and there's a conversion tool to go to/from common notebooks formats.

That said, if there aren't that many notebooks and/or people are good about clearing them before committing, I'd just ignore all of this.

PR #338, opened 2026-04-23
Deleting old notebooks and large files in repo.

djgagne:
### Description

I have deleted a bunch of old notebooks that no longer appear to be used and either pared down or deleted multi-MB netCDF files scattered throughout the repo. The total size of the repo (excluding git commit history) is now 14 MB. I used PyCharm check file usages before deleting anything. Other larger netcdf files have been copied to `/glade/campaign/cisl/aiml/credit/metadata/` if you happen to need them for anything. 

I did look into using `git filter-repo`, which does have a handy option to filter out files larger than a certain size from the entire commit history. Unfortunately, doing this would result in significant disruption to current PRs and active branches, so we should wait to do this until a less active period of development. In the meantime, removing the files at least makes the next release to PyPI more straightforward.

### Related Issues and PRs

Relates to #335 

### Checklist


kafitzgerald (approved):
Looks good to me though I am not the author of any of these scripts or notebooks and don't have a good feel for if anything would be good to keep so someone else should sign off on that.  

It might also be worth leaving the associated issue open until it's fully resolved (i.e. changing "Closes" to "Relates to" or similar.  Your call on that though and probably depends on how much of a priority fully removing these is.

Also, I don't know if this is part of the issue here, but there are tools I believe to check for pre-run notebooks (which can be quite large).  Jupytext and other text-based notebook formats are also helpful in that respect.  Making sure not to commit pre-run notebooks also works, but is sometimes easy to do and hard to catch 🙃.

danineamati:
Is the intention to delete _all_ the notebooks or simply clear output on all the notebooks? For example, `rollout_vit2d.ipynb` was large primarily because the commit included all the notebook outputs (e.g., plots inside the notebook). Granted, I am a fan of minimal or no `.ipynb` files, but just wanted to check.

charlie-becker (approved):
Good purge. There might be a couple of these notebooks that might be still useful (CONUS 404 ?), but I do think these could be recovered from other branches.

djgagne:
Thanks for the feedback @kafitzgerald. I will leave the issue open and revisit cleaning up the whole git history once we are in a less active development period. If you could share a GH Action or other code example on pre-running the notebooks or nbconverting them to remove output automatically, that would be great. The notebooks in the repo were more experimental and not meant to be public facing, so clearing them out should not be an issue. Going ahead and merging and moving forward with the next release.

PR #353, opened 2026-04-29
MSLPDiagnostic postblock

### LINKED ISSUE #347: Mean sea level pressure postblock
Create a mean sea level pressure postblock. It requires surface pressure, geopotential, and temperature at lowest model level.

jsschreck:
## Summary

- Adds `credit/postblock/mslp.py` with `mslp_from_surface_pressure()` — a fully vectorized PyTorch implementation of the Trenberth et al. (1993) formula. No numpy, no per-pixel loops.
- Adds `MSLPDiagnostic(nn.Module)` following the V2 data-dict postblock protocol: named variable-name constructor params, `forward(data_dict)` with nested dict access, no channel indices, no denormalization. Registered in `POSTBLOCK_REGISTRY` as `"mslp_diagnostic"`.
- 16 tests in `tests/test_mslp_postblock.py` covering all three Trenberth 1993 branches, batch broadcasting, the near-flat ocean case, and a regression test for the `tto` units bug fixed in #341.

## Physics

Three lapse-rate regimes from Trenberth et al. (1993):

1. **near-flat** (`|height| < 1e-4 m`): MSLP = SP
2. **cold surface / warm sea level** (`T ≤ 290.5 K`, `T_sl > 290.5 K`): `α = Rd·(290.5−T)/Φ`
3. **warm surface** (`T > 290.5 K`): `α = 0`, `T_eff = ½(290.5 + T)`
4. **very cold surface** (`T < 255 K`): `α = Γ·Rd/g`, `T_eff = ½(255 + T)`
5. **default**: `α = Γ·Rd/g`, `T_eff = T`

Spot-checked: Denver (~1600 m, SP=840 hPa) → 1015.7 hPa; sea level → 1013.25 hPa exactly.

## Config

```yaml
postblocks:
  mslp:
    type: mslp_diagnostic
    args:
      surface_pressure_var: "ARCO_ERA5/prognostic/2d/surface_pressure"
      temperature_var: "ARCO_ERA5/prognostic/2d/2m_temperature"
      surface_geopotential_var: "ARCO_ERA5/static/2d/geopotential_at_surface"
      output_name: "ARCO_ERA5/derived_diagnostic/2d/mean_sea_level_pressure"
```

## Test plan

- [x] `pytest tests/test_mslp_postblock.py -v` — 16 tests, all pass
- [x] CI passing on Python 3.11 / 3.12 / 3.13
- [x] Physics verified against ISA reference values (1013–1070 hPa over 0–4000 m terrain)

### DIFF
--- a/credit/postblock/__init__.py
+++ b/credit/postblock/__init__.py
@@ -3,7 +3,8 @@ import torch.nn as nn
 from credit.postblock.reconstruct import Reconstruct
 from credit.postblock.wet_mask_samudra import WetMaskBlock
 from credit.postblock.scaler import BridgeScalerTransformer
-from credit.postblock.mslp import MSLPDiagnostic, mslp_from_surface_pressure
+from credit.postblock.mslp import MSLPDiagnostic
+from credit.postblock.mslp import mslp_from_surface_pressure as mslp_from_surface_pressure
 
 
 POSTBLOCK_REGISTRY = {
--- a/credit/postblock/__init__.py
+++ b/credit/postblock/__init__.py
@@ -3,12 +3,14 @@ import torch.nn as nn
 from credit.postblock.reconstruct import Reconstruct
 from credit.postblock.wet_mask_samudra import WetMaskBlock
 from credit.postblock.scaler import BridgeScalerTransformer
+from credit.postblock.mslp import MSLPDiagnostic, mslp_from_surface_pressure
 
 
 POSTBLOCK_REGISTRY = {
     "reconstruct": Reconstruct,
     "bridgescaler_transform": BridgeScalerTransformer,
     "wet_mask_samudra": WetMaskBlock,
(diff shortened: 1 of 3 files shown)


djgagne:
Our plan with the postblock structure is to immediately bring the data output from the model back into the same dictionary schema as what goes into concatenate on the preblock side. That way you can pass exact variable keys into each subsequent postblock rather than having to track indices, which could break if the user changes the data properties.

jsschreck:
@djgagne i will update this when the post-block PR gets updated and merged. The structure that you guys want to impose needs to be derivable nateively from the code

jsschreck:
@djgagne I am going to rename the class here MSLPDiagnostic to match your name for Z ... hoping to pin down a naming convention before we add more.

djgagne (approved):
Looks good to me. Merging.

PR #358, opened 2026-05-04
Added registry and ability to add custom models from outside of CREDIT.

djgagne:
### Description

I have refactored the credit.models init to create a model registry dictionary that lazily imports models when they are needed rather than importing all models any time credit.models is imported. This will significantly speed up the initialization of all credit applications and make it easier to add more models in the future.

I have also added a register_model function that will allow users to either decorate pytorch model classes  to add them to the registry or they can add the path to the module in their credit config file under a new `custom_models` section. This change will allow us to move toward users using credit more as an importable library rather than editing code themselves.

### Related Issues and PRs

None.

### Checklist

### DIFF
--- a/credit/parser.py
+++ b/credit/parser.py
@@ -11,7 +11,6 @@ Content:
 import os
 import copy
 import inspect
-import importlib.util
 import warnings
 from glob import glob
 from collections import Counter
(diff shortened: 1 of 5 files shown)


credit/parser.py
jsschreck, line 1: I was not trying to use parser.py anymore right inside credit/ in V2.

jsschreck (changes requested):
Looks good to me. I will update model-zoo PR using the new registry (its super annoying adding more as you are aware, esp once the list gets pretty large). Do we still need to use parser.py for V2? I am still hoping to kill it entirely? Ignore my request changes otherwise.

jsschreck (approved):
Looks good!

PR #362, opened 2026-05-05
fix: Derecho CLI submit — GPU env vars + explicit torchrun paths

jsschreck:
## Summary

- **Multi-node training**: Added CUDA_VISIBLE_DEVICES and cray-mpich/NCCL/libfabric env vars to multi-node path in both `_submit.py` and `pbs.py`.
- **Single-node torchrun path**: Now uses `{conda_env}/bin/torchrun` explicitly. Falls back to `_find_torchrun()` when conda_env is not a directory.
- **`pbs.py` multi-node**: `launch_script_torchrun` now emits `mpiexec -n N --ppn G python` for multi-node jobs.
- **Fix `os.path.isdir(None)` TypeError (Python 3.11+)**: Added None guard in all three torchrun-path expressions in `_submit.py`. This was causing 21 test failures in CI.
- **Remove stale rdzv tests**: Removed three `TestDerechoMultiNode` tests asserting torchrun rdzv flags that no longer appear in the mpiexec-based script.

## Test plan

- [x] All 21 previously failing `test_pbs_scripts.py` tests now pass locally
- [x] `pre-commit` hooks pass
- [x] CI green on python 3.11 / 3.12 / 3.13


credit/pbs.py
```diff
@@ -205,6 +205,133 @@ def launch_script_mpi(config_file, script_path, launch=True, backend="nccl"):
...
+export NCCL_CROSS_NIC=1
+export NCCL_NCHANNELS_PER_NET_PEER=4
+export NCCL_NET="AWS Libfabric"
+export NCCL_NET_GDR_LEVEL=PBH
+export FI_CXI_DISABLE_HOST_REGISTER=1
+export FI_CXI_OPTIMIZED_MRS=false
+export FI_MR_CACHE_MONITOR=userfaultfd
+export FI_CXI_DEFAULT_CQ_SIZE=131072
+
+echo "=== CREDIT v2 parallelism ==="
```
djgagne: Chagne to gen 2 parallelism?

credit/cli/_submit.py
```diff
@@ -210,6 +214,22 @@ def _build_pbs_script(
                 echo "Head node : ${{head_node_ip}}"
                 MASTER_PORT=$(( RANDOM % 10000 + 20000 ))
 
+                export CUDA_VISIBLE_DEVICES={cuda_devices}
+                export MPICH_GPU_SUPPORT_ENABLED=1
```
djgagne: Everything but CUDA_VISIBLE_DEVICES should be auto specified in the conda environment upon activation, for example in `/glade/campaign/cisl/aiml/credit/conda_envs/credit-derecho/etc/conda/activate.d/nccl_activate.sh`. CUDA_VISIBLE_DEVICES is unset in that file, so setting here is highly recommended.

credit/pbs.py
```diff
@@ -205,6 +205,133 @@ def launch_script_mpi(config_file, script_path, launch=True, backend="nccl"):
...
+#PBS -k eod
+
+module --force purge
+module load ncarenv/24.12 nvhpc cuda/12.3.2 cray-mpich conda
+
+{torchrun_line}export PYTHONPATH="{os.path.dirname(str(script_path))}:${{PYTHONPATH:-}}"
+export LOGLEVEL=INFO
+export NCCL_DEBUG=WARN
+export CUDA_VISIBLE_DEVICES={cuda_devices}
+export MPICH_GPU_SUPPORT_ENABLED=1
```
djgagne: See my other comment on where these are already set.

djgagne (changes requested):
iIrecommend removing variables other than CUDA_VISIBLE_DEVICES because these are already set upon environment activation.

credit/pbs.py
jsschreck: Getting ahead of myself. (removing)

jsschreck:
@djgagne thanks for the spot -- I removed all the exports

djgagne (approved):
Changes look good now. Merging.

PR #364, opened 2026-05-05
Pysteps

### LINKED ISSUE #304: Delete `pysteps` dependency
Only my ensemble verification suite uses pysteps. Since i plan to refactor it anyways in the next few months, please feel free to delete pysteps, especially as its caused dependency headaches in the past.

Plan to replace with [scores](https://github.com/nci/scores) package since it it much more fleshed out now. and has very few dependencies: `dependencies = ["xarray", "pandas", "scipy", "bottleneck"]`

djgagne:
### Description

I have removed pysteps as a dependency and took the rank histogram call out of @dkimpara 's ensemble evaluate script and ensemble.py. I looked at adding scores as a dependency but it downgraded pandas to before v3 as soon as I installed it. I would rather not add dependencies to CREDIT that are only eval and have people do evaluation in their own pipelines outside of CREDIT. 

### Related Issues and PRs

Closes #304 

### Checklist

### DIFF
--- a/applications/ensemble_eval.py
+++ b/applications/ensemble_eval.py
@@ -23,7 +23,7 @@ import xarray as xr
 import yaml
 
 from credit.pbs import launch_script, launch_script_mpi, get_num_cpus
-from credit.verification.ensemble import binned_spread_skill, rank_histogram_apply, spread_error
+from credit.verification.ensemble import binned_spread_skill, spread_error
 from credit.verification.standard import average_div_rot_spectrum, average_zonal_spectrum
 from credit.xr_sampler import XRSamplerByYear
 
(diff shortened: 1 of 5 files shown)


kafitzgerald:
Should we drop the note from the docs as well?  https://github.com/NCAR/miles-credit/blob/26528cb0d5945fdcfb02d73ec83355f10d599f01/docs/source/getting-started.md?plain=1#L69-L73

It looks like `pysteps` is also specified in the CIRRUS testing Dockerfile, but it might be good to keep that in sync with what's being used currently so I don't know if it makes sense to tackle that now.  
https://github.com/NCAR/miles-credit/blob/26528cb0d5945fdcfb02d73ec83355f10d599f01/tests/CI_CIRRUS/Dockerfile#L26-L28

djgagne:
Addressed docs and dockerfile. The docker container does not get rebuilt for each CI run, but we might want to rebuild it once the Gen 2 dependency structure is settled.

kafitzgerald (approved):

PR #367, opened 2026-05-06
Change preblock output structure

charlie-becker:
### Description
Changes the output structure of `apply_preblocks()` to a single dictionary with keys `input`, `meta`, and `target` (if present) and adjusts all function calls accordingly. Also adds better error handling.

### Checklist
- [ x] I am familiar with the [contributing guidelines](https://miles-credit.readthedocs.io/en/latest/contrib.html).

### DIFF
--- a/applications/rollout_realtime_gen2.py
+++ b/applications/rollout_realtime_gen2.py
@@ -337,8 +337,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
     # ---- Load full initial state ----
     sample_full = dataset[(init_time, 0)]
     batch_full = _sample_to_batch(sample_full)
-    x, _ = apply_preblocks(preblocks, batch_full)
-    x = x.to(device).float()  # (1, C_in, 1, H, W)
+    x = apply_preblocks(preblocks, batch_full)["input"].to(device).float()  # (1, C_in, 1, H, W)
 
     results = []
     x_init = None
@@ -374,8 +373,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
                 t_next = init_time + step * dt
                 sample_frc = dataset[(t_next, 1)]  # only dynamic_forcing
                 batch_frc = _sample_to_batch(sample_frc)
-                x_frc, _ = apply_preblocks(preblocks, batch_frc)
-                x_frc = x_frc.to(device).float()
+                x_frc = apply_preblocks(preblocks, batch_frc)["input"].to(device).float()
 
                 # ERA5Dataset insertion order: [dynfrc | static | prog]
                 x[:, :n_dyn, ...] = x_frc
--- a/applications/rollout_realtime_gen2.py
+++ b/applications/rollout_realtime_gen2.py
@@ -337,7 +337,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
     # ---- Load full initial state ----
     sample_full = dataset[(init_time, 0)]
     batch_full = _sample_to_batch(sample_full)
-    x = apply_preblocks(preblocks, batch_full)["input"].to(device).float()  # (1, C_in, 1, H, W)
+    x = apply_preblocks(preblocks, batch_full, device=device)["input"].float()  # (1, C_in, 1, H, W)
 
     results = []
     x_init = None
@@ -373,7 +373,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
                 t_next = init_time + step * dt
                 sample_frc = dataset[(t_next, 1)]  # only dynamic_forcing
                 batch_frc = _sample_to_batch(sample_frc)
-                x_frc = apply_preblocks(preblocks, batch_frc)["input"].to(device).float()
+                x_frc = apply_preblocks(preblocks, batch_frc, device=device)["input"].float()
 
                 # ERA5Dataset insertion order: [dynfrc | static | prog]
                 x[:, :n_dyn, ...] = x_frc
--- a/applications/rollout_to_netcdf_gen2.py
+++ b/applications/rollout_to_netcdf_gen2.py
@@ -280,7 +280,7 @@ def predict(rank, world_size, conf, p):
             # Step 0: load full initial state
             sample_full = dataset[(t0, 0)]
             batch_full = _sample_to_batch(sample_full)
-            x = apply_preblocks(preblocks, batch_full)["input"].to(device).float()  # (1, C_in, 1, H, W)
+            x = apply_preblocks(preblocks, batch_full, device=device)["input"].float()  # (1, C_in, 1, H, W)
 
             x_init = None
             results = []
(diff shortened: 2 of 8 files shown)


applications/rollout_realtime_gen2.py
```diff
@@ -374,8 +373,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
                 t_next = init_time + step * dt
                 sample_frc = dataset[(t_next, 1)]  # only dynamic_forcing
```
djgagne, line 367: Is frc forecast or forcing? Can you change to a more descriptive variable name?

applications/rollout_realtime_gen2.py
```diff
@@ -337,8 +337,7 @@ def run_forecast(conf, init_time: pd.Timestamp, n_steps: int, save_dir: str, poo
     # ---- Load full initial state ----
     sample_full = dataset[(init_time, 0)]
     batch_full = _sample_to_batch(sample_full)
-    x, _ = apply_preblocks(preblocks, batch_full)
-    x = x.to(device).float()  # (1, C_in, 1, H, W)
+    x = apply_preblocks(preblocks, batch_full)["input"].to(device).float()  # (1, C_in, 1, H, W)
```
djgagne: This call assumes that concat is included and the output of x["input"] is a single tensor. Is that always true if concat is not included in the preblock cycle? At the very least we need to do a type check before passing to device. Converting to float automatically also could be problematic if the user wants to use a higher or lower precision. 

I am being a stickler about this because I expect future models in CREDIT to accept subsets of variables as inputs or have different heads or other combos that would benefit from the dictionary framework.

applications/rollout_to_netcdf_gen2.py
```diff
@@ -280,8 +280,7 @@ def predict(rank, world_size, conf, p):
             # Step 0: load full initial state
             sample_full = dataset[(t0, 0)]
             batch_full = _sample_to_batch(sample_full)
-            x, _ = apply_preblocks(preblocks, batch_full)
-            x = x.to(device).float()  # (1, C_in, 1, H, W)
+            x = apply_preblocks(preblocks, batch_full)["input"].to(device).float()  # (1, C_in, 1, H, W)
```
djgagne: See my related comment.

credit/preblock/__init__.py
```diff
@@ -53,7 +53,31 @@ def build_preblocks(preblock_cfg: dict | None = None) -> nn.ModuleDict:
...
+        raise RuntimeError(
+            'No preblocks configured. "preblocks" must be present in the config, with "concat" as the final preblock.'
+        )
+    meta = None
+    target = None
     for preblock in preblocks.values():
-        batch = preblock(batch)
-    return batch
+        result = preblock(batch)
+        if isinstance(result, tuple):
```
djgagne, line 72: Is this just a workaround until we standardize preblock outputs?

credit/trainers/trainerERA5gen2.py
```diff
@@ -172,7 +171,8 @@ def train_one_epoch(self, epoch, trainloader, optimizer, criterion, scaler, sche
 
             for t in range(1, self.forecast_len + 1):
                 batch = next(dl)
-                x_raw, y_raw, _ = ConcatToTensor()(apply_preblocks(self.preblocks, batch))
+                _batch = apply_preblocks(self.preblocks, batch)
+                x_raw, y_raw = _batch["input"], _batch["target"]
                 # ERA5Dataset outputs 5D tensors (B, C, frames, H, W); collapse frames dim
```
djgagne, line 172: Is the frames dim collapse something that should be more generically applied or something that should be removed?

djgagne (changes requested):
Have some questions about new implementation here.

applications/rollout_realtime_gen2.py
charlie-becker: I tried to be minimally invasive to the inherited code. We can change this, but maybe it's better suited for a separate PR focused on python variable naming?

credit/trainers/trainerERA5gen2.py
charlie-becker: I think it can be removed, but I'll have to look into this further.

credit/preblock/__init__.py
charlie-becker: No. Currently, all preblocks except for ConcatToTensor return a single dictionary with the "meta keys". But ConcatToTensor() returns a tuple: either `(input_tensor, output_tensor, meta_dict)` or `(input_tensor, meta_dict)`. So I think the logic is correct.

charlie-becker: It's then merged into a dictionary inside `apply_preblocks()` as it should always be necessary to concatenate.

applications/rollout_realtime_gen2.py
charlie-becker: I don't quite understand the use case for not having a concat in the preblock (and `apply_preblocks()` will actually raise an error if if not present), so yes it does assume concatenation. If this is not always true, we'll definitely have to modify some things.

charlie-becker:
Ready for re-review.

credit/preblock/__init__.py
```diff
@@ -52,8 +53,41 @@ def build_preblocks(preblock_cfg: dict | None = None) -> nn.ModuleDict:
...
+
+    Returns a dict with keys:
+        "input"  — concatenated tensor (if concat preblock present) or nested batch dict
+        "meta"   — metadata dict (only present if concat preblock ran)
+        "target" — target tensor (only present if target data was in the batch)
+
+    Tensors in the output are moved to ``device`` unless the concat preblock was
+    configured with ``to_device: false``.
+    """
+    if not preblocks:
```
jsschreck: Nice cleanup overall Charlie, the dict return is much cleaner than the old tuple pattern. One thing to think about on this line — raising when `preblocks` is empty is a breaking change for any config that currently omits the `preblocks:` key entirely (V1 configs especially). Before this, `apply_preblocks` was a no-op for the empty case. Could we keep that behavior and only raise if preblocks are configured but malformed? Otherwise any existing config without preblocks will crash at runtime.

credit/cli/_plot.py
```diff
@@ -150,9 +150,8 @@ def _plot(args) -> None:
     sample = dataset[(ts, 0)]
     batch = default_collate([sample])
     preblocks = build_preblocks(conf.get("preblocks", {}))
-    x, y, _ = apply_preblocks(preblocks, batch)
-    x = x.to(device)
-
+    _batch = apply_preblocks(preblocks, batch, device=device)
+    x, y, meta = _batch["input"], _batch["target"], _batch["meta"]
```
jsschreck: Quick heads up here: `_batch["target"]` and `_batch["meta"]` are only set conditionally in `apply_preblocks` (when the concat preblock actually produces them). If the preblock chain doesn't return a target this will throw a `KeyError`. Probably worth using `.get("target")` or guarding with `if "target" in _batch` since the plot path can be called in inference-only settings.

credit/preblock/__init__.py
```diff
@@ -52,8 +53,41 @@ def build_preblocks(preblock_cfg: dict | None = None) -> nn.ModuleDict:
...
-    return batch
+        result = preblock(batch)
+        if isinstance(result, tuple):
+            if len(result) == 3:
+                batch, target, meta = result
+            else:
+                batch, meta = result
+        else:
+            batch = result
+        if hasattr(preblock, "to_device"):
```
jsschreck: The `to_device` flag is getting read from whichever preblock has the attribute last in the loop, so if `ConcatToTensor` isn't the final preblock in the chain the flag could get silently overridden. Would be a bit safer to check `isinstance(preblock, ConcatToTensor)` directly, or pull the flag before the loop from the preblock that owns it.

charlie-becker:
@jsschreck Thanks for flagging these -- they have all been addressed. If tests pass, I would suggest merging if there are no other concerns.

djgagne (approved):
Looks good now. Merging.

PR #369, opened 2026-05-07
Geopotential

### LINKED ISSUE #343: Geopotential postblock
Add a postblock that calculates geopotential from surface pressure, temperature, and specific humidity on model levels.

### LINKED ISSUE #361: Add WeatherBench 2 ERA5Dataset
I would like to add a variation of the ERA5 Datasets that targets WeatherBench2 zarr files on Google Cloud specifically so that users can train off of any of their resolutions, including the 32x64 option. This would allow us to benchmark training more easily and ideally submit our own weatherbench benchmark submissions. It should be a straightforward modification of the ARCOERA5 dataset.

djgagne:
### Description

This PR adds a torch geopotential postblock to calculate geopotential in every dataset column with vmap. I have also added a WeatherBench ERA5 Dataset and made sure ARCO ERA5 is properly linked with MultiSource Dataset. 

The geopotential postblock does use permute and reshape to reorder data variables into (num columns, num levels) shape for easy vmap deployment and reconstruction into the 3D array. The reconstruction is really close to ARCO ERA5 geopotential except for some small regridding artifacts. 

### Related Issues and PRs

Closes #343 
Closes #361 

### Checklist


credit/postblock/geopotential.py
```diff
@@ -0,0 +1,124 @@
...
+import xarray as xr
+from credit.metadata import get_meta_file_path
+
+
+def pressure_on_interfaces(
+    surface_pressure: torch.Tensor,
+    model_a_half: torch.Tensor,
+    model_b_half: torch.Tensor,
+    model_top_pressure: float = 0.57,
+):
```
charlie-becker, line 14: Why is the float here 0.57? Docstring says it defaults to 1 Pa.

credit/postblock/geopotential.py
```diff
@@ -0,0 +1,124 @@
...
+        geopotential (torch.Tensor): geopotential on each model level (m**2 s**-2)
+    """
+    RDGAS = 287.06
+    RVGAS = 461.51
+    GAMMA = RVGAS / RDGAS - 1.0
+    pressure_inter = pressure_on_interfaces(surface_pressure, model_a_half, model_b_half)
+    pi_upper = torch.flip(pressure_inter[:-1], (0,))
+    pi_lower = torch.flip(pressure_inter[1:], (0,))
+    dlogp = torch.log(pi_lower / pi_upper)
+    alpha = 1.0 - ((pi_upper / (pi_lower - pi_upper)) * dlogp)
```
charlie-becker, line 74: Is the pressure level indexing (low--> high) consistent amongst all (most) datasets? Is it worth checking this and supporting both?

credit/postblock/geopotential.py
```diff
@@ -0,0 +1,124 @@
...
+        pred_shape = pred[self.temperature_var].shape
+        pred_flat = {}
+        for input_var in [
+            self.surface_geopotential_var,
+            self.surface_pressure_var,
+            self.temperature_var,
+            self.specific_humidity_var,
+        ]:
+            pred_per = torch.permute(pred[input_var], (0, 2, 3, 4, 5, 1))
+            pred_flat[input_var] = pred_per.reshape(np.prod(pred_per.shape[:-1]), pred_per.shape[-1])
```
charlie-becker: What is the 6D data represent here? I thought we mostly worked with 5D data.

djgagne: It was due to a bug in Reconstruct. I rewrote so that the permute is independent of the number of data dimensions, so we can support both rectilinear and non-uniform meshes with the same function.

credit/postblock/geopotential.py
djgagne: ERA5 and CESM both use the low pressure to high pressure convention, but I could make this a config option for datasets that happen to be oriented the other way.

credit/postblock/geopotential.py
djgagne: I tweaked the model_top_pressure to make the geopotential values line up better with ERA5. The doc wasn't updated.

credit/postblock/geopotential.py
djgagne: I added a kwarg to not run the flips since WRFout vertical level convention is bottom to top.

djgagne:
@charlie-becker I have updated the Geopotential postblock to align with the updated data dictionary schema and addressed your comments. I also converted credit.transforms and credit.trainers to lazy imported registries, which should speed initialization significantly and reduce some of the warnings CREDIT has been producing on startup. Please take a look and approve if you think it looks good.

credit/datasets/base_dataset.py
```diff
@@ -228,9 +228,24 @@ def __init__(self, data_config: dict[str, Any], return_target: bool = False) ->
         # Placeholder for static metadata
         self.static_metadata: dict[str, Any] = {}
 
+        # If the engine argument for xr.open_dataset needs to be specified, add the engine option to your data source config
+        if "engine" in self.curr_source_cfg:
+            self.engine = self.curr_source_cfg["engine"]
+        else:
+            self.engine = None
+
```
charlie-becker, line 236: Thanks for adding the engine parameter.

charlie-becker (approved):
Thanks for addressing the comments. I'm glad we have the new WeatherBench Dataset, clean imports, and the geopotential postblock. The dictionary structure looks good. 

My only other comment would be the relatively long test that only checks the shape and presence of NaNs. However, I don't have a good physical assertion off of the top of my head and it looks like you manually checked physical consistency according to the test doc. I approve.

PR #373, opened 2026-05-07
Refactor postblock pipeline to use shared dictionary

### LINKED ISSUE #323: Append Metadata within pre/post blocks
We have considered renaming variables in the pre-block transformation, such as normalization. For example `era5/prognostic/3d/U` to `era5/prognostic/3d/U_norm`, but this will likely have downstream complications with variable naming in other pre/post blocks and diagnostics.

This proposes appending to the metadata dictionary which variables have been modified by which preblock without renaming. This should be straightforward to implement inside each pre/post block as the metadata is always accessible from the state dictionary.

There are other cases where we might need the raw (un-transformed) data propagated  for some postblock diagnostics (like conservation) which this wouldn't necessarily handle. I'm not exactly sure how we'll handle that, but for now, including metadata of transformations is still probably a good idea.

charlie-becker:
### Description

  - apply_postblocks now takes a single batch_dict instead of separate (y_pred, metadata) args. The caller assembles the dict before calling — at minimum "prediction" (flat model output) and "meta" (from
   apply_preblocks), plus any optional keys like "input", "target".                                                                                                                             
  - Reconstruct.forward updated to read/write batch_dict["prediction"] in-place, using batch_dict["meta"]["_channel_map"]["output"] for the channel map.                                                                                                                                                                  
  - BridgeScalerTransformer and WetMaskBlock gain a key="prediction" arg to scope which sub-dict they operate on.                                                                                          
                                                                                                                                                                                                                                                              
  Note: apply_postblocks has no active call sites yet — wiring into the trainer and rollout scripts is a follow-on step. 

Closes #323

### Checklist

### DIFF
--- a/credit/postblock/__init__.py
+++ b/credit/postblock/__init__.py
@@ -1,4 +1,3 @@
-import torch
 import torch.nn as nn
 
 from credit.postblock.reconstruct import Reconstruct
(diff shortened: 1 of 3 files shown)


djgagne (approved):
LGTM.

PR #376, opened 2026-05-08
Base dataset for CREDIT

### LINKED ISSUE #337: Simple abstract CREDIT Dataset versus Torch Dataset for python type hinting
## Primary Issue in Type
There are a few places throughout CREDIT that type hint to a Torch Dataset, but expect attributes that are CREDIT specific and not available in a Torch Dataset. Below are some examples:

https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/datasets/multi_source.py#L99-L111

Notice that `ds` is of type Dataset (imported as `from torch.utils.data import Dataset`), but the attribute `ds.static_metadata` is not a property of a Torch Dataset, yielding a Pylance issue:

```shell
Cannot access attribute "static_metadata" for class "Dataset[Unknown]"
  Attribute "static_metadata" is unknownPylance[reportAttributeAccessIssue](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportAttributeAccessIssue.md)
```

Similarly, later, we have a request for `ds.datetimes`, which would be available for a CREDIT dataset, but not a Torch Dataset 

https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/datasets/multi_source.py#L144-L161

Yielding a similar Pylance issue:

```shell
Cannot access attribute "datetimes" for class "Dataset[Unknown]"
  Attribute "datetimes" is unknownPylance[reportAttributeAccessIssue](https://github.com/microsoft/pylance-release/blob/main/docs/diagnostics/reportAttributeAccessIssue.md)
```

Other place with this issue:

https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/samplers.py#L33-L34

## Possible Solution

We could make it such that all datasets inherit from a base CREDIT Dataset that looks something like:

```python
class BaseCREDITDataset(torch.utils.data.Dataset):
    """An Base Dataset that augments the Dataset in torch with variables that we expect later in the CREDIT pipeline"""
    def __init__(self, datetimes: pd.DatetimeIndex, static_metadata: dict):
        self.datetimes = datetimes
        self.static_metadata = static_metadata   
```

Granted, if we _do not_ expect all datasets to have this, then we need to be more restrictive in the multi source data loading or similar. It seems possible to do this with `Protocol` in the `typing` library.

## Related

Could be related, though I would think they are included in torch (perhaps the issue is the `Unknown` in `"Dataset[Unknown]"`):

https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/datasets/load_dataset_and_dataloader.py#L620-L624

(Basically in all the trainers)
https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/trainers/base_trainer.py#L262-L279

https://github.com/NCAR/miles-credit/blob/e56fbdadfcbb40a7075c8cb4cab4626a6423f9d2/credit/trainers/ic_optimization.py#L357-L369

danineamati:
### Description

Currently, each new Dataset is a new class, inheriting from torch, and developed to mirror the current datasets. If a convention/configuration changes or if you want to bring in new data, you need to start from torch. However, our datasets are not just torch datasets; they have certain consistent functionality and member variables that need to hold for the rest of the CREDIT pipeline to work. This PR addresses this issue by introducing two classes, `AbstractBaseDataset` and `BaseDataset`, that help ensure this consistency. The existing datasets now inherit from these classes, removing redundant code and hopefully smoothing the future development experience with consistent class instance typing. 

### Related Issues and PRs

Closes #337 

### Checklist
- [X] The documentation is (mostly) up to date.


djgagne:
When using the Any type, make sure to import Any from the typing library. Ruff is flagging these throughout your modified files.

config/gen_1/example/example-v2026.1.0_gen2.yml
```diff
@@ -0,0 +1,266 @@
+save_loc: /glade/derecho/scratch/$USER/credit_example_model/
+seed: 1000
+data:
+  source:
+    ERA5:
+      dataset_name: era5
+      level_coord: '''level'''
```
djgagne: Use no quotes or regular quotes, not triple quotes.

credit/datasets/__init__.py
```diff
@@ -3,13 +3,16 @@
 import glob
 import logging
 
+from credit.datasets.base_dataset import BaseDataset
```
djgagne: We should make the datasets lazy imports rather than directly importing them in the init. The additional imports here slow down initialization of the package.

credit/datasets/goes.py
```diff
@@ -204,8 +203,9 @@ class GOESDataset(Dataset):
         config: Top-level experiment configuration dictionary. The relevant
             sub-keys are:
 
-            - ``config["source"]["GOES"]``: GOES-specific settings.
+            - ``config["source"]["{USER_PROVIDED}"]``: user-provided source name.
 
+              - ``dataset_name`` (str): has to be "goes" to trigger this dataset class.
```
djgagne: Should it default to goes19 since that is the most recent GOES-E? Is there a way to combine data from goes 19 and 16 or from 17 and 18 for that matter?

credit/datasets/era5.py
```diff
@@ -1,23 +1,23 @@
...
 
 .. code-block:: python
 
     {
         "input": {
-            "era5/prognostic/3d/T":        tensor,  # (n_levels, 1, lat, lon)
-            "era5/prognostic/2d/SP":       tensor,  # (1,        1, lat, lon)
-            "era5/dynamic_forcing/2d/tsi": tensor,
-            "era5/static/2d/LSM":          tensor,
+            "{USER_PROVIDED}/era5/prognostic/3d/T":        tensor,  # (n_levels, 1, lat, lon)
```
djgagne: Is user provided a name for the dataset? Is this defined somewhere?

credit/datasets/era5.py
```diff
@@ -337,7 +207,8 @@ class ARCOERA5Dataset(Dataset):
 
         data:
           source:
-            ARCO_ERA5:
+            {USER_PROVIDED}:
+              dataset_name: "arco_era5"
```
djgagne: Should `dataset_name` be `dataset_type`? I think of user_provided as the name of the dataset.

djgagne (changes requested):
Added some naming comments. Also, can you add a more explanatory page in the docs/source directory on how the new dataset structure works?

credit/datasets/era5.py
danineamati: User-provided will be whatever the user decides to call that entry (e.g., `my_favorite_era5_subset`)

credit/datasets/__init__.py
danineamati: To be clear, you mean make it a lazy import for all datasets, not just BaseDataset, right?

config/gen_1/example/example-v2026.1.0_gen2.yml
kevinyang-cky: Sorry, this file was not intended for the PR. I was testing `credit convert` to see whether it adds the `dataset_name` (will be renamed to `dataset_type` later) key. However, the current `credit convert` prompt message may mislead the user into typing 'level' and result in triple quotes. I have modified the message to avoid confusion.

**Original**
```
level_coord (vertical coordinate name in your data files): 
  '' not found in file. Available dims: ['time', 'latitude', 'longitude', 'level']
```

**Modified**
```
level_coord (vertical coordinate name in your data files): 
  '' not found in file. Available options: time, latitude, longitude, level
```

credit/datasets/__init__.py
djgagne: Yes.

config/gen_1/example/example-v2026.1.0_gen2.yml
kevinyang-cky: (changes will reflect in the next push)

credit/datasets/era5.py
kevinyang-cky: All `dataset_name` changed to `dataset_type`.

credit/datasets/__init__.py
kevinyang-cky: Addressed. Lazy import like the `__init__.py` under `models/`.

config/gen_2/examples/base_dataset.yml
```diff
@@ -0,0 +1,14 @@
+# Example for base dataset.
+
+data:
+  source:
+    AnExample_Base:
```
charlie-becker: I don't think this is necessarily useful, as right now, there is no current use case for calling the base dataset directly. Yet, anyways. See my comment below regarding changing this.

config/gen_2/examples/multi_source_data.yaml
```diff
@@ -25,6 +26,7 @@ data:
                     # single file — filename_time_format not needed
 
         MRMS:
+            dataset_type: "mrms"
```
charlie-becker, line 44: Can you add in another ERA5 dataset to this config so it demonstrates the use case of having more than one call of the same dataset? Adding an ERA5 using a couple of pressure levels would be a good use case.

credit/datasets/_utils.py
```diff
@@ -141,3 +142,25 @@ def _find_file(
     if idx >= 0 and t <= intervals[idx][1]:
         return intervals[idx][2]
     raise KeyError(f"No file found covering timestamp {t}. Check that your data files span the requested time range.")
+
+
+def _to_cftime(ts: pd.Timestamp, calendar: str) -> cftime.datetime:
+    """Convert a pandas Timestamp to a cftime.datetime.
+
+    Args:
```
charlie-becker, line 150: I think this function still exists in some of the individual datasets, why duplicate it?

credit/datasets/multi_source.py
```diff
@@ -40,12 +40,16 @@
...
 import pandas as pd
-from torch.utils.data import Dataset
 
+from credit.datasets.base_dataset import AbstractBaseDataset
+
+from credit.datasets.base_dataset import BaseDataset
 from credit.datasets.era5 import ERA5Dataset
-from credit.datasets.MRMS import MRMSDataset
+from credit.datasets.era5 import ARCOERA5Dataset
+from credit.datasets.mrms import MRMSDataset
```
charlie-becker, line 51: Please put all the classes inherited from the same module on the same import line.

scripts/base_dataset_example.py
```diff
@@ -0,0 +1,38 @@
...
+from credit.samplers import DistributedMultiStepBatchSampler
+
+from torch.utils.data import DataLoader
+
+BASE_DIR = pathlib.Path(__file__).resolve().parent
+config_path = BASE_DIR.parent / "config" / "gen_2" / "examples" / "base_dataset.yml"
+
+with open(config_path, "r") as f:
+    config = yaml.safe_load(f)
+
```
charlie-becker: Same comment as the first regarding the config. Since there are no current use cases for calling this directly (yet), there should not be a script example -- it is better suited for a pytest. However, if we do make this a direct feature (See final comment), we could potentially leave it.

charlie-becker (changes requested):
Thank you for both of your work on this! This is a nice upgrade. I left a couple of minor comments and a couple that are addressed in more detail below. 

1. Regarding the BaseDataset: I don't believe we have solved the example use case of: prognostic only from one data source (i.e. GOES) and dynamic forcing from another (i.e. Solar Insolation). It appears that `BaseDataset`'s `extract_field` is not functional currently. But I think it would be really nice (and likely necessary) to allow the BaseDataset to function independently for the wide variety of files that contain statics and forcings that probably don't deserve their own dataset. We could still keep the static and dynamic forcings in the other datasets (especially for the ones that actually have them), but could also potentially remove that part for the datasets that don't have them (i.e. static for GEOS. The latter is less important than the former though.
2. I believe the new preblock / postblock scheme will break with this new 5-delimiter naming convention: `data_name/data_source/data_type/data_dim/data_var`. They currently depend on the 4-delimiter scheme. You'll need to investigate preblock/postblock `__init__.py`, the `ConcaToTensor()` preblock and the `Reconstruct()` postblock to make sure they play nice with this schema. Let me know if you need help with this portion.

charlie-becker:
@kevinyang-cky 

While performing some more integration testing on this PR I found a bug in `goes.py` line 496. Please change to `self._fs` instead of `self.fs`

kevinyang-cky:
Thanks, @djgagne and @charlie-becker, for the comments. Once they are all addressed, we will tag you guys for re-review. Thanks!

credit/datasets/goes.py
kevinyang-cky: Addressed!

config/gen_2/examples/base_dataset.yml
kevinyang-cky: Remove for now.

credit/datasets/_utils.py
kevinyang-cky: `_to_cftime` no longer exists in individual datasets.

credit/datasets/multi_source.py
kevinyang-cky: Addressed!

scripts/base_dataset_example.py
kevinyang-cky: Remove it for now.

kevinyang-cky:
@djgagne @charlie-becker this PR is mostly ready for re-review. Two more minor to-dos, but major comments are addressed.

Main changes:

- Key name is `dataset_type` instead of `dataset_name`
- Use lazy import in `credit/datasets/__init__.py`
- GOES Dataset combines data records from satellites at the same position (e.g., GOES-16 & 19 are both GOES-East) under the hood; instead of key `goes_id`, it is now `goes_position`, and the options are either "east" or "west."
- Preblocks and postblocks are modified to adapt to this new 5-delimeter naming convention.
- modified `credit convert` first prompt message for entering `level_coord` name

Regarding Charlie's 1st comment, Charlie and I have discussed and concluded that it may be more appropriate to address it in a separate PR by introducing two Dataset classes, "StaticDataset" and "DynamicForcingDataset".

Minor to-dos:

- [ ] explanatory page in the docs/source. @djgagne are you referring to the `DataSets.md` and `prepare_new_dataset.md` pages? --> require some time
- [x] add another ERA5 dataset to the multi_source_data.yml to demonstrate the use case of having more than one call of the same dataset. --> quick work

config/gen_2/examples/multi_source_data.yaml
kevinyang-cky: Addressed!

charlie-becker (approved):
Thank you for addressing the comments! LGTM!

djgagne (approved):
LGTM

config/gen_2/examples/multi_source_data.yaml
```diff
@@ -23,8 +24,24 @@ data:
...
                     # single file — filename_time_format not needed
+        
+        ERA5_UpperAir:
+            dataset_type: "era5"
+            level_coord: "level"
+            levels: [10, 30, 40, 50, 60, 70, 80, 90]
+            variables:
+                prognostic:
+                    vars_3D: ['U','V','T','Q']
+                    vars_2D: ['SP','t2m', 'V500','U500','T500','Z500','Q500']
```
danineamati, line 35: Do we want to include `t2m` in the `vars_2D` for `ERA5_UpperAir`? Similarly, do we want `tsi` dynamic forcing in `ERA5_UpperAir` instead of `ERA5_LowerAir`?

credit/datasets/_utils.py
```diff
@@ -141,3 +142,43 @@ def _find_file(
...
+        ts.month,
+        ts.day,
+        ts.hour,
+        ts.minute,
+        ts.second,
+        calendar=calendar,
+    )
+
+
+def _init_fs(self):
```
danineamati, line 169: Kinda minor, but do we want to call it `_start_s3_fs()`? I think it may be confusing since (1) `__init___` is protected for classes and (2) `self` is usually just for classes. Note that we do not actually use `self` in this function, so the method is static. 

Super minor, but typo on `invocation whenm``mode`` is` should be `invocation when ``mode`` is`

credit/datasets/era5.py
```diff
@@ -384,103 +269,35 @@ def __init__(self, data_config: dict, return_target: bool = False) -> None:
...
-        self.datetimes: pd.DatetimeIndex = self._build_timestamps()
-
-        # file_dict maps field_type → sorted list of (start, end, path) intervals
-        self.var_dict: dict[str, dict[str, list[str]]] = {}
-
-        for field_type, d in source_cfg["variables"].items():
-            self._register_field(field_type, d)
-        # Initialize on first call to __getitem__
+        # Initialize the s3fs on the first call to _extract_field within __getitem__
         self.fs = None
```
danineamati, line 296: Per @charlie-becker 's comment, should this also be `self._fs`?

credit/datasets/goes.py
```diff
@@ -154,19 +151,14 @@ class GOESDataset(Dataset):
 
         data:
             source:
-                GOES:
-                    goes_id: "goes16"
+                Example_GOES:  # User-provided name (arbitrary key)
+                    dataset_type: "goes"
+                    goes_position: "east"  # or "west"
```
danineamati, line 156: We could also change HRRR Dataset to have a `product` field if we prefer this system to match GOES. Otherwise, we can put it on the back burner @charlie-becker

credit/datasets/goes.py
```diff
@@ -563,7 +461,7 @@ def _extract_field(
         Dispatches to ``_load_local_var`` or ``_load_remote_var`` depending on
         ``mode``, then stores each variable as a ``torch.Tensor`` of shape
         ``(1, 1, ny, nx)`` under the key
-        ``"{goes_id}/{field_type}/2d/{vname}"`` in ``sample``. Does nothing if
+        ``"{goes_position}/{field_type}/2d/{vname}"`` in ``sample``. Does nothing if
```
danineamati, line 464: I think this also would have the user-provided name first (i.e., 5 parts)?

PR #381, opened 2026-05-12
Quick fix on BaseDataset PR

kevinyang-cky:
Address additional comments in #376


danineamati:
Copying over the comments from before:
1. Do we want to include `t2m` in the `vars_2D` for `ERA5_UpperAir`? Similarly, do we want `tsi` dynamic forcing in `ERA5_UpperAir` instead of `ERA5_LowerAir`? https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/config/gen_2/examples/multi_source_data.yaml#L4-L41
2. Do we want to call it `_start_s3_fs()`? I think it may be confusing since (1) `__init___` is protected for classes and (2) `self` is usually just for classes. Note that we do not actually use `self` in this function, so the method is static. Super minor, but typo on `invocation whenm``mode`` is` should be `invocation when ``mode`` is`. https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/credit/datasets/_utils.py#L169-L176
3. Per @charlie-becker 's comment, this might need to be `self._fs`. We may consider a search for `self.fs` across the repo if needed. https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/credit/datasets/era5.py#L295-L296
4. For consistency, we may want to make the HRRR dataset better match the GOES dataset in terms of handling different products. https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/credit/datasets/goes.py#L154-L159 and https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/credit/datasets/hrrr.py#L715-L719
5. I think this also would have the user-provided name first (i.e., 5 parts). https://github.com/NCAR/miles-credit/blob/c471819a29ddc4c6ba6a383a70c9c523409a58f4/credit/datasets/goes.py#L461-L465

kevinyang-cky:
@danineamati, I can address 1, 2, 3, and 5 quickly, and maybe 4 for you? Once they are all addressed, we can tag DJ and Charlie for review.

kevinyang-cky:
@danineamati For 1), I am keeping t2m in the ERA5_LowerAir block and moving the tsi dynamic forcing to the ERA5_UpperAir block (let me know if that aligns with what you had in mind or not). 2, 3, and 5 are addressed as you suggested.

credit/datasets/hrrr.py
```diff
@@ -5,16 +5,16 @@
 
 Supports three HRRR products (``VALID_PRODUCTS``):
 
-* ``wrfprsf`` — pressure-level output (default, ~200 MB/file)
```
djgagne, line 8: The HRRRR product types are wrfprs, wrfnat, wrfsfc, and wrfsubh. The f indicates the forecast time. Please update the nomenclature throughout. wrfsfc has all the single level diagnostic variables that are also in wrfprs but with much fewer pressure levels.

djgagne (changes requested):
I think the errors with torch harmonics might be related the whole module being imported and some underlying errors cropping up with that. When nvidia released the recent upgrade to torch-harmonics, they only published linux builds of the package. The easiest solution to these errors for now is to pin `torch-harmonics<0.9` in requirements.txt and pyproject.toml. Longer term, we can edit the skebs.py file to not import the whole torch harmonics package and only the parts that are being used, which should get around importing the files that have busted kernels.

credit/datasets/hrrr.py
danineamati: Makes sense! I updated the code, and the links now have `{product}f{FF:02d}.grib2`.

djgagne (approved):
Looks good to me.

PR #389, opened 2026-05-15
Nuke gen2 flat-key injection: derive variables from preblock channel_map

jsschreck:
## Summary

- Removes `inject_flat_var_keys()` shim from `train_gen2.py`; replaces with a one-batch preblock probe that reads `metadata["target"]["_channel_map"]` from `ConcatToTensor` to populate `conf["data"]["variables"]`, `conf["data"]["surface_variables"]`, and `conf["data"]["diagnostic_variables"]` before loss and metrics are initialized
- Adds `extract_flat_keys_from_channel_map()` to `credit/trainers/utils.py` — the shared helper used by both trainer and rollout probes
- Removes `_inject_flat_schema()` from both rollout scripts (`rollout_to_netcdf_gen2.py`, `rollout_realtime_gen2.py`); replaces with `_set_output_conf()` using the same channel_map probe pattern. `level_ids` and `scaler_type` still read from source config (no channel_map equivalent)
- Rewrites `_inject_tracer_inds()` in both rollout scripts to use the flat keys set by `_set_output_conf()` instead of re-parsing the nested source config. Moves dataset construction before model loading so the probe runs first

## Unrelated smoke config fix

`smoke_gen2_casper.yml` and `smoke_gen2_multistep_casper.yml` had stale `dataset_type: "era5"` — this was left broken by d637004 ("Convert ERA5 to Local"), which renamed the registry key to `"local"`. We discovered it while running the smoke test to validate this PR and fixed it here. It is otherwise unrelated to the flat-key work.

## Note for @charlie-becker

`_inject_tracer_inds()` in both rollout scripts is still a shim — it computes channel indices for `TracerFixer` from the flat keys since the postblock doesn't derive them itself yet. Once the postblock config work is completed, `_inject_tracer_inds()` can be removed entirely. Each function has a `# TODO` marking the spot.

## Scope

V1 pipeline is unaffected. `metrics.py`, `losses/weighted_loss.py`, and `trainerERA5gen2.py` are untouched.

## Test plan

- [x] Smoke test `smoke_gen2_casper.yml` ran to completion on Casper V100 (job 3705386): 5 train batches + 2 validation batches, checkpoint saved
- [x] Log confirms correct extraction: `Configured from preblock metadata: 4 3D vars, 7 surface vars, 0 diagnostic vars`
- [x] Channel count matches previous runs: `C_in=74`, GPU memory footprint identical

### DIFF
--- a/credit/applications/rollout_realtime_gen2.py
+++ b/credit/applications/rollout_realtime_gen2.py
@@ -47,6 +47,7 @@ from credit.models.checkpoint import load_model_state, load_state_dict_error_han
 from credit.output import load_metadata, make_xarray, save_netcdf_increment
 from credit.postblock.gen1 import GlobalMassFixer, GlobalWaterFixer, GlobalEnergyFixer
 from credit.nwp import build_GFS_init
+from credit.trainers.utils import extract_flat_keys_from_channel_map
 
 logger = logging.getLogger(__name__)
 warnings.filterwarnings("ignore")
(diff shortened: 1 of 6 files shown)


credit/applications/rollout_realtime_gen2.py
```diff
@@ -60,34 +61,33 @@
...
 
-def _inject_flat_schema(conf):
-    """Inject v1-style flat keys into conf['data'] so output.py utilities work."""
-    if "variables" in conf["data"]:
-        return
+def _set_output_conf(conf, preblocks, dataset, init_time):
+    """Populate output conf keys from a preblock channel_map probe."""
     src = conf["data"]["source"]["ERA5"]
-    v = src["variables"]
-    prog = v.get("prognostic") or {}
```
charlie-becker, line 69: Many of these functions are duplicated (defined at least twice) in `rollout_realtime_gen2.py` and `rollout_to_netcdf_gen2.py`, They should be in a shared utilities module.

charlie-becker (changes requested):
I appreciate the functionality to move towards using the metadata from the new data schema. However, as you alluded to; this is a pretty intermediate fix / change that still uses functions that create a bridge from Gen2 to Gen1 for downstream tasks in output and metrics. This is likely fine as we primarily need to solidify the post-blocks before moving on to further downstream tasks. With that said, I'm probably okay moving forward with this temporary bridge (which does lean a little more on Gen2 schema).

I have one comment regarding many duplicate functions (at least function names, I haven't verified they are identical). 

I also have one other comment regarding hardcoding "era5" in a couple of places. For example:

```
def _sample_to_batch(sample):
    """Wrap a single LocalDataset sample (no batch dim) into preblock-compatible dict."""
    return {"era5": {"input": {k: v.unsqueeze(0) for k, v in sample["input"].items()}, "metadata": sample["metadata"]}}
```

I am unsure if this is needed / hardcoded in some downstream modules (output / metrics). If it is necessary, I'm probably okay with it until we get to refactoring those. But if it's not necessary,  we should change it (It would also break training in this case, (which I don't know if we actively test for).

jsschreck:
Thanks for your comments @charlie-becker ... you are right that its not 100% gen2 yet -- that requires postblocks to be complete, including whats going on in output.py. I think you guys should prioritize getting that done before anything else. I can update these PRs accordingly once that is finalized.

credit/applications/rollout_realtime_gen2.py
jsschreck: @charlie-becker  These functions will get removed entirely once output.py is integrated into post-blocks. Ill make sure we have just one bridge working until output.py is done

jsschreck:
Superseded by Charlie's postblocks PR which handles flat-key derivation from preblock channel_map.

PR #390, opened 2026-05-15
credit convert: v1-to-v2 standard scaler migration to BridgeScaler

jsschreck:
## Summary

- Adds `credit convert` CLI command to auto-convert v1 z-score NC scaler files to v2 BridgeScaler JSON format
- Adds `--level-coord` flag and interactive prompts for level coordinate detection
- Fixes `postblock/scaler.py`: switches `load_scaler` to `load_scaler_dict` to match the bridgescaler dict API

## Files changed

- `credit/cli/_convert.py` — main conversion logic
- `credit/cli/_parser.py` — `--level-coord` flag
- `credit/postblock/scaler.py` — `load_scaler` → `load_scaler_dict` fix

## Test plan

- [x] Run `credit convert` on an existing v1 z-score NC file and verify JSON output
- [x] Smoke test v2 config with converted scaler


kevinyang-cky:
@jsschreck, I’d be glad to assist with testing the two items in the plan if an independent validation is needed.

kevinyang-cky (commented):
I ran `credit convert`, and the conversion from the existing v1 z-score NetCDF4 file to the v2 Bridgescaler JSON format produces content that is consistent across both files.

[‎config/gen_2/examples/example-v2026.2.yml](https://github.com/NCAR/miles-credit/pull/390/changes#diff-590fa154b8902f9220983f273b5a4ba2b78b868622beb988bcb88335072d735c) also ran fine on my end.

One minor question I have @jsschreck: it appears that `*_post.json `is a subset of `*_pre.json`. If so, is there a specific reason to retain both files?

An additional note: Bridgescaler emits a warning (see below) when an input tensor does not include a `variable_names` attribute. In this case, the warning is not particularly useful for CREDIT Gen2, since variable names are already implicitly defined by the keys in the nested data dictionary, making the stated concern about variable alignment less relevant. It may be worth suppressing this warning either within Bridgescaler (when using `scale_var_dict`) or in CREDIT’s train.py to reduce unnecessary log noise. This doesn't have to be addressed now and can be dealt with in a later PR.
```
/glade/work/kevinyang/conda-envs/credit/lib/python3.12/site-packages/bridgescaler/distributed_tensor.py:64: UserWarning: Input data lacks variable names. When performing fit or transform, data/scaler consistency check is limited to variable counts; order cannot be validated. Ensure variable alignment to prevent incorrect results.
  warnings.warn(
```

jsschreck:
@kevinyang-cky Good catch. `*_post.json` is structurally a subset of `*_pre.json` ... they hold the same scalers, but the pre JSON has `{"input": {...}, "target": {...}}` nesting while the post JSON is a flat `{source: {key: scaler}}`
  dict matching what the postblock reader expects. We could embed the post content inside the pre JSON under an `"output"` key and read from there, but that would couple the postblock config format to the pre JSON
  structure. Keeping them separate means a user can swap out just the postblock scaler (e.g., for a finetuned one) without touching preblock config. If you or DJ see a cleaner unification, I will update accordingly.

@djgagne when using `scale_var_dict`, BridgeScaler emits a warning that variable alignment can only be checked by count, not order. For CREDIT Gen2 this is a non-issue since variable identity is encoded in the dict keys, not tensor position. Two options: suppress it inside BridgeScaler when `scale_var_dict` is used, or we suppress it on the CREDIT side. Do you want me to handle it now in CREDIT, or should we punt to a later PR?

kevinyang-cky:
@jsschreck I am fine with just keeping them separately. For the warning message, if we decide to suppress it inside Bridgescaler, I can make that change since I wrote  `scale_var_dict`.

jsschreck:
@kevinyang-cky feel free to merge this PR then if you guys are happy.

kevinyang-cky (approved):
Merging now!

Issue #388, opened 2026-05-14
Fix grid.py


charlie-becker:
Current `grid.py` has a bug that does not not properly wrap / stitch when creating the SCRIP files for some global grids for conservative regridding.

@djgagne I will make the change to `grid.py` that will support rectilinear and curvilinear SCRIP files for both bilinear and conservative methods. However, I was wondering if we'd like to also include the call to `ESMF_RegridWeightGen` in the same script (and deleting the intermediate SCRIP) files to streamline the process for users. Thoughts?

Issue #380, opened 2026-05-12
Need to handle varying temporal resolutions


charlie-becker:
Currently, `MultiSourceDataset` takes the intersection of time stamps that are "available" in each of the individual datasets to ensure that each dataset has data for that time, using a global date range from the config. However, there are many use cases where we will have varying resolution and want to handle these situations in a variety of ways. 

The simplest, and likely easiest to implement, would be to persist data at a coarser resolution. We can still use a global date range (which should be set to the finest preferred temporal resolution, and then specify an interval in the config for each dataset with coarser resolution. 

We should be mindful to cache persisted data and not reload.

PR #391, opened 2026-05-15
Cbecker update

### LINKED ISSUE #359: Supporting different temporal resolutions in Multi-Source
Currently, `multi_souce.py` uses a timestamp intersection across all time steps for each model and discards any samples that don't have a data sample there. However, there are many use cases where we want to handle different temporal resolutions. Some use cases:

- Persisting data every `x` time steps
- Potentially interpolating / extrapolating data
- ?

### LINKED ISSUE #392: `dataset_type: "era5"` in example v2026.2 yaml not in `_SOURCE_REGISTRY` on main
## Summary

The v2 example config `config/gen_2/examples/example-v2026.2.yml` uses `dataset_type: "era5"`, but the `_SOURCE_REGISTRY` on main does not include an `"ERA5"` key.

## Details

The example yaml specifies:

```yaml
data:
  source:
    ERA5:
      dataset_type: "era5"
```

`applications/train_gen2.py` resolves this through:

```
load_dataset()                                    # credit/trainers/utils.py
  -> MultiSourceDataset(data_conf, ...)           # credit/datasets/multi_source.py
       -> route_to_dataset_class(source_cfg)
            -> _SOURCE_REGISTRY[dataset_type.upper()]
```

On a CWA HPC version I'm using, `_SOURCE_REGISTRY` contains an `"ERA5"` key mapping to `credit.datasets.era5.ERA5Dataset`, and the example runs fine. But in the current main snapshot the registry only has `"LOCAL"` (mapping to `credit.datasets.local.LocalDataset`) and there is no `ERA5Dataset` class in `credit/datasets/era5.py` — only `ARCOERA5Dataset`.

This looks like a `LOCAL` → `ERA5` rename (and corresponding move of `LocalDataset` to `ERA5Dataset` under `credit/datasets/era5.py`) that has not fully propagated to main.

charlie-becker:
### Description
Three main changes:

- Adds a persistence feature to the `BaseDataset` class which allows a user to "persist" data from a different source, particularly when it has a different temporal resolution. The general procedure to implement this is to use the `conf["data"]` global start / end / dt times at the **finest** resolution. Then under a `conf["data"]["source"]` a user can add a `temporal_mode: "persist"` key along with a `time_step:` key defining a coarser resolution for that source. If a datafile in the "global" timestep is not found, it will cache the previous timestep and use that. Future expansions will just need a different `temporal_mode:` key.
- Refactor and fix a bug in `grid.py` to generate SCRIP file formats for `ESMF_RegridWeightGen` to be used in the regridding preblock. 
- Fix a small time bug in `LocalDataset`

Closes #359, #380 (duplicate), #388

### Checklist

### DIFF
--- a/credit/datasets/base_dataset.py
+++ b/credit/datasets/base_dataset.py
@@ -225,6 +225,11 @@ class BaseDataset(AbstractBaseDataset):
         if "mode" in self.curr_source_cfg:
             self.mode = self.curr_source_cfg["mode"]
 
+        # temporal_mode: None (default, exact timestamp match required) or
+        # "persist" (snap to last native timestamp via asof(); used for coarser sources)
+        self.temporal_mode: str | None = self.curr_source_cfg.get("temporal_mode")
+        self._persist_cache: dict = {}
+
         # Placeholder for static metadata
         self.static_metadata: dict[str, Any] = {}
 
@@ -245,6 +250,12 @@ class BaseDataset(AbstractBaseDataset):
     def __getitem__(self, args: tuple[pd.Timestamp, int]) -> dict[str, Any]:
         """Return a nested input/target sample dict.
 
+        When ``temporal_mode == "persist"``, the requested timestamp *t* is
+        snapped to the last native timestamp at-or-before *t* via
+        ``pd.DatetimeIndex.asof()``.  The result is cached so that multiple
+        fine-resolution master-clock ticks within the same native interval
+        only trigger a single file read.
+
         Args:
             args: ``(t, i)`` where *t* is the current timestamp (nanoseconds
                 or pd.Timestamp) and *i* is the within-sequence step index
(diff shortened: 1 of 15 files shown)


danineamati:
Would you be okay with us also changing the `dataset_type` to `"local"` and change the path to include `$USER`?

https://github.com/NCAR/miles-credit/blob/8abb19521f091308518953b0abd138741e328e99/config/gen_2/smoke/smoke_gen2_casper.yml#L6-L13

I can make the change if easier.

charlie-becker:
> Would you be okay with us also changing the `dataset_type` to `"local"` and change the path to include `$USER`?
> 
> https://github.com/NCAR/miles-credit/blob/8abb19521f091308518953b0abd138741e328e99/config/gen_2/smoke/smoke_gen2_casper.yml#L6-L13
> 
> I can make the change if easier.

Yep! Go ahead and make the changes!

kevinyang-cky:
Minor thing, but following @danineamati suggestion: https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/gen_2/examples/example-v2026.2.yml#L23-L25

https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/gen_2/examples/wxformer_era5_025deg_6hr.yml#L15-L17

https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/gen_2/examples/wxformer_npj_era5_028deg.yml#L13-L15

https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/example-v2026.2.yml#L31-L33

These configs will also need to be updated. Thanks!

danineamati:
If I understand correctly, I think we may also have changed this naming convention a couple of weeks ago?

https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/credit/trainers/trainerERA5gen2.py#L104-L107

vs 

https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/gen_2/examples/wxformer_era5_025deg_6hr.yml#L56-L61

The example gen 2 configs have `data_valid`, and the smoke test configs have `validation_data`. As in, all good if I change these to `validation_data` or is this intentional?

charlie-becker:
> If I understand correctly, I think we may also have changed this naming convention a couple of weeks ago?
> 
> https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/credit/trainers/trainerERA5gen2.py#L104-L107
> 
> vs
> 
> https://github.com/NCAR/miles-credit/blob/bdcd57aa9d2884d3c1b4a95807615057aea3f918/config/gen_2/examples/wxformer_era5_025deg_6hr.yml#L56-L61
> 
> The example gen 2 configs have `data_valid`, and the smoke test configs have `validation_data`. As in, all good if I change these to `validation_data` or is this intentional?

Yep, you're correct. Change welcome

credit/datasets/base_dataset.py
```diff
@@ -225,6 +225,11 @@ def __init__(self, data_config: dict[str, Any], return_target: bool = False) ->
         if "mode" in self.curr_source_cfg:
             self.mode = self.curr_source_cfg["mode"]
 
+        # temporal_mode: None (default, exact timestamp match required) or
+        # "persist" (snap to last native timestamp via asof(); used for coarser sources)
+        self.temporal_mode: str | None = self.curr_source_cfg.get("temporal_mode")
```
danineamati: If I am understanding correctly, if the temporal mode is not `None` or `"persist"`, it will silently fall back to `None`. I am wondering if any of the following would be appealing:
1. Type hint `self.temporal_mode: Literal["exact", "persist"] = self.curr_source_cfg.get("temporal_mode", "exact")`
2. Raising a `ValueError` or logging error for `self.temporal_mode not in [None, "persist"]`
3. Include a logging error in `__getitem__` below.

charlie-becker: It does provide a warning if it is not `temporal_mode` is not specified and it can't find the timestep, but it keeps running and would likely error in the dataset itself. It probably makes sense to default to "exact" as you suggest.

charlie-becker: Look in `multi_source.py` for the current logging warning.

danineamati: I see. Since we expect to use Multisource as the standard in the CLI/API, we will get the warning below. But it doesn't have the warning up front, so to speak.

https://github.com/NCAR/miles-credit/blob/f7f97cb6e67fb97a27b76cab117e75a199a4d0bd/credit/datasets/multi_source.py#L224-L233

I can go ahead and make "exact" the default and remove the None case.

danineamati (approved):

djgagne (approved):
Fixes look good. Glad to have the persist option for time matching now.

PR #394, opened 2026-05-16
fix(EMATracker): exclude spectral norm buffers from EMA shadow

jsschreck:
## Bug

`use_spectral_norm: True` + `use_ema: True` causes validation loss to explode while training loss stays healthy. Reported by @djgagne (Michael Yang) on a 10-year CrossFormer run: train_loss → 0.018 by epoch 4, valid_loss → 12262 and growing every epoch. `y_pred` in eval mode reaches ±350k vs ±47 for targets.

## Root cause

`nn.utils.spectral_norm` stores `weight_u` and `weight_v` (power-iteration unit vectors) as **buffers** inside `state_dict()`. `EMATracker.__init__` copied all of `state_dict()` into `self.shadow` without filtering these out.

EMA-averaging unit vectors reduces their norm on every update:

```
||EMA(u_t)|| = decay * ||u_{t-1}|| + (1-decay) * ||u_t|| < 1
```

In eval mode, `F.spectral_norm` computes `sigma = u^T W v` using the EMA-averaged (sub-unit-norm) vectors, underestimating sigma by ~6% per layer. CrossFormer applies spectral norm to every `Conv2d / Linear / ConvTranspose2d` (~175 layers via `apply_spectral_norm(self)`). The cascade: `1.064^175 ≈ 50,000x` amplification — consistent with the reported ±350k predictions.

DDP makes this worse: `weight_u` / `weight_v` are buffers, **not** parameters, so they are **not** synced across ranks. Each rank's EMA shadow diverges independently.

## Fix

New `_is_spectral_norm_buffer(key, state)` detects `weight_u` / `weight_v` by checking for a co-located `weight_orig` key (the renamed parameter that spectral norm creates). These keys are excluded from `self.shadow` at init and silently skipped in `update()`. `swap()` requires no change — it only copies keys present in `self.shadow`, so the model retains its current well-converged training u/v after each EMA swap.

## Test

- SN buffer exclusion unit test: `Any SN keys in shadow? False` — PASS
- 4000-step DDP perturbation + EMA test: fixed code stable (output max 72.1), buggy code diverges — PASS

cc @djgagne — please also forward to Michael Yang (not on repo) who reported this on the v2026.2 run.

### DIFF
--- a/credit/trainers/base_trainer.py
+++ b/credit/trainers/base_trainer.py
@@ -70,7 +70,19 @@ class EMATracker:
         self.shadow: OrderedDict = OrderedDict()
         state = model.module.state_dict() if hasattr(model, "module") else model.state_dict()
         for k, v in state.items():
+            if self._is_spectral_norm_buffer(k, state):
+                continue  # power-iteration vectors must not be EMA-averaged
             self.shadow[k] = v.detach().float().cpu().clone()
+
+    @staticmethod
+    def _is_spectral_norm_buffer(key: str, state: dict) -> bool:
+        # nn.utils.spectral_norm stores weight_u / weight_v alongside weight_orig.
+        # Averaging these vectors destroys the power-iteration convergence and causes
+        # sigma = u^T W v → 0 in eval mode → weight explodes.
+        if key.endswith("_u") or key.endswith("_v"):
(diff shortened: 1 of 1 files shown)


djgagne (approved):
LGTM

PR #393, opened 2026-05-15
Add new postblock scheme into Trainer

charlie-becker:
### Description
This integrates the new postblock and data schema into `trainerERA5Gen2.py`. 

The new full data schema is a super dictionary that contains raw data from the loader, preprocessed data by the preblocks, and the model predictions, which is then fed to the postblocks. 

I have lightly tested it with using the reconstruct postblock, successfully. For now, I converted the original trainer into a legacy trainer module and implemented an equivalency test. We can remove the legacy trainer and this test once this is more throughly tested and approved. 

Once we settle on the details of how this operates, it should be relatively straight forward to implement into the other trainers and rollout scripts. 

I also had an LLM do a conversion of the `gen1.py` postblocks into their own modules to fit the new postblock scheme, but they are NOT ready yet and still need significant work. Once I work out a couple more details (and clean the conservation postblocks) I will convert this over from draft to a regular PR. That said, I still think it would be useful to get review feedback on this.  

Closes #374

### Checklist
- [ ] I am familiar with the [contributing guidelines](https://miles-credit.readthedocs.io/en/latest/contrib.html).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date.

PR #398, opened 2026-05-22
Postblock integration

### LINKED ISSUE #374: Integrate Gen2 Postblocks
`apply_postblocks()` now takes in a single dictionary that includes any combination of "input" / "target" / "prediction" / "meta" keys which will allow postblocks to utilize the same data schema we use for preblocks, and also includes the input data for diagnostics.

There are two next steps to implement this:

1. `apply_postblocks()` is not currently used in training and rollout scripts, so we need to utilize the postblocks through this and manually add in the keys of the data for input into the post-block scheme inside the training and rollout scripts.
2. Update configs to use postblocks. Importantly, there is an additional "key" argument needed for postblocks to know which "set" of data to operate on.

charlie-becker:
### Description
   
This PR introduces a unified state container, `full_data_dict`, as the single source of truth for all data flowing through the Gen2 autoregressive training loop. Rather than managing raw inputs, preprocessed tensors, model outputs, and postprocessed results as separate local variables, everything accumulates into one dict with well-defined keys (`ic_raw`, `x_raw`, `y_raw`, `ic_preprocessed`, `x`, `y`, `metadata`, `y_pred`, `y_processed`). This makes the rollout loop legible at any point in the loop and you can inspect the full pipeline state in one place.
                                                                                                                                                                                                                                                                                                                      
Preblocks are now structured into two named phases — `ic_only` and `per_step` — enforced by `build_preblocks()`. The `ic_only` chain runs once at `t=0` (handling things like static regridding) and its output is stored as `full_data_dict["ic_preprocessed"]`. The per_step chain runs every rollout step and terminates with `ConcatToTensor`, which collapses the nested `{source: {var_key: tensor}}` variable dict into a flat (B, C, H, W) tensor sorted deterministically by field type (prognostic → static → dynamic_forcing → diagnostic). `ConcatToTensor` also emits a `_channel_map` into `metadata["input"]` and `metadata["target"]`, so downstream code knows which channel indices correspond to which variables.                                                                                                                                                                                                                                      
                                                            
`apply_preblocks()` has a single return contract: when `ConcatToTensor` is in the chain, it returns `{"x": tensor, "y": tensor, "metadata": dict}`. The trainer merges this directly into `full_data_dict` with `.update()`, so `full_data_dict["x"]`, `full_data_dict["y"]`, and `full_data_dict["metadata"]` are always ready for the model forward pass.
                                                                                                                                                                                                                                                                                                                   After each model forward pass, `full_data_dict["y_pred"]` holds a flat tensor. The `Reconstruct` postblock reads `y_pred` alongside `metadata["target"]["_channel_map"]` and writes `full_data_dict["y_processed"]` as a nested `{source: {var_key: tensor}}` dict — mirroring the format the dataset originally produced. 
                                                                                                                                                                                                                                                                                                                   
Postblocks are similarly split into `per_step` (runs inside the rollout loop, e.g. `Reconstruct`) and `post_rollout` (runs once after the full loop).                                                                                                     
                                                               
At each step `t>1`, `assemble_rollout_batch()` constructs the next input by routing variables by field type extracted from the `source/field_type/dim/varname` key format: prognostic and diagnostic variables come from `full_data_dict["y_processed"]` (the model's last prediction), dynamic forcing comes from the current dataloader batch, and static variables always come from `full_data_dict["ic_preprocessed"]` (computed once at t=0 and never overwritten). 
                                                                                                                                                                                                                                                                                                                   
Many tests were adjusted to account for the restructuring and small additions, including more coverage for the data dict schema.

Additionally, `trainerERA5gen2.py` was generically renamed `trainer_gen2.py`. The other gen2 trainers were not touched and we need to eventually decide if they can be brought into the generic trainer or need their own. They would still need to be modified for this data structure and renamed generically (i.e. `trainer_ensemble.py`).

### Related Issues and PRs

Closes #374 #393

### Checklist


djgagne (approved):
LGTM

PR #403, opened 2026-05-27
Add PyTorch Solar Dataset: on-the-fly top-of-atmosphere solar radiation computation

### LINKED ISSUE #352: Separate Total Solar Irradiance (TSI) to own Dataset
Currently, dynamic forcings that are not a part of the specific "dataset" itself, is not supported. For example, pre-calculated TSI does not have it's own dataset but it supported as `dynamic_forcing:` within `ERA5DataSet()`. 

We should separate out dynamic forcings (TSI specifically for now) into it's own `torch.Dataset`.

kevinyang-cky:
#### Summary

- Implements TISRDataset that computes ERA5-compatible top-of-atmosphere Total Incident Solar Radiation (TISR) on-the-fly without loading files, intended for use as a dynamic forcing input.
- Implements the solar radiation pipeline in PyTorch, referencing [GraphCast's solar_radiation.py](https://github.com/google-deepmind/graphcast/blob/08cf73625c9d12bd9aaa038868bcb2fe488f2a22/graphcast/solar_radiation.py): TSI interpolation, orbital parameter calculation (declination, equation of time, Earth-Sun distance), cosine zenith angle, and trapezoidal integration over the accumulation window.
- Handles both curvilinear and rectangular lat/lon grids loaded from NetCDF via xarray.

#### Tests

`tests/test_tisr.py` is included and contains two parts: unit tests for each pipeline component, and parity tests against GraphCast that are automatically skipped when the package is not installed in the environment.

Additionally, the figure below shows validation results against ARCO ERA5 hourly data at 00:00 UTC for the following timestamps (in chronological order): 1952-03-01, 1952-06-01, 1952-09-01, 1952-12-01, 2021-03-01, 2021-06-01, 2021-09-01, 2021-12-01. The largest absolute percentage differences occur where the reference TISR values are relatively small.

arcoera5_tisr_comparison

#### Review guideline

@djgagne @danineamati — feel free to dive into the implementation if helpful (it’s ~1.7k LOC, with about half in tests), but I’d especially appreciate feedback on the new solar data config block to ensure it covers all relevant use cases. Edits to goes.py and mrms.py are just docstring updates.

### Related Issues and PRs

Closes #352 

### Checklist
- [x] The documentation is (mostly) up to date --> Docstrings and inline comments are included, but not the documentation page.


credit/datasets/tisr.py
```diff
@@ -0,0 +1,843 @@
...
+
+    Returns:
+        tuple[torch.Tensor, torch.Tensor]:
+            - **times** – 1-D tensor of fractional years, one entry per year
+              from 1951.5 to 2034.5 (mid-year sampling).
+            - **tsi_values** – 1-D tensor of TSI values (W m⁻²) corresponding
+              to each entry in *times*.
+    """
+    # Mid-year fractional years from 1951 through 2034 (one value per year)
+    times = torch.arange(1951.5, 2035.5, 1.0, dtype=_TORCH_DTYPE)
```
danineamati: There are several places, including here, where the device is not set. I believe this means it will always run on the CPU. I think we may want to include the device as a function argument to allow the user to compute the TSI on the GPU.

Would you mind checking with `nvtop` if the GPU spins up when you run the TSI calculations?

credit/datasets/tisr.py
```diff
@@ -0,0 +1,843 @@
...
+
+        Args:
+            field_type: ``"dynamic_forcing"`` only, and others set to null.
+            t: Timestamp for which to load data.
+            sample: Output dictionary that is updated in-place.
+        """
+        # Check if the var_dict exists for the field type
+        vd = self.var_dict.get(field_type)
+        if not vd:
+            return
```
danineamati, line 976: Should we include a logging warning?

tests/test_tisr.py
```diff
@@ -0,0 +1,881 @@
...
+            "num_integration_steps": 360,
+            "latlon_grid_path": rectangular_nc,
+        }
+        ds.dt = pd.Timedelta(hours=1)
+        ds.var_dict = {"dynamic_forcing": {"vars_2D": ["tisr"]}}
+        ds.init_register_all_fields = MagicMock()
+        ds._get_field_name = MagicMock(return_value="src/dynamic_forcing/2d/tisr")
+        # Run the real __init__ with a dummy data_config (BaseDataset.__init__ is mocked)
+        TISRDataset.__init__(ds, data_config={}, return_target=False)
+    return ds
```
danineamati: I am not sure we want to use Mock in this way. Why not pull from BaseDataset?

credit/datasets/tisr.py
kevinyang-cky: Thanks for flagging this. Indeed, the current module has everything computed on the CPU. Device management is now implemented through the module, allowing users to specify compute on either CPU/GPU. Note that when `Dataloader` uses `num_worker > 0`, CUDA tensors cannot be created in forked workers, so in the CREDIT pipeline, this Dataset needs to live on CPU.

credit/datasets/tisr.py
kevinyang-cky: Added!

tests/test_tisr.py
kevinyang-cky: Mock removed!

kevinyang-cky:
Addressed @danineamati's review comments. Also added the ability to specify the lat/lon range and the number of grid points per axis in the config (lat_spec/lon_spec as [start, end, num_points]), so the lat/lon grid can be built in memory instead of always being read from a grid file. Exactly one source — latlon_grid_path or the specs — must be provided.
Also renamed `_load_latlon_grid` → `_get_latlon_grid` (it now both reads from file and builds from spec), and moved grid construction into `TISRDataset.__init__ `so it's built once and cached rather than re-read on every sample. Updated tests accordingly.

danineamati (approved):
Sounds good!

djgagne (approved):
It looks like @danineamati caught issues that have been resolved, and all tests are passing, so merging now.

PR #407, opened 2026-05-28
Updates to CREDIT parallelism following pytorch's latest schema

jsschreck:
## Summary

This PR updates CREDIT's parallelism infrastructure to follow PyTorch's current distributed API, and introduces a unified `parallelism:` config block for the gen2 trainer.

**New `parallelism:` config block** replaces the legacy `trainer.mode` key for gen2 configs. Three sub-keys control independent axes:
- `data: ddp | fsdp2 | none` — data parallelism strategy
- `tensor: N` — tensor-parallel degree (splits attention heads across N GPUs; WXFormer only)
- `domain: N` — spatial domain sharding across N GPUs (splits the H dimension)

The legacy `mode:` key still works for gen1 configs and is ignored when a `parallelism:` block is present.

**`distributed_model_wrapper_gen2`** — new wrapper that applies parallelism in the correct order: domain → tensor → FSDP2/DDP. The old `distributed_model_wrapper` is unchanged and still used by gen1 trainers.

**FSDP2** replaces the old FSDP1 path for gen2. Uses `fully_shard` from `torch.distributed.fsdp` with DCP-based checkpointing. `MixedPrecisionPolicy` handles bf16; manual `autocast` is skipped in FSDP2 mode to avoid conflicts with spectral norm power-iteration buffers.

**Domain parallelism** pre-pads the full input tensor before sharding so model-internal padding is skipped (`_skip_internal_padding`). Gradients are synchronized across domain ranks after the backward pass.

**Tensor parallelism** uses `_tp_col`/`_tp_row` opt-in markers on model layers. `FeedForwardSwiGLU` hidden dim is aligned to a multiple of 8 for TP compatibility.

**Grad accumulation** is now wired into the gen2 training loop via `grad_accum_every`. Loss is divided by `grad_accum_every` before `.backward()`; the optimizer step fires at accumulation boundaries.

**Loss and metrics** updates support distributed domain/TP setups — latitude-weighted loss and metrics correctly handle sharded spatial dimensions.

**Benchmark script** (`applications/benchmark_parallelism.py`) measures step time, peak GPU memory, and throughput using synthetic data across all parallelism combinations.

## Tests

New test files cover FSDP2 parallel infrastructure (`test_fsdp2_parallel.py`), FSDP wrap/checkpoint policies (`test_fsdp_policies.py`), and distributed loss (`test_loss.py`). All run CPU-only for CI.

## Update 2026-06-09: bugs found and fixed in review

After merging main into this branch, we re-ran the full smoke matrix and did a close audit of the parallelism code. Several real bugs turned up; all are fixed here.

**1. Dataloader sharded by global rank (correctness, the big one).** The sampler was built with the global rank/world_size. With `tensor > 1` or `domain > 1`, ranks in the same TP/domain group received *different* samples: the TP row-parallel all_reduce summed partial outputs of different inputs, the domain halo exchange passed boundary rows from the wrong sample, and replicated (non-TP) parameters silently drifted apart across TP ranks. Training "worked" (loss decreased), so smoke tests never caught it. Fix: new `data_parallel_coords()` in `credit/parallel/mesh.py` computes `(dp_rank, dp_size)` from the mesh layout (row-major over dp/tp/domain, domain innermost); `train_gen2.py` shards data and seeds RNG by the dp coordinate. The full contract is documented in `Training.md` ("Data sharding and rank layout") and in the `data_parallel_coords` docstring. Anyone writing a new entry point against the `parallelism:` block must use this — global rank is only correct when `tensor: 1, domain: 1`.

**2. Per-rank sampler seed (correctness, affects every multi-GPU run).** `load_dataloader` seeded the sampler with `seed + rank`. `DistributedSampler` requires the *same* seed on every rank (each rank slices one shared permutation); per-rank seeds made each rank slice a different permutation, silently duplicating and dropping samples even for plain DDP/FSDP2. Fix: identical seed everywhere; the trainer now calls `sampler.set_epoch(epoch)` for per-epoch shuffling (it previously never did).

**3. Row-parallel TP bias trained at tp× the learning rate.** The bias was baked into the per-rank layer (full on tp rank 0, zeroed elsewhere), but every copy receives the same gradient, so the summed effective bias moved tp× as far per step. Fix: single replicated bias added after the all_reduce. Regression-tested with serial tp=2 parity tests.

**4. `grad_max_norm: "dynamic"` mixed norms and squared norms** in its cross-rank reduction. Fixed to sum squared norms before the sqrt.

**5. Smaller items:** 1×1 convs with non-default stride/padding/groups are now rejected by the TP wrappers instead of silently dropped; the "gradients will NOT sync" warning now covers all `data: none, dp > 1` cases; domain gradient sync is bucketed (one all_reduce per dtype instead of per parameter); removed per-timestep barriers in the train/valid loops; `v2` identifiers renamed to `gen2` (`distributed_model_wrapper_gen2`, `_has_gen2_parallelism`).

**Smoke configs moved to `tests/manual/gen2_parallelism/`** (new `tests/manual` tree for HPC-only tests, excluded from CI via `norecursedirs`). Configs were also updated to the `%Y` path-template convention from the recursive-file-lookup change merged from main — the old bare-glob style no longer parses, which the smoke run caught.

**Validation:** 58/58 CPU unit tests pass (incl. 9 new regression tests for the above), and all 8 single-node smoke modes pass on a Derecho gpudev node (ddp, fsdp2, fsdp2+AC, domain, fsdp2+domain, tp+fsdp2, tp+domain, fsdp2+tp+domain). The full 3-way combo also passes on 8 GPUs / 2 nodes via `run_smoke_2node.pbs`: mesh `{dp: 2, tp: 2, domain: 2}`, TP applied to 32 blocks, FSDP2 genuinely sharding over dp_mesh (no dp=1 self-skip). That makes the matrix 9/9.

## Update 2026-06-09 (evening): pipeline audit, persistence fixes, longer validation

A second audit pass focused on the training pipeline plumbing (checkpoint/resume, EMA, validation aggregation), validated with longer 2-epoch + resume runs. All fixes are in this PR.

**FSDP2 checkpoint save/resume was broken end-to-end.** Save stored `optimizer.state_dict()` raw — under FSDP2 that holds only rank 0's DTensor shards, silently dropping all other ranks' optimizer state. The load path had no fsdp2 branch at all (it branched on the legacy `mode:` key and crashed on `model.module`). Both now use the DCP full-state-dict APIs. A subtlety found along the way: params that never receive gradients have no lazily-created AdamW state and `set_optimizer_state_dict` raises KeyError on them — WXFormer's `cube_embedding` is allocated but never called when patch sizes are 1 (a model wart worth its own cleanup). Lazy-init state is synthesized at load.

**Guards for unsupported combinations:** `use_ema` + fsdp2 raises with a clear message (#414 tracks the sharded-EMA fix); TP resume raises NotImplementedError and TP save warns (#415 — TP is experimental, checkpoint support deliberately deferred).

**`valid_loss` is now all-reduced** like the train loss. Previously each rank tracked its local value and `fit()` made early-stopping/best-checkpoint decisions from divergent per-rank histories — ranks could stop at different epochs and hang. (gen1 has the same wart; out of scope here.)

**Gradient accumulation no longer syncs every micro-step.** Non-boundary steps suppress DDP all-reduce / FSDP2 reduce-scatter via the per-iteration sync flags, cutting comm cost by grad_accum_every×. DDP `static_graph` is enabled only when not accumulating (incompatible with sync toggling); `find_unused_parameters=True` covers the unused-module case when it's off.

**Replicated-parameter gradients now sync across the TP group** at accumulation boundaries — previously nothing reduced them with `data: none`, and nondeterministic kernels would silently drift the replicas apart.

**Cross-mode parity is now a gated check** (`check_parity.py` in the smoke matrix): tp_domain vs fsdp2_tp_domain must match exactly; ddp vs domain to 5% (crop-vs-interp output operator difference). This codifies the check that caught the dp-rank dataloader bug.

**Smoke configs now train on normalized data**: one-time `credit convert` produced 0.25° BridgeScaler JSONs; all 9 configs use bridgescaler pre/postblocks (losses O(1) instead of ~1e11 on raw ERA5).

**Validation:** 8/8 single-node smoke modes + both parity gates; 2-node 3-way (dp=2×tp=2×domain=2); 2-epoch train → checkpoint → resume-at-epoch-2 with optimizer/scaler/scheduler reload for ddp, fsdp2, and fsdp2+domain, with and without gradient accumulation; 58/58 CPU unit tests.


jsschreck:
@charlie-becker I did wind up having to fix a few bits with the sampling seed so please have a look I did not break anything on the data front.

jsschreck:
@charlie-becker  Note also that 16 files I have brought in are under the new tests/manual section that we discussed. This tests the 9 parallel config options and keeps all needed material within.

credit/models/wxformer/crossformer.py
```diff
@@ -67,6 +67,8 @@ def forward(self, x: torch.Tensor):
 
 
 class UpBlock(nn.Module):
+    _fsdp2_shard = True
+
```
charlie-becker: Let's not hardcode these variables. Please add them to the init so they can be specified; you can leave these as the defaults.

credit/models/wxformer/crossformer.py
```diff
@@ -108,6 +110,8 @@ def forward(self, x):
 
 
 class UpBlockPS(nn.Module):
+    _fsdp2_shard = True
+
     def __init__(self, in_ch, out_ch, num_groups, scale=2, num_residuals=2):
```
charlie-becker: See above.

credit/models/wxformer/crossformer.py
```diff
@@ -202,6 +206,9 @@ def forward(self, x):
 
 
 class FeedForward(nn.Module):
+    _tp_col = "layers.1"  # Conv2d(dim → dim*mult) — output channels sharded
+    _tp_row = "layers.4"  # Conv2d(dim*mult → dim) — input channels sharded + all_reduce
+
     def __init__(self, dim, mult=4, dropout=0.0):
```
charlie-becker: See above.

credit/models/wxformer/crossformer.py
```diff
@@ -354,6 +375,8 @@ def forward(self, x):
 
 
 class Transformer(nn.Module):
+    _fsdp2_shard = True
+
     def __init__(
```
charlie-becker, line 407: see above.

charlie-becker (changes requested):
I left a couple of small comments regarding hard coded variables. I will run some of these manual tests myself to verify this is working as intended for both gen 1 and gen 2 and report back.

jsschreck:
@charlie-becker Addressed in 4cdae2f — the `_fsdp2_shard` / `_tp_col` / `_tp_row` class attributes on UpBlock, UpBlockPS, Transformer, FeedForward, and Attention are now `__init__` parameters with the previous values as defaults, so they can be specified per instance. The detection helpers (`_has_fsdp2_shard`, the TP walker) read instance attributes and still recognize class-attribute declarations, so third-party blocks using either style work. Added tests covering the defaults, per-instance overrides, and the legacy class-attr style.

Heads up on two things that landed since your review, in case they affect your manual runs: (1) a seeding fix (35686f8) — model init is now seeded identically on all ranks before construction (FSDP2 never broadcasts params, so per-rank init seeds made the global model a mixture of inits), with the per-dp-rank seed applied after wrapping for runtime diversity; (2) the manual matrix now trains every mode multistep and adds a save/resume phase per mode — `qsub tests/manual/gen2_parallelism/run_smoke.pbs` runs 10 training legs plus a ddp-vs-domain parity gate. Latest full run is 14/14 green including a 2-node fsdp2+domain job.

charlie-becker:
@jsschreck I spent time testing the new parallel schema and found it to mostly work well. There was one bug regarding FSDP that crashed when `grad_max_norm` was set to "dynamic", which I fixed. The other thing I found is that Tensor Parallelism does not work here, it is blocked all together, despite issue #415 claiming it only blocks checkpoints (see comment below). Lastly, I made some minor changes to eliminate warnings (not suppressed). 
 Oh, and I discoverd that domain parallelism only works when the domain size is evenly divisible by n_GPUs. There are ways around it with padding, but I definitely don't think it necessary now. The error is clear, but we may want to look into this at a later date. 
 
 Personally, I'm okay not waiting on TP to work until it's fully functional (#415). So if you are good with these changes. I am okay merging. 
  
  
  credit/distributed.py

  - Added device_id parameter to setup(), forwarded to dist.init_process_group — eliminates PyTorch 2.3+ barrier() UserWarning

  credit/applications/train_gen2.py

  - Moved device assignment before setup() so device_id can be passed at init time
  - Added if not root.handlers: guard before addHandler — 
  
  Combined, this prevents duplicate log messages on single-GPU runs

  credit/trainers/trainer_gen2.py

  - In grad_max_norm: dynamic path, added .to_local() call on DTensor grads before .norm(2) — fixes FSDP2 crash (NotImplementedError: Operator c10d.allreduce_.default)
  
  The above was not tested in the smoke

  credit/trainers/utils.py

  - Added persistent_workers=num_workers > 0 to DataLoader — workers persist across epochs, reducing fork deprecation warnings to startup-only and slightly spedsup data loading

credit/parallel/tensor_parallel.py
```diff
@@ -0,0 +1,362 @@
...
+    """
+    raise NotImplementedError(
+        "Tensor parallelism (trainer.parallelism.tensor > 1) is disabled: the "
+        "hand-rolled column sharding slices fused projections (e.g. WXFormer's "
+        "to_qkv) across q/k/v boundaries, and the backward all-reduce at the "
+        "column-parallel input (Megatron's 'f' operator) is missing, so "
+        "tensor > 1 silently trains wrong outputs and gradients. Native TP "
+        "via torch parallelize_module lands with issue #415. Set "
+        "trainer.parallelism.tensor: 1."
+    )
```
charlie-becker, line 293: Just flagging this, as issue #415 claims it works from scratch and only disabled from checkpointing, but it appears to not implemented all together.

charlie-becker (approved):

PR #409, opened 2026-05-30
fix(EMATracker): strip legacy weight_u/v from shadow on resume to prevent val loss explosion

jsschreck:
## Summary

- `EMATracker.load_state_dict` now strips `weight_u` / `weight_v` keys from the loaded shadow and logs a warning if any are found
- Checkpoints saved before 2026-05-15 (prior to the `_is_spectral_norm_buffer` filter in #TODO) include EMA-averaged power-iteration vectors in the shadow; leaving them causes validation loss to explode on the first epoch after resume
- Two new regression tests: one unit test verifying the strip, one end-to-end test verifying val loss stays stable after resume from a legacy checkpoint

## Root cause

`nn.utils.spectral_norm` stores `weight_u` and `weight_v` as buffers alongside `weight_orig`. Before the spectral-norm filter was added, `EMATracker.update()` included these in the shadow. EMA-averaging unit vectors reduces their norm: `||EMA(u_t)|| < 1`. When `swap()` loads the averaged vectors into the model, `sigma = u^T W v` is underestimated. In eval mode (no power iteration to correct it), `weight = weight_orig / sigma` inflates across every spectral-norm layer — compounding to a ~4000x loss explosion (observed: `0.074 → 466`).

## Test plan

- [x] `pytest tests/test_trainer_components.py::TestEMATracker` — all pass
- [x] `pytest tests/test_trainer_components.py::TestEMATrackerEdgeCases` — all pass, including two new tests
- [x] Empirically confirmed: injecting legacy u/v into shadow → loss explodes; with fix applied → loss stable at reference value

### DIFF
--- a/credit/trainers/base_trainer.py
+++ b/credit/trainers/base_trainer.py
@@ -116,6 +116,21 @@ class EMATracker:
         self.decay = d["decay"]
         self.shadow = d["shadow"]
         self.step = d.get("step", 0)
+        # Checkpoints saved before the _is_spectral_norm_buffer filter was added
+        # (pre-2026-05-15) may have EMA-averaged weight_u / weight_v in the shadow.
+        # Averaged unit vectors are not unit vectors, so sigma = u^T W v becomes
+        # wrong in eval mode and causes val loss to explode on resume.  Strip them.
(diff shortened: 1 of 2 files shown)


djgagne (approved):
Changes look good to me. Thank you for the quick fix and hope this helps the CWA team.

PR #411, opened 2026-06-03
CREDIT preprocess app and bridgescaler fitting support

### LINKED ISSUE #345: credit preprocess CLI application
We need a CLI application to load data, run the preblocks up to the bridgescaler, and fit the bridgescaler scaler objects to enough data to create representative scaling values.

djgagne:
#### What this PR does
Adds an end-to-end **preprocessing / scaler-fitting pipeline** to CREDIT and wires it into the CLI, so scaler statistics can be computed over the training data as a first-class, config-driven step (locally or as a PBS job).
#### Highlights
- **New `preprocess` application** (`credit/applications/preprocess.py`) — fits `bridgescaler` scalers over the training data batch-by-batch, merges per-rank/per-batch fits, and saves the combined scaler to disk. Supports distributed runs (DDP/FSDP) with rank-0 gather + combine.
- **CLI integration** — adds `credit preprocess` (standalone) and `credit submit preprocess` (generate/submit a PBS job on Casper/Derecho), following the existing `train` / `rollout` / `realtime` patterns.
- **`FillNan` preblock** (`credit/preblock/nan.py`, registered as `fill_nan`) — replaces NaNs with a configurable `fill_value` (default `0`) for a selectable subset of variables, so multiple fill policies can be composed. Only NaNs are touched; `±inf` and finite values are preserved.
- **Variable-selection helper** (`credit/preblock/_utils.py::_parse_variable_selection`) — expands partial/empty variable names against the nested batch dict, shared by the scaler transform and `FillNan`.
- **Fitted-scaler logging** — after saving, `preprocess` logs per-variable/per-level statistics: mean/variance (standard), min/max (minmax), plus an inverse-transform of the normalized range for all scaler types (including quantile) to show the physical values covered.
- **Distributed improvements** (`credit/distributed.py`) — rank-info and gather handling needed for multi-process scaler fitting; sets `channels_last=False` to fix scaler fitting.
### Tests & config
- New tests: `tests/test_preprocess.py`, `tests/test_preblock_nan.py`, and additions to `tests/test_preblock.py` covering scaler fitting, variable selection, and NaN filling.
- Example configs added under `config/.../gen_2/examples/`, including a small "tiny" config for fast end-to-end runs.
### Notes for reviewers
- A benign `ResourceWarning` from `gcsfs`/`aiohttp` (unclosed async SSL transport at teardown) is filtered in `preprocess.py`; it doesn't affect output.
- For `credit preprocess`, place any `fill_nan` block **before** the scaler block in the config so NaNs are filled prior to fitting.

### Related Issues and PRs

Closes #345 

### Checklist

### DIFF
--- a/config/gen_2/examples/weatherbench2_era5_wxformer.yml
+++ b/config/gen_2/examples/weatherbench2_era5_wxformer.yml
@@ -38,7 +38,7 @@ preblocks:
     args:
       variables:
         - 'WBERA5/prognostic/3d/specific_humidity'
-        - "WBERA5/prognostic/3d/surface_pressure"
+        - "WBERA5/prognostic/2d/surface_pressure"
         - "WBERA5/diagnostic/2d/total_precipitation_6hr"
       data_types:
         - 'input'
--- a/config/gen_2/examples/weatherbench2_era5_wxformer.yml
+++ b/config/gen_2/examples/weatherbench2_era5_wxformer.yml
@@ -46,6 +46,7 @@ preblocks:
   scaler:
     type: "bridgescaler_transformer"
     args:
+      variables: []
       scaler_path: "/Users/dgagne/data/weatherbench2/scaler.json"
       scaler_type: "standard"
 
--- a/config/gen_2/examples/weatherbench2_era5_wxformer.yml
+++ b/config/gen_2/examples/weatherbench2_era5_wxformer.yml
@@ -1,4 +1,5 @@
-save_loc: '/glade/derecho/scratch/$USER/CREDIT/weatherbench2'
+save_loc: '/Users/dgagne/data/weatherbench2'
+#save_loc: '/glade/derecho/scratch/$USER/CREDIT/weatherbench2'
 seed: 31415
 
 data:
@@ -42,12 +43,29 @@ preblocks:
       data_types:
         - 'input'
         - 'target'
+  scaler:
+    type: "bridgescaler_transformer"
+    args:
+      scaler_path: "/Users/dgagne/data/weatherbench2/scaler.json"
+      scaler_type: "standard"
+
   vars_concat:
     type: "concat"
 
 postblocks:
   reconstruct:
     type: "reconstruct"
+
+trainer:
+  type: era5-gen2           # use the v2 trainer (new nested data schema)
+
+  mode: ddp
+  epochs: 100
+  train_batch_size: 32
+  batches_per_epoch: 100
+  thread_workers: 2
+  valid_thread_workers: 2
+  prefetch_factor: 2
 
 
 
--- a/config/gen_2/examples/weatherbench2_era5_wxformer.yml
+++ b/config/gen_2/examples/weatherbench2_era5_wxformer.yml
@@ -1,0 +1,54 @@
+save_loc: '/glade/derecho/scratch/$USER/CREDIT/weatherbench2'
+seed: 31415
+
+data:
+  source:
+    WBERA5:
+      dataset_type: "weatherbench2_era5"
+      resolution: "240x121"   # optional; overridden by the resolution kwarg
+      level_coord: "level"
+      levels: [ 50, 100, 150, 200, 250,  300, 400, 500, 600, 700, 850, 925, 1000 ]
+      variables:
+        prognostic:
+          vars_3D: ["u_component_of_wind", "v_component_of_wind", "temperature", "specific_humidity", "geopotential"]
+          vars_2D: ["surface_pressure", "2m_temperature", '2m_dewpoint_temperature',
+                    '10m_u_component_of_wind', '10m_v_component_of_wind']
+        diagnostic:
+          vars_2D: ['total_precipitation_6hr', 'total_cloud_cover', 'sea_ice_cover',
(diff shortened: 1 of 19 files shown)


djgagne:
I fixed an issue with the target scaler values not appearing due to argument order issues in WB2ERA5. Please test on your own when you get a chance. Also, main has been merged in already.

config/gen_2/examples/weatherbench2_era5_wxformer.yml
```diff
@@ -0,0 +1,75 @@
...
+    type: "bridgescaler_transformer"
+    args:
+      variables: []
+      scaler_path: "/Users/dgagne/data/weatherbench2/scaler.json"
+      scaler_type: "standard"
+      scaler_params:
+        channels_last: False
+
+  vars_concat:
+    type: "concat"
```
kevinyang-cky: This is the old flat preblock format and needs to be migrated to the two-section layout (e.g., under either `ic_only` or `per_step`).

charlie-becker (approved):
I made a couple of small fixes:

The config and some of the tests did not account for the extra `per_step` key needed in the config for preblocks. I also made a small change to the preprocess which allows you to run on all of the data instead of a specified amount of `batches_per_epoch`. If it is omitted or set to 0, it will use all data. 

I successfully tested this on Casper through the CLI `credit submit` using 4 GPUs and 32 workers over a single node with various types of scalers (on 35 years of small weatherbench data) and it was successful. This is a great addition! Once CI passes, I'll merge.

credit/preblock/scaler.py
```diff
@@ -22,13 +60,63 @@ class BridgeScalerTransformer(BasePreblock):
...
-        self.scaler_path = scaler_path
-        self.scaler = load_scaler_dict(scaler_path)
+        self.scaler_path = expandvars(scaler_path)
+        if scaler_params is None:
+            scaler_params = {}
+        self.scaler_params = scaler_params
+        # ``scaler`` holds the nested dict of fitted scalers used for transform /
+        # inverse_transform (loaded from disk, or accumulated by fit_scaler_batch).
+        # ``scaler_template`` is the single unfitted prototype used to fit new data.
+        if exists(expandvars(self.scaler_path)):
```
kevinyang-cky, line 82: Should it be `if exists(self.scaler_path)`?

credit/preblock/scaler.py
```diff
@@ -22,13 +60,63 @@ class BridgeScalerTransformer(BasePreblock):
...
+            self.scaler = load_scaler_dict(scaler_path)
```
kevinyang-cky, line 83: Should it be `self.scaler = load_scaler_dict(self.scaler_path)`?

credit/preblock/scaler.py
```diff
@@ -22,13 +60,63 @@ class BridgeScalerTransformer(BasePreblock):
...
+            self.scaler_template = None
+        else:
+            full_scaler_path = expandvars(self.scaler_path)
```
kevinyang-cky, line 86: Should it be `full_scaler_path = self.scaler_path`?

credit/applications/preprocess.py
```diff
@@ -0,0 +1,191 @@
...
+            scaler_block_key = k
+            break
+    if scaler_block_key is None:
+        raise ValueError("BridgeScalerTransformer not found in preblocks.")
+    batches_per_epoch = trainer_conf.get("batches_per_epoch", 1)
+    dl = cycle(train_loader)
+    for i in range(batches_per_epoch):
+        root.info(f"Worker {rank}: Processing batch {i} of {batches_per_epoch}.")
+        batch = next(dl)
+        processed_batch = apply_preblocks_before_scaler(preblocks, batch, device)
```
kevinyang-cky, line 168: Question: I'm assuming that if a GPU is available and one chooses to perform minmax and standard fitting, all operations will run on the GPU, correct? But how does it work for a quantile scaler, since the fitting part isn't in PyTorch?

kevinyang-cky (commented):
(This PR is already merged, but I will just add my comments here for future reference.)

I think @charlie-becker already addressed my first comment; comments 2-4 are minor edits and won't break anything.

@djgagne, if you could provide some insights into comment 5, that would be really appreciated. It is not clear to me whether the operations are on the GPU or the CPU. 

Tested both [config/gen_2/examples/weatherbench2_era5_wxformer.yml‎](https://github.com/NCAR/miles-credit/pull/411/changes#diff-5900a2f20b1031bd8507d6fca0eda2433a85261b2de22ca40d7384cd346b9cee) and [‎config/gen_2/examples/weatherbench2_era5_wxformer_tiny.yml](https://github.com/NCAR/miles-credit/pull/411/changes#diff-c8c6f1d2dba3bc8e67813a03f60e47214b9eac4e753e44f1a7ae64e3c7115c02) and did not run into any issue. 

I also tested with GOES data using a config called [preprocess_goes.yml](https://github.com/NCAR/miles-credit/blob/goes-preprocess/config/gen_2/examples/preprocess_goes.yml) in this branch [goes-preprocess](https://github.com/NCAR/miles-credit/tree/goes-preprocess) and validated against the result from a script that simply loads the data the same way and uses Bridgescaler to do the fitting; it produces the same result.

PR #412, opened 2026-06-03
fix: four CLI bugs found during systematic testing

jsschreck:
## Summary

Systematic testing of the `credit` CLI surface turned up four bugs, all in `credit/cli/`. Each was found by running the command, observing the failure, and tracing it to the root cause.

- **`credit init`** — crashes with `FileNotFoundError` when the output path's parent directory does not exist. Fixed by calling `os.makedirs(parent, exist_ok=True)` before `shutil.copy`.

- **`credit submit --mode rollout --cluster casper`** — generated PBS scripts were missing `module load conda/latest` and `conda activate`. Every other Casper script (train, realtime) included these lines; the rollout generator did not. Jobs would fail immediately on submission.

- **`credit plot`** — crashed with `KeyError: 'input'` because `apply_preblocks()` renames the batch keys from `'input'`/`'target'` to `'x'`/`'y'`. Fixed with `.get("x", .get("input"))` fallback and a clear error message if neither key is present.

- **`credit plot`** — two related issues: (1) used `conf.get("data_valid", ...)` to find the validation dataset config, but all v2 configs use the key `validation_data`, so the fallback to training data always triggered and plots sampled the wrong date range; (2) `_build_denorm_stats` looked for norm stats at `preblocks.norm.args` (v1 schema) instead of `preblocks.per_step.norm.args` (v2 schema), causing `KeyError: 'mean_path'` whenever `--denorm` was passed with a v2 config. Both fixed together.

## Test plan

- [ ] `credit init -o /tmp/newdir/config.yml` — should create the directory and write the file
- [ ] `credit submit --cluster casper --mode rollout -c config.yml --dry-run` — PBS script should contain `module load conda/latest` and `conda activate`
- [ ] `credit plot -c config.yml --field t2m` — should produce a PNG without crashing
- [ ] `credit plot -c config.yml --field t2m --denorm` — should produce a denormalised PNG when norm stats are configured

### DIFF
--- a/credit/cli/_convert.py
+++ b/credit/cli/_convert.py
@@ -526,6 +526,9 @@ def _init(args: argparse.Namespace) -> None:
         print(f"File already exists: {output}  (use --force to overwrite)", file=sys.stderr)
         sys.exit(1)
 
+    parent = os.path.dirname(output)
+    if parent:
+        os.makedirs(parent, exist_ok=True)
     shutil.copy(template, output)
     print(f"Created  : {output}")
     print(f"Template : {template_rel}")
--- a/credit/cli/_plot.py
+++ b/credit/cli/_plot.py
@@ -44,7 +44,16 @@ def _build_denorm_stats(conf):
     prog = v.get("prognostic") or {}
     diag = v.get("diagnostic") or {}
 
-    norm_args = conf.get("preblocks", {}).get("norm", {}).get("args", {})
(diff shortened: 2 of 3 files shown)


charlie-becker (approved):
Now that the other parallel changes were stripped, this looks good.

charlie-becker:
GTG. Merging.

PR #416, opened 2026-06-10
EMA support under FSDP2 via sharded shadow

### LINKED ISSUE #414: EMA support under FSDP2 (sharded shadow)
## Problem

`use_ema: true` with `parallelism: {data: fsdp2}` currently raises a ValueError (added during the #407 audit). EMATracker clones `model.state_dict()`, which under FSDP2 returns this rank's DTensor shards — the shadow would mix DTensor shards with plain CPU tensors and only cover 1/dp of the weights. Production runs use EMA, so this blocks production-quality FSDP2 training.

## Design

EMA is elementwise, so each rank can EMA its own DTensor shard locally — mathematically identical to EMA of the full weights, zero extra communication per step.

- **EMATracker sharded mode**: shadow = `to_local()` clones (CPU); `update()` unchanged (already elementwise); `swap()` copies shards back in place. ~40 lines.
- **Save**: gather once per epoch — `DTensor.from_local(shadow, mesh, placements).full_tensor()` per key, rank 0 writes `checkpoint_ema.pt` in the normal full format so rollout/inference loads it unchanged. ~40 lines.
- **Resume**: load full EMA dict, slice each tensor to the local shard using the live param's placements. ~30 lines.
- Remove the ValueError guard in BaseTrainer; add a `use_ema: true` fsdp2 leg to the plumbing test (`tests/manual/gen2_parallelism` + gen2_plumbing harness) to prove the save→resume round-trip.

Edge cases already handled: spectral-norm u/v exclusion is name-based; all FSDP2 params are uniformly DTensors. Save-path gather memory matches the existing checkpoint gather (no new constraint).

## Estimate

~1–1.5 days including GPU validation. Do this before any TP checkpoint work — it unblocks production FSDP2 runs.

jsschreck:
Closes #414.

## Summary

EMA now works under FSDP2. The design exploits that EMA is elementwise: each rank maintains the EMA of its own DTensor shard, which is mathematically identical to EMA of the full weights and adds zero communication per training step. The full shadow is only materialized at checkpoint save (a collective gather) and re-sharded on resume.

**EMATracker changes** (auto-detects DTensor state; non-sharded path unchanged):
- `__init__` / `update` operate on `to_local()` views — shadow holds local shards on CPU
- `swap()` copies shards in place; DTensor locals are views of the param storage, so no `load_state_dict` round trip
- `state_dict()` gathers the full shadow via `DTensor.from_local(...).full_tensor()` and rank 0 writes the standard `checkpoint_ema.pt` format — rollout/inference and non-FSDP2 resume load it unchanged
- `load_state_dict()` re-shards the full checkpoint via `distribute_tensor` using the live params' placements

The `use_ema` + fsdp2 ValueError guard is removed, and the fsdp2 checkpoint branch now saves EMA (it previously never did, even before the guard).

## A bug the GPU test caught

The first validation run deadlocked at save: `DTensor.from_local` infers the global shape as `local x world_size`, which is wrong for FSDP2's unevenly padded dim-0 shards — ranks disagreed on the all_gather size and hung until the NCCL watchdog fired. Fixed by passing the live param's explicit `shape`/`stride`. This failure mode cannot surface in mocked CPU tests, which is exactly why the plumbing leg exists.

## Validation

- Derecho gpudev, fsdp2 + `use_ema: true`: 2-epoch train → `checkpoint_ema.pt` + `best_checkpoint_ema.pt` written → resume reloads EMA ("Resumed EMA from") and continues at epoch 2
- Regression legs (ddp / fsdp2 / fsdp2+domain, grad_accum=2) stay green
- 15/15 existing EMA unit tests pass (non-sharded path untouched)

Base branch is `v2/fsdp2-parallel` (#407) — this builds on the checkpoint/resume fixes there and should merge after it.

### DIFF
--- a/credit/trainers/base_trainer.py
+++ b/credit/trainers/base_trainer.py
@@ -66,13 +66,28 @@ class EMATracker:
     def __init__(self, model: torch.nn.Module, decay: float = 0.9999):
         self.decay = decay
         self.step = 0
+        # Keep a reference for the sharded (FSDP2/DTensor) path: gathering the
+        # full shadow at save time and re-sharding on load need the live
+        # params' mesh/placements.
+        self._model = model
         # shadow lives on CPU to save GPU memory
         self.shadow: OrderedDict = OrderedDict()
         state = model.module.state_dict() if hasattr(model, "module") else model.state_dict()
+        # FSDP2 state dicts hold DTensors. EMA is elementwise, so each rank
+        # EMAs its own LOCAL shard — mathematically identical to EMA of the
+        # full weights, with zero extra communication per step. The full
+        # shadow is only materialized at save time (state_dict gathers) and
+        # re-sharded on load.
+        self._sharded = any(hasattr(v, "to_local") for v in state.values())
         for k, v in state.items():
             if self._is_spectral_norm_buffer(k, state):
                 continue  # power-iteration vectors must not be EMA-averaged
-            self.shadow[k] = v.detach().float().cpu().clone()
+            self.shadow[k] = self._local(v).detach().float().cpu().clone()
+
+    @staticmethod
+    def _local(v: torch.Tensor) -> torch.Tensor:
+        """Return the local shard of a DTensor (a view of its storage), else v."""
+        return v.to_local() if hasattr(v, "to_local") else v
 
     @staticmethod
     def _is_spectral_norm_buffer(key: str, state: dict) -> bool:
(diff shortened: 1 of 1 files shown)


jsschreck:
Folded into #407 as commit 029da90 (cherry-picked verbatim, all trainer tests pass there). Closing to keep the merge chain to a single PR.

PR #417, opened 2026-06-10
Gen2 ensemble training: ring-CRPS as a loss on trainer_gen2

jsschreck:
## Summary

Brings ensemble training up to gen2 specs as **one trainer, not two**: ensemble training is a loss choice (`loss.training_loss: ring-crps`) on `TrainerERA5Gen2`, not a separate trainer class. The ring-reduce CRPS from the `ensemble-v2` branch is ported to `credit/losses/crps.py` and registered in `base_losses`.

Stacked on `v2/fsdp2-parallel` (#407); GitHub will retarget to `main` when that merges.

## Why one trainer

The `TrainerEnsembleGen2` on `ensemble-v2` was a fork of an older `trainer_gen2` plus ensemble. It had already drifted: old preblock API, manual rollout assembly, no grad-accum sync suppression, no valid-loss all-reduce, and the dynamic grad-norm mixed-units bug fixed in #407. The actual ensemble delta is small (training loss + spread logging + member expansion, which the gen2 trainer already had), so it becomes a registered loss instead of a second 400-line trainer.

## Semantics fixed relative to ensemble-v2

The ring loss treats each data-parallel rank as one ensemble member of the **same forecast** - so every dp rank must see the same training batches. On `ensemble-v2`, the dataloader still sharded samples across ranks (with per-rank shuffle seeds), so ring members were predictions of *different samples*. Now:

- `train_gen2` disables dp data sharding for ring-crps training; member diversity comes from the existing per-dp-rank seed (dropout, SKEBS, IC perturbations).
- The ring runs over the **dp group** registered by `distributed_model_wrapper_gen2`, so TP/domain ranks (identical batches by contract) are never counted as members.
- Validation keeps dp sharding and the standard criterion (eval-mode forwards are deterministic).

## Guards

- `ring-crps` + `use_latitude_weights`/`use_variable_weights` raises (`VariableTotalLoss2D` needs an elementwise loss; the ring returns a scalar).
- `ring-crps` + `trainer.ensemble_size > 1` raises (batch-axis members would pair member i on rank r only with member i on rank j, biasing the spread).

## Gen1/gen2 equivalence (tested)

`tests/test_ring_crps.py` proves the chain:

1. The gen1 ensemble trainer's computation (Gather + per-latitude-slice KCRPS, `biased=False`) equals a fair-CRPS autograd reference in **value and gradient**.
2. The ring loss, run for real over 2- and 3-process gloo groups, matches the same reference's **gradient** per member, and its per-rank value matches the documented local formula.

Same members in, identical parameter gradients out. Only the logged loss value differs by design: the ring's detached spread counts each pair twice and carries a 1/K factor (the docstring documents this).

## Validation

- 110 tests pass locally (`test_ring_crps`, `test_loss`, `test_fsdp2_parallel`, `test_trainer_components`).
- 2x V100 DDP smoke (`config/gen_2/smoke/smoke_ensemble_gen2_casper.yml`): full train/validate/checkpoint cycle, `train_std` ~3.1 (nonzero member spread from per-rank dropout), train and valid losses decreasing across 2 epochs, exit 0.


djgagne:
@jsschreck Should this be merged into main rather than v2/fsdp2-parallel, or do you plan to merge into that branch first before merging into main?

jsschreck:
@djgagne should be updated now.

djgagne (approved):
I updated the smoke config to follow the latest gen 2 standards, replaced the predict section with Charlie's new inference section, and adjusted file paths to be more generic.

PR #418, opened 2026-06-11
Native tensor parallelism for wxformer_next (closes #415)

### LINKED ISSUE #415: Tensor parallelism: experimental status; checkpoint support deliberately deferred
## Status

TP (`parallelism: {tensor: N}`) trains correctly after the #407 fixes (replicated-grad sync, row-bias post-reduce, dp-rank data contract) and passes the smoke matrix, but **checkpoints are not usable**: saving warns and emits only this rank's TP shards under rewritten keys (`block.proj_up.conv.weight` instead of `block.proj_up.weight`), and resume raises NotImplementedError. Treat TP as experimental / throughput-research only.

## Why deferred (decision, not oversight)

Our TP is hand-rolled module swapping (`TpColConv2d`/`TpRowConv2d`/`TpColLinear`/`TpRowLinear`) because WXFormer's 1x1-conv FFNs and conv attention projections don't map onto torch's DTensor TP (`parallelize_module` is Linear-centric). Module swapping renames params and changes shapes, so checkpointing, EMA, and optimizer state all see a different model than the serial one — supporting them means inverting the rewrite everywhere.

Weather models at CREDIT's scale don't need TP: weights fit on one GPU (FSDP2 handles weight memory, AC + domain parallelism handle activation/resolution memory). TP is load-bearing only when single layers outgrow a GPU (LLM-scale). No current CREDIT model is there.

## If/when a model needs it

Phase 1 (~1.5 days): gather-on-save (all-gather col dim 0 / row dim 1, restore original key names) + weights-only resume by loading the full checkpoint into the raw model BEFORE apply_tensor_parallel (optimizer restarts fresh, warn loudly). Covers fine-tune/restart.
Phase 2 (~1 day): optimizer-state gather/reshard across tp degrees.
Better long-term: if a future model is Linear-based or large enough to justify it, port to DTensor TP instead of extending the wrappers — checkpointing then comes for free via get_model_state_dict, like FSDP2.

## Trigger to revisit

A model whose FFN/attention widths exceed single-GPU memory even with FSDP2 + activation checkpointing + domain parallelism.

jsschreck:
Native DTensor tensor parallelism for the wxformer_next family, replacing the disabled hand-rolled TpCol/TpRow path. The acceptance gate passed on Derecho: **tp=1 and tp=2 produce digit-for-digit identical loss trajectories** from the same seed (4 GPUs, fsdp2 dp=2 × tp=2), and a tp=2 run resumes from a DCP checkpoint and trains a second epoch. Closes #415.

Stacked on #407 (`v2/fsdp2-parallel`); retarget to main after it merges.

## What's here

- **wxformer_next on nn.Linear projections.** The transformer blocks' 1×1 Conv2d projections are now nn.Linear, and the fused `to_qkv` is split into separate `to_q`/`to_k`/`to_v` (torchtitan pattern). ColwiseParallel sharding is per-head-correct by construction, so the q/k/v boundary scrambling that sank the legacy path is structurally impossible. Attention derives its head count from the local channel width (heads/tp per rank).
- **Checkpoint remap.** `remap_conv_state_dict` converts old checkpoints on load: qkv split + (O,I,1,1) → (O,I), spectral-norm weight_orig/u/v handled, idempotent. Conv-vs-Linear numerical equivalence is tested to 7e-7.
- **Opt-in native TP.** `apply_native_tensor_parallel` builds Colwise/Rowwise plans from a `_tp_plan` class attribute and composes with `fully_shard` over the dp×tp mesh. Models without the opt-in still hit the legacy NotImplementedError. Checkpointing goes through the DCP full-state APIs, so native TP + fsdp2 saves/resumes like plain fsdp2.
- **Reduced TP under spectral norm.** SN-wrapped blocks are skipped as whole colwise/rowwise pairs (warn, not raise). Note `apply_spectral_norm` wraps every layer in the transformer stages, so with `use_spectral_norm: true` nothing shards and a loud warning says so; full TP needs `use_spectral_norm: false` for now. A DTensor-compatible spectral norm is the long-term fix.
- **Mixed-mesh gradient clipping.** TP'd params live on the (dp, tp) mesh and the rest on the dp mesh; torch's `clip_grad_norm_` cannot stack norms across meshes. New `clip_grad_norm_`/`total_grad_norm` in `credit/parallel/collectives.py` group by mesh and combine; the plain-tensor path is bit-identical to torch's (tested).
- **Init seeding fix.** Per-dp-rank seeding before model construction made fsdp2's global init a mesh-layout-dependent mixture (fully_shard never broadcasts). All ranks now seed identically for construction, then re-seed per dp rank for runtime diversity. This commit is also cherry-picked onto #407 (35686f8) since it affects every fsdp2 run.

## Testing

- 843 tests pass (47 new), including two real 2-process gloo DTensor tests: tp=2 matches serial to 7e-7 with gradients reaching all params, and the mixed-mesh clip reproduces and fixes the exact GPU crash.
- GPU acceptance (`tests/manual/gen2_parallelism/run_tp_parity.pbs`, Derecho job 6431309): tp1 PASS, tp2 PASS, tp2 DCP resume PASS, parity tp1 == tp2 EXACT.


djgagne (approved):
I merged main into here and fixed a floating point tolerance issue with one of the tests. Otherwise everything looks good to go. Will merge as soon as the last test finishes.

PR #419, opened 2026-06-15
Fix device management in preprocessing pipeline

kevinyang-cky:
### Description

- Fix `apply_preblocks_before_scaler` to actually move batch tensors to the target device — previously the `device` argument was accepted but never applied, so batches always stayed on their original device regardless of what was passed in.
- Add `--device` CLI arg to `preprocess.py` for explicit device selection and change `--backend` default to `None`, auto-selecting `nccl` for GPU and `gloo` for CPU. This gives users the flexibility to run on CPU even when a GPU is available — for example, when preprocessing is I/O-bound rather than compute-bound, keeping the computation on CPU is sufficient.
- Add `move_scaler_dict_to_cpu` to move scaler tensor statistics to CPU before `gather_object`. When using multi-GPU runs, each rank holds scaler tensors on its own device, which causes a device-mismatch error during gathering. Moving all tensors to CPU first resolves this.
- Fix `BridgeScalerTransformer.__init__` to use `self.scaler_path` directly instead of wrapping it in `expandvars()`.

### Related Issues and PRs
Address comments in #411.

### Checklist

### DIFF
--- a/credit/applications/preprocess.py
+++ b/credit/applications/preprocess.py
@@ -109,6 +109,13 @@ def main():
     parser = argparse.ArgumentParser(
         epilog="""
 Examples:
+  # Recommended: submit via PBS using the credit CLI.
+  # Note: torchrun is hardcoded internally; --device and --backend cannot be forwarded.
+  # Device selection is automatic based on hardware detected at runtime (GPU if available, CPU otherwise).
+  credit submit --cluster casper  -c config.yml --mode preprocess
+  credit submit --cluster derecho -c config.yml --mode preprocess
+
+  # For reference only (not the recommended path): run directly with torchrun.
   # GPU run (auto-selects nccl backend)
   torchrun --nproc_per_node=4 preprocess.py -c config.yml
 
--- a/credit/applications/preprocess.py
+++ b/credit/applications/preprocess.py
@@ -13,7 +13,7 @@ import yaml
 import sys
 import shutil
 from credit.preblock import build_preblocks, apply_preblocks_before_scaler, BridgeScalerTransformer
-from credit.preblock.scaler import combine_scaler_dicts
+from credit.preblock.scaler import combine_scaler_dicts, move_scaler_dict_to_cpu
 from credit.trainers.utils import cycle, load_dataset, load_dataloader
 import torch.distributed as dist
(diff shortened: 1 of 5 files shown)


credit/applications/preprocess.py
```diff
@@ -106,19 +106,37 @@ def main():
     # teardown (after the scalers have already been fitted and saved). Silence it.
     warnings.filterwarnings("ignore", category=ResourceWarning)
     warnings.filterwarnings("ignore", category=UserWarning, module="bridgescaler")
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(
+        epilog="""
+Examples:
+  # GPU run (auto-selects nccl backend)
+  torchrun --nproc_per_node=4 preprocess.py -c config.yml
```
djgagne, line 120: Ideally a user should not be calling preprocess.py directly and instead should call the function through the CREDIT CLI. Right now the only way to do parallel calls is through CREDIT submit but there are certainly cases where a user would want to perform a parallel run on their laptop or on an interactive compute node. We should add another CLI option to launch torchrun on preprocess/train/rollout on a single node without the user having to call torchrun directly.

djgagne: This can be done in a separate PR. For now, change these examples to credit submit ones.

djgagne (changes requested):
Change the docs based on my review comment. Otherwise changes look good.

kevinyang-cky:
@djgagne. Please review the changes that addressed your comment in the last commit.

I’ve included a `credit submit` example and verified it with:
`credit submit --cluster casper -c config/gen_2/examples/weatherbench2_era5_wxformer_tiny.yml --mode preprocess`. The run completed successfully and produced the scaler JSON file. I did need to strip out a few irrelevant keys, and I also added a PBS block to `weatherbench2_era5_wxformer_tiny.yml` to make the example runnable on Casper.

The torchrun example is still included but marked as not recommended. It’s mainly there for cases where someone wants explicit control over device placement during preprocessing (`credit submit` cannot pass the `--device` flag to `preprocess.py`), but feel free to strip it out if you think that's better. Thanks!

djgagne (approved):
Updated docs look good and glad it passes.

PR #421, opened 2026-06-16
Replace FillNan with general-purpose FillValues preblock

kevinyang-cky:
### Summary

- Replaces the `FillNan` preblock (NaN → fixed fill value) with a new `FillValues` preblock that supports arbitrary search-and-replace rules on batch tensors.
- Removes `credit/preblock/nan.py` and `fill_nan` registry entry; adds `credit/preblock/fill_values.py` registered as `fill_values`.
- Replaces `test_preblock_nan.py` with a comprehensive `test_preblock_fill_values.py` test suite.
- Updates `config/gen_2/examples/weatherbench2_era5_wxformer_tiny.yml` to use the new `fill_values` config format.

### What `FillValues` adds over `FillNan`
`FillValues` accepts a list of rules, each specifying:

- `search` — `"nan"` or a numeric threshold
- `op` — comparison operator (`eq`, `ne`, `lt`, `le`, `gt`, `ge`; default `eq`)
- `fill` — replacement constant

All masks are computed on the **original** tensor simultaneously, so rules are independent of one another (a NaN-filled zero is not subsequently matched by a zero rule). Numeric ops never match NaN positions — those must be targeted explicitly via `search: "nan"`. A warning is logged on the first forward pass if any two rules overlap.

Example config:

```
type: "fill_values"
args:
    rules:
        - search: nan
          fill: -1.0
        - search: 0.0
          op: eq
          fill: 1.0e-4
        - search: 0.0
          op: lt
          fill: 0.0        # clamp negatives
```

### Checklist


credit/preblock/fill_values.py
```diff
@@ -0,0 +1,171 @@
...
+
+from credit.preblock.base import BasePreblock
+from ._utils import _parse_variable_selection
+
+logger = logging.getLogger(__name__)
+
+# Explicit torch functions rather than Python operators so the lookup in forward()
+# is a single dict access with no branching.
+_OPS = {
+    "eq": torch.eq,  # x == search
```
djgagne: Can you use the symbols for the ops dictionary and settings rather than "eq", "ne", etc?The symbols are more intuitive to me and should work equivalently.

djgagne (changes requested):
Code looks good other than changing the operation config. Should be an easy fix hopefully.

kevinyang-cky:
@djgagne, please review the changes by using the symbol operator instead. Thanks.

djgagne (approved):
Changes look good. Merging.

Issue #422, opened 2026-06-16
Should scaler type be bridgescaler_transform or bridgescaler_transformer?


ahijevyc:
The command `credit init --grid 1deg` produces a configuration with the preblock registry

```
preblocks:
  per_step:
    scaler:
      type: bridgescaler_transform
...
```
However, preblock only recognizes `bridgescaler_transformer` (with `er` at the end) See [__init__.py](credit/preblock/__init__.py). So it only works if type is changed to bridgescalar_transformer. 

The same string `bridgescaler_transform` is used in other example/converter paths: [example-v2026.2.yml:165](config/gen_2/examples/example-v2026.2.yml) and  [_convert.py:363](https://github.com/NCAR/miles-credit/blob/b7e760d9d99dd7c106606036ffc55dab6fb0ac3d/credit/cli/_convert.py#L363). Postblocks uses `bridgescaler_transform`, as registered in [__init__.py:13](credit/postblock/__init__.py).

Either the preblock registry reference should be changed to `bridgescaler_transform` or all the other references should be changed to `bridgescaler_transformer`.

PR #427, opened 2026-06-18
Add custom objects registry

### LINKED ISSUE #404: Custom objects registry
In order to make it easier for users to incorporate CREDIT as a library in their own repositories, I would like to add a custom objects registry that users can add to their config file to properly import their own custom models, datasets, pre/post blocks, losses, etc. Example config:

```yaml
custom_objects:
  djformer: 
    object_type: model # Assigns object to correct registry of datasets, preblocks, postblocks, models, trainers, losses
    module_path:  djulator.models # importlib supports python module import syntax and can convert file paths to imports
    module_name: DJFormer 
```

You can access module imports from strings using `importlib.import_module`. 

Having this custom object registry in place will allow CREDIT users to test out new modules and functionality within the CREDIT CLI and APIs without having to modify the CREDIT library code directly and get us toward the ideal of most users eventually interacting with CREDIT via PyPI and imports or CLI rather than Github.

kevinyang-cky:
### Summary

This PR introduces a **custom objects registry** so users can plug in external dataset, model, loss, preblock, and postblock classes via the YAML config — without modifying CREDIT source code. **The diff touches many files, but the core logic is concentrated in a few key places — see the breakdown below so you don't have to review everything. Sections that warrant particular attention are called out explicitly and tagged with a suggested reviewer**:

- @djgagne : `credit/registry.py`, `datasets/__init__.py`, `models/__init__.py`, `losses/__init__.py`, `preblock/__init__.py`, `postblock/__init__.py` — key changes explained: [#1](#1-design-choice-custom-classes-must-inherit-from-credit-base-classes), [#2](#2-creditregistrypy-new), [#3](#3-harmonized-__init__py-across-five-modules)
- @charlie-becker : `credit/preblock/`, `credit/postblock/`, `credit/trainers/trainer_gen2.py`, and call sites in `applications/` — key changes explained: [#4](#4-full-config-passed-to-build_preblocks--build_postblocks), [#5](#5-basepreblock--basepostblock-abstract-enforcement)
- @jsschreck  `credit/losses/__init__.py`, `credit/losses/base_losses.py` (removed), `weighted_loss.py`, `downscaling_loss.py` — key changes explained: [#6](#6-creditlossesbase_lossespy-removed)

---

#### 1. Design choice: custom classes must inherit from CREDIT base classes
Registration enforces this — a `TypeError` is raised at registration time if the class does not subclass the required base:

| Object type | Required base |
|-------------|--------------|
| dataset | `credit.datasets.base_dataset.BaseDataset` |
| preblock | `credit.preblock.base.BasePreblock` |
| model | `credit.models.base_model.BaseModel` |
| postblock | `credit.postblock.base.BasePostblock` |
| loss | `torch.nn.Module` |

This ensures custom objects are compatible with CREDIT's internal dispatch without requiring any changes to core code.

#### 2. `credit/registry.py` (new)
The central new addition. `load_custom_objects(conf)` reads the `custom_objects` block from the config, imports each listed class by dotted module path, and registers it into the appropriate CREDIT registry. It is idempotent — repeated calls within the same training run (e.g. from both `build_preblocks` and `build_postblocks`) silently skip already-registered entries.

Usage in config — register the custom class, then reference it by key elsewhere in the same config:
```yaml
# 1. Register
custom_objects:
  MyPreBlock:
    object_type: preblock
    module_path: mypackage.preblock   # dotted Python import path; pip install -e . if needed

  mymodel:                            # key differs from class name → module_name required
    object_type: model
    module_path: mypackage.models
    module_name: MyModel

  MyLoss:
    object_type: loss
    module_path: mypackage.losses

# 2. Use — reference by the key defined above, exactly like a built-in type
preblocks:
  per_step:
    my_norm:
      type: MyPreBlock
      args:
        fill_value: 0.0

model:
  type: mymodel

loss:
  training_loss: MyLoss
```

For a full example covering all five object types (dataset, preblock, model, postblock, loss), see the docstring in `credit/registry.py`.

#### 3. Harmonized `__init__.py` across five modules
`datasets`, `models`, `losses`, `preblock`, and `postblock` now share a consistent internal structure:
- `_XXX_REGISTRY` dict (built-in types)
- `_CLASS_SOURCES` / lazy `__getattr__` (deferred imports — fixes eager import issues)
- `register_*` decorator (validates base class, adds to registry)
- `_load_*_entry` / builder/loader functions

Previously these five modules were inconsistent in layout and some used eager imports that pulled in heavy dependencies at import time.

#### 4. Full config passed to `build_preblocks` / `build_postblocks`
Both functions now receive the **full** config dict (not just `conf["preblocks"]` / `conf["postblocks"]`), so they can see the `custom_objects` block and call `load_custom_objects` before building blocks. The slicing into the relevant sub-dict is now done internally. This required updating all six call sites across applications and the trainer.

#### 5. `BasePreblock` / `BasePostblock` abstract enforcement
`forward()` in both base classes now raises `NotImplementedError` instead of returning `None` silently, matching PyTorch convention.

#### 6. `credit/losses/base_losses.py` removed
This file was misnamed — it contained no base class, just a dict lookup + instantiation function. That logic is now consolidated into `_instantiate_loss` in `losses/__init__.py`.

#### 7. Registry key rename: `bridgescaler_transform` → `bridgescaler_transformer`
Corrects the key to match the class name (`BridgeScalerTransformer`) consistently across the registry, CLI converter, docstrings, example config, and tests. Relates to #422.

---

### Smoke tested
- [`smoke_gen2_casper.yml`](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/smoke/smoke_gen2_casper.yml) — ran fine
- [`example-v2026.2.yml`](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/examples/example-v2026.2.yml) — data loading, model, and loss all initialize correctly; full training not validated (scaler files not available in test environment)

---

### Out of scope/follow-up
- **Custom trainer not supported**: `custom_objects` does not support plugging in a custom trainer.
- **`losses` cleanup**: further consolidation of the loss module structure.
- **`load_*` placement**: some loader functions live in the module's `__init__.py`, others in the submodule itself (e.g. `load_dataset_dataloader`). Worth standardizing in a follow-up.
- **Direct import vs. loader function**: for some registries, the builder goes straight to the class source; others go through an intermediate loader. Worth aligning in a follow-up.

## Related Issues and PRs
Closes #404 #422 

## Checklist


kevinyang-cky:
Marking this as a draft for now—there are upcoming PRs that modify the same files, and this should come after those.

credit/datasets/__init__.py
```diff
@@ -32,6 +49,81 @@ def __getattr__(name):
...
+        _DATASET_REGISTRY[dataset_type] = cls  # store the class under the given key
+        return cls  # must return the class, otherwise it becomes None after decoration
+
+    return decorator
+
+
+def _load_dataset_entry(dataset_type):
+    """Return the class for a registered dataset type, importing lazily if needed.
+
+    NOTE: Not currently used — load_dataset_and_dataloader.py still uses the legacy
```
djgagne: `load_dataset_and_dataloader.py` is only used for CREDIT Gen 1. The `load_dataset` and `load_dataloader` functions in the new `credit_train_gen2` are being imported from `credit.trainers.utils`. The dataset registry needs to be linked up with the `_SOURCE_REGISTRY` in `credit/datasets/multi_source.py`. The note here should be removed.

djgagne (changes requested):
@kevinyang-cky Most of the PR looks good. I would like you to update the dataset registry based on my comment to make sure it interfaces correctly with `credit/datasets/multi_source.py` for Gen 2 workflows.

credit/datasets/__init__.py
kevinyang-cky: @djgagne, will do. Once #444 is merged into main, I will also resolve the conflict and tag you again for a final review.

kevinyang-cky:
@djgagne, `route_to_dataset_class` now falls back to `credit.datasets._DATASET_REGISTRY` when a `dataset_type` isn't a built-in source, so datasets registered via `custom_objects` actually get picked up (previously nothing called `load_custom_objects` on the Gen2 path, so custom registrations never took effect). Also removed the stale note on `_load_dataset_entry` that referenced the old Gen1 behavior.

Separately, I ran `credit submit --mode [preprocess/train/rollout]` end-to-end to verify, and found/fixed an unrelated rollout warning (missing device_id on setup(), and forking a multi-threaded process for the output writer pool)

This PR should be ready for a re-review. Thanks!

djgagne (approved):
Changes look good to me. Merging.

PR #430, opened 2026-06-19
CAMulator: gen2 config + conservation postblocks (tracer/mass/water/energy)

jsschreck:
## What

Adds the **CAMulator** model to the Gen2 (`era5-gen2`) training path:
- New config `config/gen_2/camulator/camulator_gen2_casper.yml`.
- All four conservation postblocks wired through the gen2 postblock schema (index-based, `args.post_conf`), using the native registry types — no forked code: `tracer_fixer`, `global_mass_fixer`, `global_water_fixer`, `global_energy_fixer_updown`.
- `credit/postblock/gen1.py`: backward-compatible dummy-time-dim shim for the 4D gen2 hand-off (no-op on 5D), and a config-gated `TOA_down_solar_from_input` flag so the up/down energy fixer can source SOLIN (an input-only dynamic forcing) from `x["x"]`.
- Registers the existing `GlobalEnergyFixerUpDown` in `POSTBLOCK_REGISTRY`.

## Base

Targets **`v2/fsdp2-parallel`** (the new parallelism trainer this config uses, `trainer.parallelism`). That branch merges to `main` first; retarget this PR to `main` after it lands.

## Verification

Trains on a single V100 (Casper) on the fsdp2-parallel base with all four fixers active: `train_loss` → 1.32, `valid_loss` → 1.11, finite, no NaN.

## Related issues
- #428 — gen2 trainer should hand postblocks 5D `(B,C,T,H,W)` (the shim works around the current 4D flatten).
- #429 — gen2 postblocks need first-class access to input-only forcings (the `TOA_down_solar_from_input` flag works around it).

Issue #429, opened 2026-06-19
gen2 postblocks can't access input-only dynamic forcings (e.g. SOLIN) — blocks GlobalEnergyFixerUpDown


jsschreck:
## Summary

The gen2 postblock schema gives postblocks no clean, model-agnostic way to address an **input-only dynamic forcing** channel. `GlobalEnergyFixerUpDown` needs incoming TOA down-solar (SOLIN / DSWRFtoa), which in gen2 is a prescribed forcing in the *input*, not a predicted *output*.

## Where

`credit/postblock/gen1.py:1012`
```python
TOA_down_solar = y_pred[:, self.TOA_down_solar_ind, 0, ...]
R_T = (TOA_down_solar - TOA_up_solar - TOA_up_OLR) / self.N_seconds
```
The fixer expects TOA down-solar to live in `y_pred`.

## Problem

In the gen2 data structure SOLIN is a `dynamic_forcing` input-only variable — prescribed, not predicted — so it is absent from `y_pred`. It is present in `x["x"]` (CAMulator: channel 132), but the gen2 schema exposes only output channels to postblocks in a uniform way; input-only forcing channels aren't mapped for postblock use. As a result `global_energy_fixer_updown` cannot be enabled for CAMulator gen2 (the up/down energy balance needs TOA down-solar). Deferred on branch `camulator_v2`; `tracer`/`mass`/`water` fixers work.

## Workaround (follow-up PR, linked)

We will add a config-gated path so `GlobalEnergyFixerUpDown` sources TOA down-solar from the input tensor (`x["x"][:, <solin_ind>, ...]`) when a flag is set, and enable it in the CAMulator gen2 config. That PR will link this issue.

## Ask

@charlie-becker — longer term, please provide a first-class way for postblocks to address input-only dynamic forcings (e.g. a forcing-channel map in `full_data_dict`, or a documented convention), so fixers don't hard-code input indices. Also note: `credit convert` does not yet emit any conservation fixers into the gen2 postblock schema.

jsschreck:
Workaround in PR #431 (config-gated `TOA_down_solar_from_input` sources SOLIN from the input tensor). This issue tracks the longer-term fix: first-class postblock access to input-only dynamic forcings.

djgagne:
[...]

Issue #428, opened 2026-06-19
gen2 trainer flattens 5D y_pred to 4D before postblocks — conservation fixers expect (B, C, T, H, W)


jsschreck:
## Summary

The gen2 trainer collapses the time dimension of `y_pred` **before** running postblocks, but the gen1 conservation fixers are written for a 5D `(B, C, T, H, W)` tensor. The hand-off to postblocks should stay 5D `(B, C, T, H, W)`.

## Where

`credit/trainers/trainer_gen2.py:167`
```python
full_data_dict["y_pred"] = full_data_dict["y_pred"].flatten(1, 2)
```
This runs before `apply_postblocks(self.step_postblocks, ...)` (`trainer_gen2.py:307`, also `:332`, `:483`, `:503`) and merges channel and time into `(B, C*T, H, W)`.

## Problem

`GlobalMassFixer`, `GlobalWaterFixer`, and `GlobalEnergyFixer*` in `credit/postblock/gen1.py` index time explicitly, e.g. `y_pred[:, inds, 0, ...]` and `x_input[:, inds, -1, ...]`. Given a flattened 4D tensor they index spatial dims instead of time, producing wrong physics / shape errors. CAMulator is the first gen2 model to wire conservation fixers, so this surfaced now (see branch `camulator_v2`).

## Workaround in place (camulator_v2)

Backward-compatible dummy-time-dim shim added to `GlobalMassFixer.forward` (`gen1.py:302`) and `GlobalWaterFixer.forward` (`gen1.py:522`): unsqueeze a time dim when a 4D tensor arrives, squeeze it back before returning. No-op on 5D, no physics changed. This is only correct for single-step (`T=1`); for multi-step (`T>1`) `flatten(1, 2)` merges `C*T` and the shim is **not** a correct substitute.

## Ask

@charlie-becker — restore the 5D `(B, C, T, H, W)` hand-off to postblocks (don't flatten time into channels before `apply_postblocks`). Once that's in, the `gen1.py` shim on `camulator_v2` can be removed. Flagging since this is your postblock/trainer area and we're working around it for now.

jsschreck:
Workaround in PR #430 (dummy-time-dim shim in the mass/water fixers). This issue tracks the proper fix: restore the 5D (B, C, T, H, W) hand-off in the gen2 trainer so the shim can be removed.

charlie-becker:
Sounds good, I'm happy to keep everything in 5D format. Will implement ASAP.

[...]

PR #431, opened 2026-06-19
CAMulator: enable global_energy_fixer_updown (input-sourced TOA solar)

jsschreck:
## What

Enables the fourth conservation fixer, `global_energy_fixer_updown`, for CAMulator gen2:
- Registers the existing `GlobalEnergyFixerUpDown` class in `POSTBLOCK_REGISTRY`.
- Adds a config-gated `TOA_down_solar_from_input` flag so the fixer sources TOA down-solar (SOLIN) from the **input** tensor `x["x"]` instead of `y_pred`.
- Enables the fixer in the CAMulator gen2 config.

## Why

In gen2, SOLIN is a `dynamic_forcing` **input-only** variable (input ch 132) and is absent from `y_pred`, but `GlobalEnergyFixerUpDown.forward` reads TOA down-solar from `y_pred`. The flag (default **off** → byte-identical to existing behavior) lets it read the prescribed forcing from the input. This is the workaround for the schema limitation tracked in #429.

Also applies the same backward-compatible dummy-time-dim shim (the 4D gen2 hand-off, see #428) to this fixer.

## Verification

Trains on a single V100 with **all four** conservation fixers (tracer/mass/water/energy_updown) active:
- `train_loss` → 1.32, `valid_loss` → 1.11, finite, no NaN
- energy fixer fires every batch with finite `R_T` / `F_S`

## Stacking

⚠️ This PR is **stacked on #430** (base = `camulator_v2`) so the diff shows only the energy delta. After PR1 merges, retarget the base to `v2/ensemble-gen2`.

Relates to #429 (longer-term: first-class postblock access to input forcings — @charlie-becker).


jsschreck:
Consolidated into a single PR #430 (one branch camulator_v2 → main). Closing.

Issue #432, opened 2026-06-23
Gen2 Trainer (and other derivative functions) do not support `levels: null` for selecting all levels, despite LocalDataset supporting it


lj-dunphy:
# Gen2 Trainer (and other derivative functions) do not support `levels: null` for selecting all levels, despite LocalDataset supporting it

When setting the levels param to null in the config, such as the following:

```yaml
data:
  source:
    CAM6:
      dataset_type: "local"
      level_coord: "lev"
      levels: null
```

The `LocalDataset` correctly reads it as selecting all available layers in this snippit of code:
https://github.com/NCAR/miles-credit/blob/595cf700f1fb415d035263687f032f92c153d327/credit/datasets/local.py#L181-L186

However, when passed through the Gen2 Trainer, it throws an error here due to the param `levels` still existing (even though it is of value null), causing the default value to not be invoked, then attempting to get the length of a NoneType object:

https://github.com/NCAR/miles-credit/blob/595cf700f1fb415d035263687f032f92c153d327/credit/trainers/trainer_gen2.py#L57 

If you instead pass `levels: []`, `LocalDataset` will interpret it as wanting 0 levels, as `[]` is not `None`, and the Gen2 Trainer also interprets it as 0 levels, meaning it won't crash, however it will think there are no vertical levels (which will cause crashes later in the pipeline)

Finally, if you instead completely omit the `levels:` param from the config, `LocalDataset` will correctly load in all the vertical levels; however, the trainer again defaults to the `[]` value, making it think that there are 0 vertical levels (as shown below), and that incorrect value is passed on to many other functions, leading to buggy behavior:

https://github.com/NCAR/miles-credit/blob/95861e7f6343cdc63df4a3c09b8b723b1d8d9da9/credit/trainers/trainer_gen2.py#L58

### `preflight.py` (see lines 62-66, 84-85):
If `levels: null`, memory estimate will be 0 (silent fail). If `levels: []` or it is omitted, then memory will be estimated as if there were no vertical levels.
https://github.com/NCAR/miles-credit/blob/95861e7f6343cdc63df4a3c09b8b723b1d8d9da9/credit/trainers/preflight.py#L53-L85

### `channel_layout.py`:
If `levels: null`, `len(None)` returns a `TypeError`. If `levels: []`, all the channel slices will be empty.  If `levels` is omitted, a `KeyError` is returned.
[...]

PR #433, opened 2026-06-23
Small bug fixes

### LINKED ISSUE #422: Should scaler type be bridgescaler_transform or bridgescaler_transformer?

charlie-becker:
### Description
This address two bug related issues.

1. The `level_coord:` now uses the "nearest" method that will be useful for floating point level (CESM) that a user wants to subset. `levels:` in the data config now properly defaults to use all available levels if set to `null` or is absent all together. 

2. The configs and code have been modified to consistently use the `bridgescaler_transform` registry key instead of `bridgescaler_transformer` key, where there was inconsistencies.  
 
Closes #422 and #432

### Checklist

### DIFF
--- a/credit/trainers/trainer_gen2.py
+++ b/credit/trainers/trainer_gen2.py
@@ -54,7 +54,7 @@ class TrainerERA5Gen2(BaseTrainer):
         source = next(iter(data_conf["source"].values()))
         vars_conf = source["variables"]
         diag = vars_conf.get("diagnostic") or {}
-        num_levels = len(source.get("levels", []))
+        num_levels = len(source.get("levels") or [])
         self.varnum_diag = (len(diag.get("vars_3D", [])) * num_levels + len(diag.get("vars_2D", []))) if diag else 0
 
         self.retain_graph = data_conf.get("retain_graph", False)
--- a/config/gen_2/examples/weatherbench2_era5_wxformer.yml
+++ b/config/gen_2/examples/weatherbench2_era5_wxformer.yml
@@ -44,7 +44,7 @@ preblocks:
           - 'input'
           - 'target'
     scaler:
-      type: "bridgescaler_transformer"
+      type: "bridgescaler_transform"
       args:
         variables: []
         scaler_path: "/glade/derecho/scratch/$USER/CREDIT/weatherbench2/scaler_quantile.json"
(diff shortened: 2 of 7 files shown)


kevinyang-cky (approved):
LGTM on my end. One note: when distances are tied, ds.isel(method="nearest") returns the larger value, though this is probably rare.

credit/trainers/trainer_gen2.py
```diff
@@ -54,7 +54,7 @@ def __init__(self, model: torch.nn.Module, rank: int, conf: dict):
         source = next(iter(data_conf["source"].values()))
         vars_conf = source["variables"]
         diag = vars_conf.get("diagnostic") or {}
-        num_levels = len(source.get("levels", []))
+        num_levels = len(source.get("levels") or [])
```
lj-dunphy, line 57: I don't think that this fixes the issue, although the crash is now avoided, downstream `num_levels` will now be 0, which is still incorrect if the user wants to select all the levels by either omitting the parameter or passing None/null

charlie-becker: Ah, I think you're right. I thought it was pulling from the dataset not the config there. Apologies for the premature merge -- I'll fix this up and send it your way.

lj-dunphy: If you want to pull some number from the config for `num_levels`, cfg['models']['levels'] has an explicit number, however I know you mentioned earlier that you guys would rather derive some of those params rather than state them in the config

lj-dunphy: Or you could use `dataset.static_metadata["levels"]`, however I don't know totally if the Dataset is initalized at all points where we grab the levels or the num_levels (see [the issue](https://github.com/NCAR/miles-credit/issues/432) for all the points where we do, as from my understanding, we would need to fix this at all of those points)

PR #435, opened 2026-06-24
Remove dead code

charlie-becker:
### Description

@lj-dunphy (regarding #433)

It looks like that block was all dead code from gen 1. The diagnostics are cleanly handled with the preblock / postblock / data schema

I removed the dead code, and it should now succesfully operate with `levels:` as `null`, a subset list, or excluded all together which will default to all variables.


lj-dunphy:
Are all the rest of the places where we grab the `levels` param also dead code then? Like for example, this preflight code where we grab `len(src.get("levels", []))`:

https://github.com/NCAR/miles-credit/blob/95861e7f6343cdc63df4a3c09b8b723b1d8d9da9/credit/trainers/preflight.py#L53-L85

PR #437, opened 2026-06-25
Allow validation_data to be fully independent from training data config

kevinyang-cky:
### Description
It is natural for `validation_data` to share the same data source as the training data. However, keys under source could differ between training and validation, and the original design intended `validation_data` to override training settings — but source was always forced back to the training value.

Previously, the validation config was built as:

https://github.com/NCAR/miles-credit/blob/6a6e5f44fe2c945cf263368b3ee3f64340a826dd/credit/trainers/utils.py#L449-L450

The first line merges the training config with `validation_data`, so any key not specified in `validation_data` would silently inherit the training value. The second line then unconditionally overwrites the source with the training value, ignoring anything the user set in `validation_data`.

This PR makes `validation_data` fully independent when present: it is used as-is, with source inherited from training only as a convenience when omitted. Incomplete configs raise ValueError with the missing keys, making misconfiguration explicit rather than silent.

### Checklist

### DIFF
--- a/credit/trainers/utils.py
+++ b/credit/trainers/utils.py
@@ -440,8 +440,8 @@ def inject_postblock_info(conf: dict) -> None:
             cfg_ud["sp_inds"] = _find_ind("surface_pressure_name", single=True)
 
 
-def _resolve_data_conf(conf: dict, is_train: bool) -> dict:
-    """Resolve the data config for train or validation.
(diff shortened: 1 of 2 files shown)


jsschreck (approved):
Looks good thanks for enabling this!

PR #439, opened 2026-06-26
Wire up history_len > 1 for Gen2 training

### LINKED ISSUE #408: gen2: support for multi-step input history (`history_len > 1`)
## Summary

The Gen2 training pipeline currently does not support multi-step input history
(`history_len > 1`), despite the config schema accepting the key and the model
backbone (e.g., `crossformer.frames`) supporting it. The value silently has no
effect on training.

## Reproduce

Set `data.history_len: 2` in any Gen2 config and train. The loss curve is
identical to `history_len: 1`. Setting `model.frames: 2` even runs without
error, because the dataset still feeds a single timestep regardless.

## Where the gap is

The Gen2 training loop reaches the dataset through this call chain:

- `credit/applications/train_gen2.py` calls `load_dataset()` from
  `trainers/utils.py`, which builds a `MultiSourceDataset`.
- `MultiSourceDataset` instantiates one sub-dataset per source key via
  `route_to_dataset_class`. For `dataset_type: "local"`, the sub-dataset is a
  `LocalDataset` (subclass of `BaseDataset`).
- For each dataloader index `(t, i)`, `MultiSourceDataset.__getitem__`
  forwards it to each sub-dataset's `__getitem__`, which calls
  `BaseDataset._load_sample(t, i)`.
- `_load_sample` opens the source file(s) and reads exactly one timestep at
  `t` for each field, regardless of `history_len`.

Grepping `history_len` across the Gen2 path, the only hit is in
`trainerERA5gen2.py`, which assigns `self.valid_history_len` and never reads
it again -- a dead store. No other file (`base_dataset.py`, `local.py`,
`multi_source.py`, `channel_layout.py`, `samplers.py`, `trainers/utils.py`)
references it at all.

This contrasts with Gen1, where `history_len > 1` is wired end-to-end via
`era5_multistep_batcher` and `trainerERA5gen1`.

## Why it matters

Having the key in the config but not wired up is a footgun: users may believe
they are training with history when they aren't, and downstream comparisons
against history-aware baselines become misleading.

## Proposed changes

Working implementation on a branch: https://github.com/NCAR/miles-credit/tree/rwrf_run

**Note**: this branch also contains experimental work on multi-step output
(`output_frames > 1`), which I'll file as a separate issue. The `history_len`
changes are confined to the four files listed below; happy to split the
branch into a clean PR if there's interest in merging.

Design summary:

- Keep the existing Gen2 contract -- the sampler yields `(t, i)` and the
  dataset handles how much data each index maps to. No sampler changes.
- Split decision from execution: `_load_sample` (in `BaseDataset`, shared
  across source types) decides *which* timesteps to read and assembles the
  history window. `_extract_field` (per source type, e.g. `LocalDataset`)
  just reads what it's told. Future source types only need to implement
  multi-step reading in their own `_extract_field`; they don't have to
  re-derive the history logic.
- Touched files: `base_dataset.py`, `local.py`, `channel_layout.py`,
  `trainerERA5gen2.py`. All have a fast path for `history_len == 1`.

See the branch for the full diff (~150 LOC for history_len).

## Backwards compatibility

`history_len == 1` (or unset) is a fast path with identical behaviour to
current main. The only interface change is `_extract_field` gaining a
keyword-only `t_history` parameter; subclasses that don't opt in can ignore
it. `crossformer.frames` must match `data.history_len` (same as current main).

## Out of scope / known limitations

- `credit/applications/rollout_to_netcdf_gen2.py` (inference) is not modified.
  When using `history_len > 1` at inference, the same `update_x` change needs
  to be applied there; happy to do that in a follow-up.
- Only `LocalDataset` is updated. Other source types inherit the new
  `_extract_field` signature (so calls won't break) but fall back to the
  single-step path until they opt in.

## Validation

Trained `history_len: 2, output_frames: 1, forecast_len: 1` for 5 epochs on
0.25-degree ERA5. Training is stable; single-step valid_loss tracks the
`history_len: 1` baseline within a few percent across all 5 epochs.

## Question for maintainers

Would you accept a PR along these lines, or do you have a different design in
mind for multi-step input in Gen2 (for example, exposing it via the sampler
instead of the dataset)? Happy to revise before submitting.

ILoveSorasakiHina:
### Description

Wire up `history_len > 1` for Gen2 training so that the config key is actually
used to build multi-step input windows, instead of being silently ignored.

Previously the Gen2 path accepted `history_len` in the config but never acted on
it — the only trace was a dead `self.valid_history_len` store — so every run
trained as 1-in-1-out regardless of the configured value. This PR makes the
sampler, dataset, rollout assembly, and trainer build and consume a real
history window, while leaving `history_len == 1` on its existing fast path
(no behavioural change).

**Changes by file:**

- **`credit/datasets/base_dataset.py`** — Reads `history_len` (optional key,
  defaults to 1, validated as a positive int; a per-source value overrides the
  data-level default). When `history_len > 1`, `_load_sample` builds a
  chronological history window `t_history = [t-(H-1)·dt, …, t]` and passes it to
  `_extract_field` for the field types that carry history (prognostic, the newest
  dynamic_forcing, static). The target stays single-step at `t+dt`. The timestamp
  range is shifted forward by `(H-1)·dt` so every sample has enough history.

- **`credit/datasets/local.py`** — `_extract_field` gains a keyword-only
  `t_history`. When `None`, behaviour is identical to the original single-step
  read. When provided, each timestamp in the window is loaded and stacked along
  the time dimension, grouping timestamps by file so each file is opened at most
  once. 3D variables become `(n_levels, n_t, lat, lon)`; 2D (surface) variables
  become `(1, n_t, lat, lon)`.

- **`credit/trainers/rollout_utils.py`** — `assemble_rollout_batch` takes a
  `history_len`; when `> 1`, carried-forward (static / non-predicted) fields are sliced to 
   their newest step so they don't accumulate stale history. The training target is
  forwarded so preblocks normalize it in the same pass.

- **`credit/trainers/trainer_gen2.py`** — Reads `self.history_len` and slides the
  history window during rollout (drop oldest, append newest along the time dim),
  maintaining a separate un-ensembled history tensor.

- **`credit/preblock/concat.py`** — Fixes a latent bug in `ConcatToTensor` that
  was first triggered by enabling `history_len > 1`. The output channel map was
  built inside the input loop using the *input* tensor's time dim, but the model
  predicts a single step (`forecast_len == 1`), so `Reconstruct` then tried to
  unflatten a single-step `y_pred` with the input's `T`. With `history_len == 1`
  the input and output time dims coincide (both 1), so the bug was masked; with
  `history_len > 1` it surfaced as an unflatten shape error in the postblock
  chain. The fix builds the output map from the **target** tensor's own shape
  (the target is the true output step count), keeping the input map unchanged.
  Verified that `history_len == 1` produces an identical output map to before
  (zero regression).

**Validation:** Both scenarios were trained end-to-end on real data (single
source, 8×A100 DDP) and converge cleanly with no NaNs or divergence:

- **1-in-1-out (`history_len: 1`)** — confirms the fast path is unchanged.
- **2-in-1-out (`history_len: 2`)** — first real exercise of the multi-step
  path; train/valid loss decrease monotonically and validation accuracy climbs
  steadily across epochs.

Existing dataset test suites (base-dataset, multi-source, ERA5) pass.

Note: `history_len > 1` is currently exercised with single-source
(`LocalDataset`) training. Other source types do not yet accept the multi-step
read path; that is tracked separately and will be handled before launch.
`history_len == 1` is unaffected and works with any source combination.

### Related Issues and PRs

Closes #408 

### Checklist
- [x] I am familiar with the [[contributing guidelines](https://miles-credit.readthedocs.io/en/latest/contrib.html)](https://miles-credit.readthedocs.io/en/latest/contrib.html).


ILoveSorasakiHina:
pre-commit.ci autofix

kevinyang-cky:
@ILoveSorasakiHina It looks like Pandas' [latest release](https://pandas.pydata.org/docs/whatsnew/v3.0.4.html) from yesterday (v3.0.4) is causing the issue. Constraining Pandas to <3.0.4 for now should let the CI pass.

djgagne:
I pushed the pandas version constraint, and it looks like that is letting tests pass now.

djgagne:
@ILoveSorasakiHina is there a way to move the history length selection logic from LocalDataset to MultiSourceDataset so that all datasets can benefit from without having to reimplement the logic each time?

ILoveSorasakiHina:
@djgagne 

Thanks, that's a good suggestion — I'm happy to take this on. It should be doable technically; I'll just need a bit of time to work through it properly.
One quick question that decides where the logic should live: right now all gen2 datasets go through MultiSourceDataset, but is that guaranteed to stay the case going forward? If it'll always be the entry point, MultiSourceDataset works well for this. If a dataset might ever be used outside it down the line, I'd lean towards putting the shared logic in BaseDataset instead. Happy to go whichever way fits the intended direction.

charlie-becker:
> One quick question that decides where the logic should live: right now all gen2 datasets go through MultiSourceDataset, but is that guaranteed to stay the case going forward? If it'll always be the entry point, MultiSourceDataset works well for this. If a dataset might ever be used outside it down the line, I'd lean towards putting the shared logic in BaseDataset instead. Happy to go whichever way fits the intended direction.

Thanks for working on this. I believe we always intend to flow through MultiSourceDataset (even if there is a single source) so using wiring it there would work. However, there might be uses of the individual dataset classes (downloading data for example), but with a training or inference run it should always flow through MultiSource.

ILoveSorasakiHina:
Thanks @djgagne @charlie-becker. I gave this a go — rather than moving the logic to MultiSourceDataset, I made the multi-step read generic in BaseDataset, and I wanted to check the direction with you.

1. Why BaseDataset rather than MultiSourceDataset.
MultiSourceDataset only delegates to each sub-dataset and merges the results — it never reads files itself, and doesn't hold dt / var_dict / _extract_field. Driving the history window from there would mean either reaching into the sub-datasets or re-implementing _load_sample's field/target semantics one layer up. In short, history_step reading is a "how one source reads its own timesteps" problem, so it belongs where the reading happens (BaseDataset), not where sources are merged (MultiSourceDataset). Keeping it in BaseDataset: individual dataset classes still work standalone, and every source gets history_len > 1 for free. Concretely, I added BaseDataset._extract_field_window(...), which calls the subclass's existing single-step _extract_field once per timestamp and stacks along time — so era5 / hrrr / goes / mrms / tisr need no changes to support multi-step.


2. Why LocalDataset gets its own override.
The generic version re-opens the underlying file once per history step. For LocalDataset a single yearly file covers the whole window, so I overrode _extract_field_window to group the window by file and open each file once (identical output, fewer opens). tisr doesn't need this — it's pure computation, no I/O.

For the remaining sources (era5 / hrrr / goes / mrms), whether the per-step re-open matters depends on how each one actually reads. I've read through them but haven't run them myself, so I'd rather not assert — could whoever works with each confirm whether the extra opens are a real concern, or acceptable as-is? If any needs it, the same override pattern as LocalDataset applies.

Does this direction look right?

kevinyang-cky:
GOES Dataset pulls one timestep per file GOES Level 2 Cloud and Moisture Imagery (CMI) product from AWS, so the generic version should work just fine. If you have `BaseDataset._extract_field_window` pushed to a branch, I can take a quick peek. Thanks!

ILoveSorasakiHina:
pre-commit.ci autofix

ILoveSorasakiHina:
Thanks @kevinyang-cky 

that settles GOES, no override needed there. I've pushed my changes to the history_len branch. Summary (as above): base_dataset.py gains _extract_field_window, and LocalDataset overrides it to keep its open-once file optimization. Whenever you all have time, a code review would be much appreciated. Thanks again!

kevinyang-cky:
@ILoveSorasakiHina Thanks for making the changes!

I ran `preprocess` (preprocess.py), `train` (train_gen2.py), and `rollout` (rollout_gen2.py) using [example-end-to-end.yml](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/examples/example-end-to-end.yml). `preprocess` and `train` ran fine, but `rollout` failed.

The two commits that I made: 

- 1st: removed unused keys and added comments explaining how to use `history_len`. [example-end-to-end.yml](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/examples/example-end-to-end.yml) is currently the go-to end-to-end example (preprocess, train, and rollout), and I encourage you to use it to test if any of the pipeline failed when new changes are made.

- 2nd: `history_len` needs to be considered in the multisource's master clock.

`Rollout` failed due to an issue in `concat.py:150-162`, in the `ConcatToTensor` fallback path used when the target is not present (e.g., during rollout, `return_target=False`). It falls back to reading the time dimension from the input tensor rather than the target's — but the input's time axis is `history_len` (2 in this config), not the model's true per-step output width (always 1). That inflates the expected output channel width per variable by 2×, causing a downstream unflatten crash. Training never hits this fallback path (it always has a real target), so the bug surfaces only during rollout and only when `history_len > 1`. Let me know if it makes sense.

If you can address this issue and the merge conflicts, that would be greatly appreciated. Thanks.

ILoveSorasakiHina:
Thanks @kevinyang-cky, and thanks for the two commits!

You spotted a critical bug, and I’ve just implemented a fix for it. Since the inference/rollout output should always have a time dimension of 1 for now, I’ve forced it to 1 when no target is present. (If we support multi-step output in the future, I'll update this logic accordingly). The training path remains unchanged as it still relies on the actual target shape.

I’ve also resolved the merge conflicts. I just finished testing both preprocess and train using WRF data on the CWA machine, and everything ran fine. I’ll test the rollout part tomorrow.

ILoveSorasakiHina:
pre-commit.ci autofix

ILoveSorasakiHina:
@kevinyang-cky 

The previous inference code didn't support a history length greater than 1, so I updated rollout_utils.py to resolve the issue, and the pipeline now runs smoothly.I've successfully verified the entire workflow (preprocess $\rightarrow$ train $\rightarrow$ rollout) with WRF data on the CWA HPC. 

Please take a look and let me know if you have any feedback. Thanks!

kevinyang-cky (approved):
@ILoveSorasakiHina Thanks again for your work implementing this important piece for Gen2!

I made a few follow-up commits, mostly minor improvements:

* Renamed the `multistep` flag to a more intuitive name derived from `history_len`, since "multistep" in CREDIT is primarily associated with autoregressive rollout.
* Avoided loading static data multiple times.
* Log `forecast_len` to `training_log.csv`.

I also tested the following workflows using `history_len: 2` in `example-end-to-end.yml`:

* Preprocessing
* Training:
  * 1-in-1-out single-step
  * 2-in-1-out single-step
  * 2-in-1-out multi-step
* Rollout

Everything looks good (e.g., reasonable loss behavior) from my testing.

A couple of notes for future reference:

* Based on recent discussion, the multi-output implementation can be deferred to the 2026.2.1 release (most likely in the fall) or later.
* In `concat.py`, the fallback logic for the target channel map when a batch has no target (e.g., during rollout) could be revisited in the future (seems to be more of a workaround method).

Merging now!

Issue #401, opened 2026-05-27
Update Rollout for Gen2


charlie-becker:
The Gen 2 trainer (`trainers/trainer_gen2.py`) has been updated with #374 to use a new dictionary to hold all data and to integrate the new preblock and postblock schemes.

kevinyang-cky:
Close with #440 merged.

PR #440, opened 2026-06-26
Rollout updates

### LINKED ISSUE #349: Config Organization Ideas
The current config could be adjusted to be a little more readable, while still only using a single config (though we could also potentially split out the config: train / inference / etc).

`save_loc:` Everything should pipe to this base directory and make the directory if it doesn't exist. This should have sub-directories such as: `logs`, `plots`, `models`, `config`, etc

```
trainer:
    type: era5-gen2
    mode: none
```
The `type` should no longer be necessary! Everything will always go through `MultiSourceDataSet()`...

`predict:` This section needs the largest overhaul IMO. 

1.    
 ```
forecasts:
          type: "custom"       # keep it as "custom". See also credit.forecast
          start_year: 2019     # year of the first initialization (where rollout will start)
          start_month: 1       # month of the first initialization
          start_day: 1         # day of the first initialization
          start_hours: [0] # hour-of-day for each initialization, 0 for 00Z, 12 for 12Z
          duration: 3       # number of days to initialize from, starting from the (year, mon, day) above
```
This should be replaced with the following three arguments (where all others can be derived from if we want to keep the behind the hood logic the same -- but there is probably optimizations to be had there).
```
forecast_start: <pandas interpretable datetime string>
forecast_end: <pandas interpretable datetime string>
time_step: <pandas interpretable interval (i.e. "6h")>
```
`save_foreacst:` Let's remove this and default save to `save_loc:` as mentioned above

```
        ic_interval_days: 1   # spacing between init days. defaults to 1 total number of ICs will be duration // ic_interval_days
        days: 2             # forecast lead time as days (1 means 24-hour forecast)
```
I don't see the need to keep these either.

Much of the other stuff in the current predict will also be removed as it will be incorporated into postblocks (interpolation, conservation, metadata, etc.) but some of this will have to wait until after postblocks is completed.

We will also need a flag for "realtime" (bool?) and then a flag for which data to initialize from.

I would prefer to rename `predict` to `inference` but fine if other disagree. Lastly, I also thin it would make more sense to move the `predict:` block more towards the top of the config (beneath data perhaps).

### PR DESCRIPTION
**Pull request by charlie-becker**
### Gen2 Rollout Pipeline

This PR adds a complete gen2 inference pipeline (`rollout_gen2.py` + `output_gen2.py`), replaces the old `predict:` config block with a new `inference:` block, and updates the trainer registry key from `era5-gen2` to `gen2`. 
  
  ### New: `credit/applications/rollout_gen2.py`

  A combined batch + single-forecast rollout that mirrors the gen2 training inner loop
  exactly — the same `full_data_dict` super-dict flows through preblocks → model →
  postblocks at every step. No manual denormalization or flat-tensor surgery.

  **Run modes** (selected via `inference.run_mode` or `--run-mode` CLI flag):
  - `batch` — hindcast over a range of init times defined in `inference.batch_forecast`
  - `single` — one forecast from a specific init time in `inference.single_forecast`

  **Distributed inference** (`inference.mode`):
  - `none` — single GPU
  - `ddp` — multi-GPU with `torchrun`
  - `fsdp` — sharded model for very large checkpoints

  > **Note:** tensor and domain parallelism are training-only features and are not used
  > during inference.

  ---
  
  ### New: `credit/output_gen2.py` — `ForecastWriter`

  A stateful, buffered NetCDF writer. Called once per forecast step; handles temporal
  subsampling, file grouping, variable/level filtering, CF metadata, and async writes
  via a multiprocessing pool.

  | Option | Description |
  |--------|-------------|
  | `output_freq` | Only write steps whose forecast hour is a multiple of this value (`null` = every step) |
  | `group_by` | `null` = one file per step; `"day"` = one file per day window; `"full"` = one file per forecast |
  | `variables` | Subset by full variable key and optionally by pressure level |
  | `metadata` | Inject CF attributes from a YAML file (e.g. `era5.yaml`) |
  | `encoding` | Per-variable NetCDF compression with a default that applies to all variables |

  ---

  ### New `inference:` config block (replaces `predict:`)

  The old `predict:` block has been replaced in all `config/gen_2/` files. The new
  block separates run-mode selection, initialization windows, and output options cleanly according to #349

  ```yaml
    inference:
      run_mode: batch               # batch | single
      mode: none                    # none | ddp | fsdp2
      save_forecast: '/glade/derecho/scratch/$USER/CREDIT_runs/my_run/rollout'
      num_save_workers: 4           # async write pool size (0 = synchronous)
  
      batch_forecast:               # active when run_mode: batch
        length: "10d"               # forecast duration per init (e.g. "10d", "1m", "1y")
        initialization:
          start_date: "2020-01-01"  # first init date (inclusive), ISO 8601
          end_date: "2020-12-31"    # last init date (inclusive)
          date_interval: "2d"       # step between init dates
          times: ["00:00", "12:00"] # UTC init times each day (HH:MM)
  
      single_forecast:              # active when run_mode: single
        length: "10d"               # forecast duration
        start_datetime: "2020-06-01T00:00Z"   # init time, ISO 8601
  
      output:
        format: netcdf              # netcdf | zarr
        output_freq: null           # write every Nth step; null = every step (e.g. "6h", "24h")
        group_by: null              # null (per-step) | "day" | "month" | "year" | "full"
        variables: null             # null = save all; or list of {name:, levels:} to subset
        metadata: "era5.yaml"       # CF attribute YAML (null = no variable attributes)
        encoding:
          default:                  # applies to all variables
            dtype: float32
            zlib: true
            complevel: 4            # 1 (fast) to 9 (small)
      variables:
        - name: "ERA5/prognostic/3d/temperature"
          levels: [500, 850, 1000]   # subset by model-level value; null = all levels
        - name: "ERA5/prognostic/2d/SP"
        - name: "ERA5/diagnostic/2d/total_precipitation"
```
  ---
  Other changes

  - Trainer registry: "era5-gen2" renamed to "gen2"; old key kept as a
  backward-compat alias.
  - credit/cli/_submit.py: credit rollout now dispatches to rollout_gen2.py
  with the correct CLI flags.
  - credit/applications/preprocess.py: Fixed KeyError: 'mode' when using gen2
  configs that have trainer.parallelism instead of trainer.mode.
  - credit/samplers.py: Fixed IndexError on single-init rollouts caused by eager
  evaluation of getattr default.
  - credit/output_gen2.py: Fixed ValueError: different number of dimensions for
  2D variables in _to_dataset.
  - tests/test_output_gen2.py: 47 new unit tests covering all ForecastWriter
  methods (shape correctness, path format, buffering, filtering, encoding, coordinates,
  level subsetting).

I tested this entire pipeline sequentially using:

1. `credit preprocess -c config/gen_2/examples/example-v2026.3yml`
2. `credit train -c config/gen_2/examples/example-v2026.3yml`
3. `credit rollout -c config/gen_2/examples/example-v2026.3yml`

Appreciate if others would try this. 

Closes #349 #401 

### Checklist


charlie-becker:
I forgot to mention. Due to the complexity of GFS and GEFS inegration, I choose to leave that out for now to make it easier to review. I'm working on converting GFS and GEFS to torch datasets to integrate with this as well as a separate PR and am making it high priority.

kevinyang-cky:
Some updates on my initial review: (will submit a more thorough one later) 
`credit [preprocess/train/rollout] -c config/gen_2/examples/example-v2026.3.yml` all ran successfully. I made two commits to fix:
1) Multi-preprocessing warning messages (should be all gone)
2) Set the `channels_last` flag to `False` in the scaling preblock

While going through preprocess, train, and rollout, I also cleaned up the logging messages (to improve user experience), but I am still deciding whether to push the changes to this PR or open another PR if it fits better.

config/gen_2/examples/example-v2026.2.yml
kevinyang-cky, line 1: Also ran training with this config successfully. The scaling preblock is missing the "scaler_type" and "scaler_params" keys, so it wouldn't be appropriate to run preprocess with it (maybe good to leave a note at the beginning of the config or just wait until later when we focus on documentation).

config/gen_2/examples/example-end-to-end.yml
kevinyang-cky, line 1: Suggest renaming this config to something that reflects it is an end-to-end (preprocess/train/rollout) example and should be the go-to for anyone who wants to learn how to prepare a config.

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+# CLI overrides (all optional):
+#   --run-mode batch|single     overrides run_mode
+#   --init-time YYYY-MM-DDTHH  overrides single_forecast.start_datetime
+#   --save-dir /path            overrides save_forecast
+#
+inference:
+  run_mode: single                   # batch | single
+  mode: none                        # none | ddp | fsdp
+  save_forecast: '/glade/derecho/scratch/$USER/CREDIT_runs/starter_gen2/rollout'
+  num_save_workers: 4               # CPU workers for async NetCDF save pool
```
kevinyang-cky: 1) Suggest using "thread_workers" to be consistent with the trainer block
2) `rollout_gen2.py` seems not to be using this key at all. See https://github.com/NCAR/miles-credit/pull/440/changes#diff-e51f673c1d43d1fb93580a1c202ef138684f952b082a5b26eec3265f0ffb79b5R98-R99 and https://github.com/NCAR/miles-credit/pull/440/changes#diff-e51f673c1d43d1fb93580a1c202ef138684f952b082a5b26eec3265f0ffb79b5R237

config/gen_2/examples/example-end-to-end.yml
```diff
@@ -0,0 +1,362 @@
...
+# run_mode selects between batch and single forecast without requiring CLI flags.
+#   batch:  generates forecasts for a range of init times defined under batch_forecast.initialization
+#   single: runs one forecast from a specific init time defined under single_forecast.start_datetime
+#
+# CLI overrides (all optional):
+#   --run-mode batch|single     overrides run_mode
+#   --init-time YYYY-MM-DDTHH  overrides single_forecast.start_datetime
+#   --save-dir /path            overrides save_forecast
+#
+inference:
```
kevinyang-cky, line 282: I can definitely see the benefit of having separate modes for single and batch forecasts. That said, I was wondering whether having a separate mode might also introduce some additional configuration keys.

Would it make sense for a single forecast to simply fall under the batch forecast mode instead? For example, a single forecast could be represented by setting the start and end dates to the same value and specifying a single forecast time. To me, that feels a bit more conceptually intuitive, but I'm curious what your thoughts are, or whether there are use cases that make keeping the two modes separate preferable.

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+
+  batch_forecast:
+    length: "10d"                   # forecast duration; parsed relative to data.timestep
+    initialization:
```
kevinyang-cky: Would suggest removing "initialization" and just having all the keys under batch_forecast.

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+      start_date: "2020-01-01"      # first init time (inclusive); must be within data date range
+      end_date: "2020-12-31"        # last init time (inclusive)
+      date_interval: "2d"           # interval between init times
+      times: ["00:00", "12:00"]     # UTC times (HH:MM) for each date
```
kevinyang-cky: I would personally prefer "daily_times".

kevinyang-cky: (minor) Is there any default value for these two keys?

charlie-becker:
> While going through preprocess, train, and rollout, I also cleaned up the logging messages (to improve user experience), but I am still deciding whether to push the changes to this PR or open another PR if it fits better.

Feel free to inject them into this PR!

config/gen_2/examples/example-end-to-end.yml
```diff
@@ -0,0 +1,362 @@
...
+    # variables:
+    #   - name: "ERA5/prognostic/3d/temperature"
+    #     levels: null                              # null = all levels
+    #   - name: "ERA5/prognostic/3d/u_component_of_wind"
+    #     levels: [500, 850, 1000]                  # subset to these model levels
+    #   - name: "ERA5/prognostic/2d/SP"
+    #   - name: "ERA5/diagnostic/2d/total_precipitation"
+
+    # CF attribute metadata. null = no variable attributes added.
+    metadata: "era5.yaml"
```
kevinyang-cky, line 332: Question: Where does this YAML file live? In `save_loc` maybe?

config/gen_2/examples/example-end-to-end.yml
```diff
@@ -0,0 +1,362 @@
...
+    # "year"  → one file per year:    2020.nc
+    # "full"  → one file per run:     full.nc
+
+    group_by: day           # null | "day" | "month" | "year" | "full"
+
+    # Variable filter. null = save all predicted variables.
+    # Each entry names a y_processed key (source/type/dim/varname).
+    # For 3D variables, "levels" subsets by model-level value (matches
+    # data.source.ERA5.levels); null or omitted = all levels.
+    variables: null
```
kevinyang-cky, line 322: This suggestion might be stretching it a bit, but I think it would also be nice to support spatial extent here, if possible.

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+    #   - name: "ERA5/prognostic/2d/SP"
+    #   - name: "ERA5/diagnostic/2d/total_precipitation"
+
+    # CF attribute metadata. null = no variable attributes added.
+    metadata: "era5.yaml"
+
+    # NetCDF encoding. "default" applies to all variables.
+    # Per-variable overrides use the full y_processed key or the short variable name.
+    encoding:
+      default:
```
kevinyang-cky: Could `dtype`, `zlib`, and `complevel` all just live under encoding?

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+    #   type: "gfs"                 # era5 (default) | gfs | gefs
+    #   cache_path: "/glade/scratch/$USER/ic_cache"
+    #   force_download: false
+
+  output:
+    format: netcdf              # netcdf | zarr
+
+    # Temporal subsampling: only write steps whose forecast hour is a multiple
+    # of this value. null = write every step.
+    output_freq: null           # null | "6h" | "24h" | ...
```
kevinyang-cky: I personally prefer "output_interval".

config/gen_2/examples/example-v2026.3.yml
```diff
@@ -0,0 +1,362 @@
...
+  batch_forecast:
+    length: "10d"                   # forecast duration; parsed relative to data.timestep
+    initialization:
+      start_date: "2020-01-01"      # first init time (inclusive); must be within data date range
+      end_date: "2020-12-31"        # last init time (inclusive)
+      date_interval: "2d"           # interval between init times
+      times: ["00:00", "12:00"]     # UTC times (HH:MM) for each date
+
+  single_forecast:
+    length: "10d"
```
kevinyang-cky: Could you add an inline comment, something like: # forecast duration (e.g. "6h", "10d", "2W") and I believe `pd.Timedelta` does not support months and years.

credit/datasets/multi_source.py
```diff
@@ -150,6 +150,7 @@ def __init__(self, config: dict[str, Any], return_target: bool = False) -> None:
             self.datasets[user_dataset_name] = cls(sub_config, return_target=return_target)
             logger.info(f"MultiSourceDataset: registered dataset '{user_dataset_name}' with class '{cls.__name__}'")
 
+        self.dt: pd.Timedelta = pd.Timedelta(config["timestep"])
```
kevinyang-cky, line 153: Isn't `config["timestep"]` read by `self._build_master_clock`? see https://github.com/NCAR/miles-credit/blob/2ef413c89001d30bb77630feba48b3636996f198/credit/datasets/multi_source.py#L209

credit/datasets/multi_source.py
```diff
@@ -203,6 +204,9 @@ def _build_master_clock(self, config: dict[str, Any]) -> pd.DatetimeIndex:
           ticks are snapped to the last native timestamp inside
           ``BaseDataset.__getitem__``.
         """
+        if "datetimes" in config:
+            return pd.DatetimeIndex(config["datetimes"])
+
```
kevinyang-cky, line 209: Curious what this is for? Is it a new key in the config?

kevinyang-cky (changes requested):
Thanks for implementing this—I’m really excited that CREDIT Gen2 now has an end-to-end pipeline in place.

I primarily focused on `example-v2026.3.yml` and left the other configs for a future pass. I ran `credit preprocess/train/rollout` as well as `torchrun --standalone --nproc_per_node=4 credit/applications/[preprocess.py/train_gen2.py/rollout_gen2.py]`. I also tried `credit submit` for training, which unexpectedly submitted 35 jobs for me (good to learn that `num_epochs` control that).

For rollout, I tested both single and batch forecasting and also experimented with different inference parameters. Everything ran successfully, and I've included my feedback in the review comments.

Besides modifying the scaling preblock in `example-v2026.3.yml`, I also enabled validation and had to make a small change to the source name.

I'll submit a separate PR to clean up the logging messages since that touches quite a few files, and I'd rather not mix those changes into this PR.

Let me know if my comments make sense, and thanks again for all the work!

config/gen_2/examples/example-v2026.3.yml
kevinyang-cky: Also, if someone wants to do very high-frequency init times, for example, GOES every 10 minutes, could it support something like just giving a start, end, and frequency? Or would one need to list them all out?

credit/datasets/multi_source.py
charlie-becker: Yes. However, it's actually used in the sampler. The sampler used to always infer "dt" from `datetimes[1] - datetimes[0]`, but this doesn't exist for single inference cases with only an initial condition, which is why the sampler now uses `dataset.dt`. This also prevents the wrong datetimes from being inferred if there is a missing timestep, for example.

credit/datasets/multi_source.py
charlie-becker: It's not a new key, it's derived from the inference mode choice (single vs batch). 

see lines 192-203 in `rollout_gen2`

config/gen_2/examples/example-v2026.3.yml
charlie-becker: I changed this to `forecast_length` which is more interpretable, and added a comment regarding timeDelta's vs Calendars

config/gen_2/examples/example-end-to-end.yml
charlie-becker: So I think there might be some confusion here, but correct me if I'm wrong. 

A "batch" forecast entails that there are more than one initialization time. I suppose you could set the start and end to the same for a single forecast, but then the other parameters (init_times and init_frequency) would be irrelevant, or you'd need to set them to null. I find this more confusing, but I am happy to continue the discussion.

config/gen_2/examples/example-end-to-end.yml
charlie-becker: Agreed, including individual point data (or a set of points), but I think this is out of scope for this PR.

config/gen_2/examples/example-end-to-end.yml
kevinyang-cky: Thanks for the explanation, and I can see it go either way. Let's keep the two-mode approaches for now, as I can imagine it will require some refactoring to collapse into a single mode.

charlie-becker:
@kevinyang-cky 

Regarding your inference sections config comments (see above replies), I do think it can be clarified a bit. What do you think about this structure:
```
inference:
  run_mode: batch                   # batch | single
  mode: ddp                         # none | ddp | fsdp
  save_forecast: '/glade/derecho/scratch/$USER/CREDIT_runs/starter_gen2/rollout'

  batch_forecast:
    forecast_length: "10d"             # total forecast duration (total valid lead time)
    first_init_date: "2020-01-01"      # first init time; must be within data date range
    last_init_date: "2020-01-31"       # last init time; must be within data date range
    init_interval: "12h"               # interval between init times

  single_forecast:
    forecast_length: "10d".            # total forecast duration (total valid lead time)
    start_datetime: "2020-06-01T00:00Z"   # must be within data date range
```

config/gen_2/examples/example-v2026.3.yml
charlie-becker: see comment below

kevinyang-cky:
> @kevinyang-cky
> 
> Regarding your inference sections config comments (see above replies), I do think it can be clarified a bit. What do you think about this structure:
> 
> ```
> inference:
>   run_mode: batch                   # batch | single
>   mode: ddp                         # none | ddp | fsdp
>   save_forecast: '/glade/derecho/scratch/$USER/CREDIT_runs/starter_gen2/rollout'
> 
>   batch_forecast:
>     forecast_length: "10d"             # total forecast duration (total valid lead time)
>     first_init_date: "2020-01-01"      # first init time; must be within data date range
>     last_init_date: "2020-01-31"       # last init time; must be within data date range
>     init_interval: "12h"               # interval between init times
> 
>   single_forecast:
>     forecast_length: "10d".            # total forecast duration (total valid lead time)
>     start_datetime: "2020-06-01T00:00Z"   # must be within data date range
> ```

I like this structure better! So abandoning keys `num_save_workers` and `date_interval`?

charlie-becker:
> I like this structure better! So abandoning keys `num_save_workers` and `date_interval`?

Yep. Thanks for flagging it. I took it out (still defaulted to 4) and is now controlled by the CLI.

Great, I'lll push these changes shortly.

charlie-becker:
Thanks for the thourough review, @kevinyang-cky!

I have made a number of suggested changes, including the new format we discussed. Ready for re-review.

kevinyang-cky (approved):
Changes look good! Ran preprocess, train, and rollout with the `example-end-to-end.yml` and successfully completed all steps.

PR #444, opened 2026-07-01
Clean up logging for preprocess, train, and rollout to improve user experience

kevinyang-cky:
## Summary
Improves logging readability across `preprocess`, `train`, and `rollout`. For DDP runs, progress output is now shown only from rank 0 to avoid interleaved logs. A `--log-all-ranks` flag has been added to all three applications to restore full multi-rank output for debugging.

Also fixes `credit submit --mode rollout` for gen2 (`inference:`-schema) configs, which previously crashed or produced broken jobs:
- Forecast counting now reads `inference.batch_forecast` instead of the gen1-only `predict.forecasts` path, which silently failed and crashed the plan-summary printer with a `TypeError`.
- Dropped the `--subset`/`--no_subset` flags passed to `rollout_gen2.py` — it doesn't implement per-job subsetting yet, so they errored as unrecognized arguments. `--jobs N > 1` currently submits N redundant full-range jobs rather than splitting the work; `--help` text now says so, and recommends `--jobs 1` until subsetting lands.
- `rollout_gen2.py`'s DDP sampler no longer pads init times to force even divisibility across ranks (previously via `DistributedMultiStepBatchSampler`), so ranks can cover an uneven number of forecasts without silently duplicating work. Added a `dist.barrier()` before process-group teardown, and disabled `broadcast_buffers` on the inference DDP wrap, so an uneven per-rank forecast count can no longer deadlock mid-loop. **This resolved the issue I raised during standup yesterday.**

**Note:** the rollout CLI flags (`--jobs`, `--subset`/`--no_subset`) should be revisited once ensemble support is fully merged into CREDIT Gen2 — this PR only unblocks single-job (`--jobs 1`) submission.

## How to test

```bash
credit submit -c config/gen_2/examples/example-end-to-end.yml --mode [preprocess | train | rollout] --cpus 9 --gpu-type v100 --gpus 2 --walltime 00:15:00 --cluster casper --conda-env credit-casper
```

The only remaining issue is that, during DDP training, the following warning message appears. @jsschreck, could you take a look when you get a chance? I believe it has something to do with WXFormer.
image

### Checklist
- [X] The documentation is up to date.

### DIFF
--- a/config/gen_2/examples/example-v2026.3.yml
+++ b/config/gen_2/examples/example-v2026.3.yml
@@ -136,7 +136,7 @@ trainer:
 
   start_epoch: 0
   num_epoch: 2            # epochs per qsub job (checkpoint after each)
-  epochs: &epochs 70      # total training target
+  epochs: &epochs 2      # total training target
 
   use_tensorboard: True
   use_ema: True           # exponential moving average of weights for validation
--- a/credit/applications/preprocess.py
+++ b/credit/applications/preprocess.py
@@ -1,22 +1,24 @@
 import argparse
 import logging
+import os
+import shutil
+import sys
 import warnings
+from os.path import expandvars
 
+import torch
+import torch.distributed as dist
+import yaml
 from bridgescaler import save_scaler_dict
+from torch.distributed import barrier, gather_object
 
 from credit.distributed import get_rank_info, setup
-from credit.seed import seed_everything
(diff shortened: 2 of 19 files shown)


djgagne (approved):
I just ran credit preprocess, train, and rollout and all three provide much better log messages with no weird errors. Merging.

PR #453, opened 2026-07-14
Camulator charlie

charlie-becker:
### Description

Updates and followup to Camulator integration. Fixes two bugs:

1. The `GlobalEnergyFixerUpDown` postblock had some inconsistent insolation and flux math when converted from gen 1 which resulted in unstable runs. Thanks @WillyChap !

2. Current datasets correctly identifies cf-time calendars and converts them properly, but errored if the source dataset had a non "standard" calendar such as `no-leap` which failed when the master time clock was built and fed through the sampler (which would be built with leap days).

Also converted the the gen1 post-processing filter to a Postblock.

### Checklist


kevinyang-cky (approved):
All merged conflicts were resolved, and process, train, and rollout using `example-end-to-end.yml` all passed. `camulator_gen2_casper.yml` also passed all three steps. 

Code changes look good to me and merging now! Thanks for the effort putting this together.

PR #456, opened 2026-07-15
Fixed miscellaneous issues with hrrr_download.py

### LINKED ISSUE #455: hrrr_download.py will not redownload missing indexes when only the index file is missing
When preprocessing HRRR data, if an .idx file is not found, hrrr.py raises a FileNotFoundError such as this:

https://github.com/NCAR/miles-credit/blob/19618822fbf1e684bc4772f876fdce8da3acda2b/credit/datasets/hrrr.py#L620-L622

In this case, the user is told to run hrrr_download.py again to re-download the grib files. The issue is what when hrrr_download.py checks for an existing file, it will only check files ending in grib2.

Local path is defined in hrrr.py and defines the path name with the .grib2 extension

https://github.com/NCAR/miles-credit/blob/19618822fbf1e684bc4772f876fdce8da3acda2b/credit/datasets/hrrr.py#L271-L285

And the download function in hrrr_download.py only checks local path without checking for a corresponding .grib2.idx before skipping. hrrr_download.py _will_ re-download both if overwriting is enabled–but this will re-download everything, not just missing files.

https://github.com/NCAR/miles-credit/blob/19618822fbf1e684bc4772f876fdce8da3acda2b/credit/datasets/hrrr_download.py#L93-L94

I already wrote a simple fix to this which is to just append .idx to `task.local_path` and check if both paths exist. If either one doesn't, then the script will redownload both the grib and the index file.

ecampb34:
### Description

Fixes a bug with hrrr_download.py where the script will not check for missing index files before deciding to skip a re-download of the gribs for a given HRRR timestep. Also adds a few quality of life features, allowing environment variables (such as $USER) to be used, allowing users to use the "~" shortcut for their home directory, and fixing an inconsistency between the docstring and the CLI arguments.

### Related Issues and PRs
Closes #455 

### Checklist

### DIFF
--- a/credit/datasets/hrrr_download.py
+++ b/credit/datasets/hrrr_download.py
@@ -90,7 +90,9 @@ def _download_one(task: _DownloadTask) -> str:
     """
     import s3fs  # noqa: PLC0415 # pyright: ignore[reportMissingTypeStubs] # local import for s3 bucket access only if needed
 
-    if os.path.exists(task.local_path) and not task.overwrite:
+    idx_local_path = task.local_path + ".idx"
+
+    if os.path.exists(task.local_path) and os.path.exists(idx_local_path) and not task.overwrite:
         return f"skip  {task.local_path}"
 
     os.makedirs(os.path.dirname(task.local_path), exist_ok=True)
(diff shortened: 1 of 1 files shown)


charlie-becker (approved):
Thank you very much for the bug report and fix! Looks good to me. Merging.

PR #461, opened 2026-07-18
Slurm

djgagne:
# Description
This PR adds SLURM queue support for `credit submit` so that we can run on Perlmutter and other systems that use SLURM rather than PBS. The new code has been tested on perlmutter and can support multi-node training over the high-speed interconnect. 

Adds SLURM support to `credit submit` / `credit_submit` alongside the existing PBS backend, and fixes a conda-env path resolution bug uncovered while adapting the launcher for NERSC Perlmutter.

- **New `credit/slurm.py`** — a SLURM analog of `credit/pbs.py` with the same public API (build/submit training, rollout, realtime, and preprocess jobs), so callers can switch schedulers without other code changes. All jobs launch via `torchrun` (never `mpirun`/`srun` alone), since `LOCAL_RANK`/`RANK`/`WORLD_SIZE` come from torchrun and plain `srun` only exports `SLURM_PROCID`. Single-node jobs run torchrun directly; multi-node jobs use `srun` to launch one torchrun per node with a c10d rendezvous on the head node.
- **`credit/cli/_submit.py`** — mirrors the PBS helpers with SLURM equivalents (`_load_slurm_config`, `_resolve_slurm_opts`, `_build_slurm_script`, `_build_realtime_slurm_script`, `_build_preprocess_slurm_script`, `_build_rollout_slurm_script`, `_sbatch`), and adds a `_scheduler_ctx()` dispatcher so `_submit`, `_do_submit_rollout`, `_do_submit_realtime`, and `_do_submit_preprocess` route to PBS or SLURM based on a new `--scheduler` flag, instead of duplicating each entry point.
- **`credit/cli/_parser.py`** — adds `--scheduler {pbs,slurm}` (default `pbs`) and SLURM-specific `--constraint`/`--qos` flags; `--cluster` is no longer restricted to `choices=["casper", "derecho"]` so arbitrary SLURM sites can be named.
- **Per-cluster defaults** (`credit/cli/_common.py`) — generic SLURM sites request GPUs with `--gres=gpu:N`; Perlmutter is special-cased (`_SLURM_CLUSTER_DEFAULTS["perlmutter"]`) since it rejects `--gres` and instead needs `--constraint=gpu` + `--qos` + `--gpus-per-node`, no `--partition`/`--mem` line, and a `_g`-suffixed account for GPU allocations.
- **Bug fix — `_resolve_torchrun()`**: previously, a conda env given as a bare name (e.g. `conda: credit`) was checked with `os.path.isdir(conda_env)` against the *current working directory*, so running from the repo root treated the `credit/` package directory as the conda env and produced a bogus `credit/bin/torchrun`. `_resolve_torchrun` now only treats a value as a literal path if it contains a path separator; a bare name instead resolves lazily at job runtime via `$(conda info --base)/envs/<name>/bin/torchrun`. Applied uniformly across all PBS and SLURM script builders.
- **Perlmutter NCCL/libfabric env** (`credit/cli/_common.py`, `credit/slurm.py`) — Perlmutter jobs now auto-inject the `module load nccl/2.24.3` line and the `NCCL_*`/`FI_CXI_*`/`MPICH_GPU_SUPPORT_ENABLED` exports needed to route torch's bundled NCCL over the Slingshot 11 fabric via the system AWS-OFI plugin (`LD_LIBRARY_PATH` pointed at the NCCL plugin dir, `NCCL_NET=AWS Libfabric`, `NCCL_SOCKET_IFNAME=hsn`, etc.), since `module load nccl` alone doesn't reliably export these. Config-supplied `modules`/`env_setup` extend rather than replace these defaults, so a config can add e.g. `NCCL_DEBUG` without losing the baked-in Perlmutter setup. `credit/slurm.py`'s own script builder was updated to match, reusing the same `_PERLMUTTER_ENV_SETUP` constant so both entry points stay in sync.

## Tests

- `tests/test_pbs_scripts.py` — SLURM script generation (directives, GPU/constraint/qos handling, single- and multi-node torchrun launch commands, Perlmutter-specific overrides including the NCCL/libfabric env injection).
- `tests/test_cli.py` — `_resolve_torchrun` unit tests, including a regression test for the CWD-shadowing bug (bare env name must resolve via `conda info --base`, not a same-named directory in the cwd).

# Related Issues and PRs

None
### Checklist


jsschreck (approved):
Looks good please merge!

PR #463, opened 2026-07-21
Parallel submission on Derecho with pbsdsh

djgagne:
### Summary

Adds an alternative multi-node job launcher for Derecho jobs. Alongside the existing MPI-based launch, `credit submit` can now spawn one `torchrun` per node via PBS Pro's native task launcher (`pbsdsh`), with NCCL running over the `libfabric` module (aws-ofi-nccl plugin) for inter-node communication.

### Motivation

The default `mpiexec` launcher runs one Python process per GPU under cray-mpich, relying on MPI's PMI environment for rank identity. The new `pbsdsh` path removes MPI from process management entirely — each node runs a self-contained `torchrun` with static rendezvous, and rank identity comes from torchrun (`LOCAL_RANK`/`RANK`/`WORLD_SIZE`). This gives a torchrun-native launch on Derecho to trial side-by-side against the mpiexec approach.

### Changes

**CLI (`credit/cli/`)**
- New `--launcher {mpiexec,pbsdsh}` flag (default `mpiexec`), Derecho-only.
- When `--nodes > 1` and `--launcher pbsdsh`, both `--mode train` and `--mode preprocess` generate a pbsdsh script via the shared launch block.
- Single-node jobs always use `torchrun --standalone` regardless of `--launcher`.

**PBS script generation (`credit/pbs.py`)**
- `_pbsdsh_launch_block()` — the shared, launcher-specific bash tail, used by both `credit submit`'s integrated path and the standalone builder so the two stay in lockstep. Key design points, documented inline:
  - **One pbsdsh task per node**, each running `torchrun --nproc-per-node` (not torchrun *under* mpiexec, which double-spawns ranks).
  - **Static rendezvous** with explicit `--node-rank` (independently-spawned tasks can't safely negotiate ranks by arrival order).
  - **Baked environment** — the mother-superior's resolved `PATH`/`LD_LIBRARY_PATH` are written into each per-node script, because pbsdsh's spawned shell inherits neither modules nor conda.
  - **PBS Pro node indexing** — `MASTER_ADDR` resolved locally (`hostname -i`) since pbsdsh index 0 is the mother superior.
  - Per-node logs written to disk (`pbsdsh_logs/`) independently of the job's aggregate stdout.
- `PBSDSH_DERECHO_MODULES` — the derecho module set plus `libfabric`; retains `cray-mpich` because libtorch links Cray's `libmpi` at runtime.
- `_build_pbsdsh_script()` / `launch_script_pbsdsh()` — standalone/legacy full-script builder and entry point (marked **prototype**, for on-hardware trial alongside `launch_script_torchrun`).

**Docs (`docs/source/Training.md`)**
- New "Multi-node launcher (derecho)" section comparing the two launchers with examples.

**Tests (`tests/test_pbs_dsh.py`)**
- String-building tests (no filesystem/qsub side effects, no cluster needed) covering both the standalone builder and the `credit submit --launcher pbsdsh` integrated path, checked side-by-side against the default mpiexec launcher.

**Smoke test configs (`tests/manual/gen2_parallelism/`)**
- Updated paths (`save_loc`, conda env) to the current user and aligned settings (`history_len`, validation `timestep`/`forecast_len`, pre-scaler path) to match the end-to-end config.

### Testing

- `pytest tests/test_pbs_dsh.py` — passes (string-building only, CI-safe).
- The launch mechanism itself (pbsdsh rendezvous + NCCL over libfabric) is validated separately on real Derecho hardware. NCCL transport can be confirmed via `NCCL_DEBUG=INFO` (expect `NET/OFI` over libfabric).

### Notes

This is intended as a side-by-side option, not a replacement for the mpiexec launcher — recommend confirming behavior with `--dry-run` before submitting.

### Related Issues and PRs

Fixes issues reported on Slack with manual 2-node smoke tests not running and if working should pave the way for removing the need to pytorch compiled with MPI on Derecho. 

### Checklist


jsschreck (approved):
Looks good to me! Thanks for the gen2 smoke test updates.

PR #464, opened 2026-07-22
Rollout coords

### LINKED ISSUE #457: Rollout output coordinate schema (ForecastWriter) breaks for curvilinear (non-rectangular) grids
### Summary

`ForecastWriter._init_coords` (`credit/output_gen2.py:393-415`) always builds the output NetCDF/Zarr `latitude`/`longitude` coordinates as a synthetic **global** grid:

```python
lat = np.linspace(-90.0, 90.0, H) if H else np.array([])
lon = np.linspace(0.0, 360.0 - 360.0 / W, W) if W else np.array([])
```

where `H, W` come from `conf["model"]["image_height"/"image_width"]`. This is only correct if the source's actual grid is a regular lat/lon mesh spanning the full globe at that exact resolution. It is silently wrong for any source that is regional and/or curvilinear/projected.

### Impact

Metadata-only — model training, inference, and forecast *values* are unaffected (the model correctly consumes/produces the cropped tensor). The coordinate *labels* attached to those values in the output file are fabricated and don't correspond to real geography. Any downstream consumer that trusts the written `latitude`/`longitude` for georeferencing, plotting, or alignment with other datasets gets silently mis-georeferenced data.

### Affected datasets (`credit/datasets/gen_2/`)

| Dataset | Affected? | Why |
|---|---|---|
| `GOESDataset` (`goes.py`) | **Yes** | Curvilinear satellite projection, regional via `extent` |
| `HRRRDataset` (`hrrr.py`) | **Yes** | Curvilinear Lambert Conformal, regional (CONUS + `extent`) |
| `MRMSDataset` (`mrms.py`) | **Yes** | Regular grid but regional (`region`/`extent`) — wrong coverage, not wrong projection |
| `TISRDataset` (`tisr.py`) | **Possibly** | Can be curvilinear/regional depending on config |
| `ARCOERA5Dataset` / `WeatherBench2ERA5Dataset` (`era5.py`) | No | Regular, global grid — correct as long as `model.image_height/width` matches the store resolution |
| `LocalDataset` (`local.py`) | Depends on user's data | No built-in guardrail either way |

charlie-becker:
### Description

Two main changes:

1. Added support for bridgescaler blocks to scale and handle scaling independently in space. This is effectively done by reshaping the spatial coordinates and stacking on top of the channels dimension which controls how many many scalers are fit. 
2. Fixed the rollout to align with proper coordinates from the input grid (previously assumed global lat / lon). This also works with a regrdding preblock, and preserves the coordinates on the regridded destination grid for output. It now also writes out grid_files, similar to how channel_schema, in `save_loc`. When a regrdding preblock is used, it saves both the source grid and the output grid. This metadata is now also tracked as metadata as an attribute of each dataset. 

I have tested this using the `example-end-to-end` and `intergration_test_hrrr` configs from the API across preprocessing, training, and rollout successfully. 

### Limitations

Currently only supports rectilinear and curvilinear grids. Support for unstructured grids will be its own PR. 

### Related Issues and PRs

Closes #457 

### Checklist


djgagne:
I had Claude run through the PR checklist and it found some missing test coverage on the spatial remapping for bridgescaler. It added some tests that passed.

djgagne (approved):
The changes look good to me. I appreciate having a way to do local scalers easily that still runs smoothly.

PR #468, opened 2026-07-29
Inference Override + Unstructured Data support

### LINKED ISSUE #448: Inference vs training config
Make sure training and inference sections in config file are clearly separate and demarcated. Be able to override training_data, preblocks, and postblocks for inference with same model.

charlie-becker:
### Description
Two main updates:

1. Support for Unstructured grids as input for LocalDataset and the Regridder preblock. 
-  Adds proper metadata into the grid schema
- Regridder now properly utilizes `reshape_to_xy` when going from unstructured to structured.
2. Allows users to override the "data:", "preblocks:", and "postblocks:" under the "inference:" portion of the config which preserves training data details in the config when running inference with a separate configuration.
- This is an all-or-nothing replacement for any/all of the three types (data / preblocks / postblocks). For example, a single additional postblock under "inference" will not append to the "training" version, it replaces it in whole.
- `start_datetime` and `end_datetime` are not necessary in the data override version as it's derived earlier in the inference dictionary.
- Added a `RenameVariables` preblock that utilizes a metadata file to map variable names from one dataset to another to resolve any mismatches in sources / names for the inference preblock chain. Can be read from a file or listed explicitly as an argument. 

### Related Issues and PRs

Closes #448 

### Checklist


djgagne (commented):
## Review against the PR checklist

### Checklist

**Required**

| Item | Status |
|---|---|
| Clear description of issue/feature | ✅ Thorough, with the all-or-nothing override semantics called out explicitly |
| Issues tagged | ✅ Closes #448 |
| CI checks and tests pass | ✅ build, py3.11/3.12/3.13, pre-commit, RTD all green |
| Tests updated to cover code changes | ⚠️ Partial — see (5) |
| Documentation updated | ❌ Unchecked by the author; no example config or doc page exercises either feature |
| pyproject.toml / requirements.txt updated | ✅ N/A — no new dependencies (`yaml` was already in use) |
| Paths user-generic | ⚠️ One dangling example path in a docstring — see (8) |
| GPU verification on Casper/Derecho | ❓ Could you confirm you've run this end-to-end on Casper/Derecho? |

**Recommended:** docstrings ✅, naming ✅, comments on complex operations ✅, redundancy minimized ✅, type annotations ✅, no dependency changes ✅.

---

### What's good

The docstrings here carry real reasoning rather than restating signatures, which made this a fast review. Specifically: the `temporal_mode: persist` / `history_len > 1` caveat in `with_inference_datetime_bounds`, the "auto-detection is a size-based heuristic and can misfire on a square rectilinear grid" note in `LocalDataset`, and the explanation in `_read_3d_array` of *why* a positional dimension assumption needs an explicit guard. Those are all things a future reader needs and wouldn't otherwise have.

Also worth calling out:
- `{"start_datetime": ..., "end_datetime": ..., **data_conf}` is exactly the right idiom for explicit-wins precedence.
- The new `GridSchema.resolve` guard produces an actionable error message pointing at the fix (add a Regridder preblock) instead of a downstream shape crash.
- The unstructured `_regrid` path is a clean generalization, and `test_regrid_unstructured_input` asserting bit-equality against the structured result is precisely the right test.
- Commits are atomic and readable.

---

### Blocking

**1. `history_len > 1` now fails with a spurious `ChannelSchema mismatch` — in training as well as rollout.**

`credit/preblock/concat.py:174-176` validates the input channel map unconditionally. But `ChannelSchema.from_config` emits `n_time: 1` for every input entry (`credit/datasets/gen_2/channel_utils.py`), while `ConcatToTensor` reads `T = tensor.shape[2]`, which is `history_len` for input tensors. Minimal repro:

```python
conf = {'data': {'source': {'era5': {
    'levels': [1, 2], 'history_len': 2,
    'variables': {'prognostic': {'vars_3D': ['T'], 'vars_2D': ['SP']}}}}}}
schema = ChannelSchema.from_config(conf)
batch = {'input': {'era5': {
    'era5/prognostic/3d/T':  torch.randn(1, 2, 2, 4, 4),
    'era5/prognostic/2d/SP': torch.randn(1, 1, 2, 4, 4)}}}
ct = ConcatToTensor(); ct.set_schema(schema); ct(batch)
```

```
ChannelSchema mismatch (input map): ...
Schema expects:  [0:2] era5/prognostic/3d/T shape=(2, 1)
Data produced:   [0:4] era5/prognostic/3d/T shape=(2, 2)
```

`trainer_gen2.py:104` attaches the schema to the preblocks, so this breaks `credit_train` too, not just rollout. Every shipped gen2 config uses `history_len: 1`, which is why CI stays green — but `history_len` is a documented, supported feature (`example-end-to-end.yml:54`) with explicit handling in the trainer and in `assemble_rollout_batch`. Fix by deriving the input layout's `n_time` from `history_len`, or by comparing only `(var_key, n_levels)` on the input side.

**2. `RenameVariables` silently breaks multi-step rollout — there is no working placement.**

`assemble_rollout_batch` (`credit/trainers/rollout_utils.py:369-397`) keys off `ic_preprocessed["input"]` and then looks up both `y_processed` (post-`Reconstruct`, so schema/training names) and `curr_batch["input"]` (the raw dataloader batch, so original source names):

- **`rename` under `ic_only`:** `ic_preprocessed` and `y_processed` agree on `ERA5/...`, but `frc_batch` is still `GFS/...` → every dynamic_forcing lookup misses → `logger.warning` + forcing frozen at t=0 for the entire forecast.
- **`rename` under `per_step`:** `ic_preprocessed` keeps `GFS/...` while `y_processed` is `ERA5/...` → every prognostic lookup misses → prognostic state frozen at the IC, i.e. a constant forecast.
- **Both phases:** fixes the prognostic side; dynamic_forcing still breaks, because assembly runs before `step_preblocks`.

Both failure modes are `warning`-level, so the user gets a plausible-looking but wrong multi-day forecast rather than a crash. Since GFS-driven rollout of an ERA5-trained model is the motivating use case for this preblock, this needs resolving — either rename `frc_batch` before assembly, or have `assemble_rollout_batch` reconcile through the channel map instead of raw source keys. (Single-step rollouts are unaffected.)

---

### Non-blocking

**3. `credit/preblock/regrid.py:143` — the `flip_axis` safeguard doesn't guard much.** `[d for d in self.flip_axis if d >= -spatial_dims]` filters only *negative* out-of-range axes; a positive index such as `2` passes through and flips a leading dim. It also drops requested flips silently — a warning when `valid_flips != self.flip_axis` would surface a misconfiguration instead of hiding it.

**4. `credit/applications/rollout_gen2.py:152` vs `:225` — override ordering makes the new input validation vacuous in one case.** `apply_inference_overrides(conf)` runs before `ChannelSchema.load_or_from_config(conf)`. When `channel_schema.yaml` isn't present in `save_loc`, the schema is derived from the *inference* data block, so the new input-side validation trivially passes and the target reconstruction map describes the inference source rather than what the model actually predicts. Consider deriving the schema from the pre-override config, or warning loudly when the file is missing and an override is active.

**5. Test coverage gaps.** Nothing covers `apply_inference_overrides`, `with_inference_datetime_bounds`, `LocalDataset._find_grid`'s unstructured detection (including the "weaker fallback" branch the docstring itself flags as misfire-prone), the new `_read_3d_array` level-order guard, or the unstructured branches in `write_source_grid_schema_if_missing` / `GridSchema.resolve`. These are all cheap — pure-function tests plus a small synthetic NetCDF fixture.

**6. `credit/preblock/regrid.py` (around lines 90 and 105) — three explanatory comments were deleted with no corresponding code change.** The "Real destination-grid coordinates…" and "Lazy sparse weight cache…" notes look accidentally dropped; both explain non-obvious state.

**7. `credit/preblock/rename.py:81` — chained mappings are iteration-order dependent.** `{"A": "B", "B": "C"}` walks A's tensor all the way to C. Either resolve the whole mapping against a snapshot of the original keys, or document/reject chains explicitly.

**8. `credit/preblock/rename.py:47` — the docstring example points at `credit/metadata/gfs_to_era5_rename.yaml`, which doesn't exist.** `credit/metadata/` is a real directory, so this reads as a real path. Either ship the file (it would be genuinely useful given the motivating use case) or use an obviously-placeholder `$SCRATCH/...` path, matching `Regridder`'s convention.

**9. Minor:** `credit/preblock/rename.py` has no module-level docstring, unlike every sibling preblock module.

djgagne (approved):
I addressed the issues brought up by the LLM review and fixed some other formatting issues I caught. Looks good to me now. Waiting for CIRRUS check to finish and then will merge.

PR #472, opened 2026-08-03
Gen 2 Loss and Metrics

### LINKED ISSUE #350: Add computed_diagnostic data type to data schema
I want to add a new data category for diagnostic variables that are computed by a postblock and are included in the loss calculation, like geopotential or mean sea level pressure. These variables are not output by the model directly, but we do want to incorporate them in the training loss to encourage the model to enforce physical consistency. The new category is `computed_diagnostic`. 

For variables that we want to compute later but not include in the training, like fields on pressure levels, we can just define postblocks that can be added at inference time to the config file. Maybe we also include an inference_only flag for those postblocks so they are skipped during training?

### LINKED ISSUE #351: Align metadata with downstream tasks (metrics, etc)
There are downstream tasks such as calculating metrics that rely on metadata structure mostly in a "gen 1" schema. 

An example of this is the function ` _inject_flat_schema()` in `credit/applications/rollout_to_netcdf_gen2.py`, which creates some meta data based on the config. However, in our new data scheme we output metadata directly from `apply_postblocks()`.

We'll need to adjust downstream tasks to use this metadata directly and potentially add additional metadata if needed.

### LINKED ISSUE #402: Update Losses to Gen 2
Losses need to be updated to accept Gen 2 state dict, handle `computed_diagnostics` variables. MSE definitely should follow this framework. Other losses could potentially be added in 2026.2.1.

djgagne:
### Description
# Gen 2 loss and metrics frameworks

Adds two config-driven, per-variable scoring frameworks for Gen 2 — `BaseLoss`
and the `credit.metrics` package — that operate on the postblock output in
**physical units** rather than on the flat, normalized `y_pred` tensor. Also
reorganizes the Gen 1 losses/metrics into `gen_1/` subpackages and fixes several
multi-rank/cluster robustness issues found along the way.

`main...g2loss`: 10 commits, 58 files changed, +3880 / −314.

## `BaseLoss` — per-variable loss on the postblock state (`credit/losses/base.py`,

Scores `full_data_dict["y_processed"]` against a **target twin**
`y_target_processed` (the same postblock chain applied to the flat target `y`),
variable by variable, then combines the per-variable scores.

- Any elementwise registry loss (`mse`, `mae`, `huber`, `logcosh`, …) as the
  univariate base loss, with optional per-variable overrides via
  `base_loss_overrides`. CRPS-family losses are rejected with a clear error.
- `var_weighting` modes:
  - `inverse_variance` (default) — `1/σ_v²` read from the fitted bridgescaler
    (`DStandardScalerTensor.var_x_`, `DQuantileScalerTensor` t-digest centroid
    moments, numpy `DeepStandardScaler.sd_`). For MSE this approximates
    training in normalized space.
  - `manual`, `none`, and `learnable` (Kendall–Gal uncertainty weighting with a
    trainable per-variable `log σ_v²`).
- Optional `cos(lat)` spatial weighting; `variable_weights` multipliers apply on
  top of every mode.
- Data diagnostics are always scored; postblock-**computed** diagnostics (e.g.
  `mslp_diagnostic`) are opt-in via `include_computed_diagnostics`.
- `last_var_losses` exposes detached per-variable scores for logging.

New-style `{type, args}` loss section (mirroring preblocks/postblocks):

```yaml
loss:
  type: base
  args:
    training_loss: "mse"
    var_weighting: "inverse_variance"
    scaler_path: /path/scaler.json
```

`load_loss` dispatches on `type` when present and falls back to the legacy flat
keys otherwise, so existing configs are unaffected.

## `credit.metrics` package (new)

Mirrors the loss framework for verification metrics.

- `BaseVariableMetric` / `BaseCombinedMetric` (`base.py`) — same `full_data_dict`
  contract, same variable-weighting and latitude-weighting options, run under
  `torch.no_grad()` and return detached floats.
- `common.py` — `RMSEMetric`, `MSEMetric`, `MAEMetric`, `BiasMetric`,
  `R2ScoreMetric`, `LogVarianceRatioMetric`.
- `anomaly.py` — `AnomalyCorrelationCoefficientMetric` (ACC) and
  `ForecastActivityMetric` (SDAF) following Bonavita & Geer (2026), *Forecast
  verification using information and noise*, QJRMS 152, e70109
  ([doi:10.1002/qj.70109](https://doi.org/10.1002/qj.70109)). Climatology comes
  from a file, a supplied dict, or an online running mean of the validation
  targets.
- `load_metric(conf)` dispatches on `conf["metrics"]["type"]`; when the section
  is omitted it defaults to a combined `rmse` / `r2score` / `bias` with uniform
  weighting. `@register_metric` plus a `metric` entry in `custom_objects` allows
  external metric classes.

## Package reorganization

- `credit/losses/{almost_fair_crps,covariance,downscaling_loss,kcrps,les_loss,logco_loss,xsigmoid,xtanh}.py` → `credit/losses/gen_1/`
- `credit/metrics.py` → `credit/metrics/gen_1/metrics.py`

`from credit.metrics import LatWeightedMetrics` still works — the Gen 1 classes
are re-exported lazily from the new package `__init__`. Direct submodule imports
(`credit.losses.kcrps` → `credit.losses.gen_1.kcrps`) changed and were updated
across applications and tests.

## Trainer and application integration

- `TrainerERA5Gen2` branches on `isinstance(criterion, BaseLoss)` /
  `BaseCombinedMetric` for both train and validate; the legacy flat-tensor path
  is unchanged. Per-variable losses are all-reduced and logged as
  `train_loss_var/<var>` / `valid_loss_var/<var>`.
- Clamping of `y` now happens **before** the postblocks when `BaseLoss` is
  active, so `y_target_processed` inherits the clamp.
- `BaseTrainer` gets configurable `display_metrics` (derived from the `metrics`
  config, default `rmse`/`r2score`/`bias`) replacing the hard-coded
  `acc`/`mae` in progress bars, CSV columns, and TensorBoard tags, plus a new
  `_log_epoch_metrics` epoch summary.
- `train_gen2` loads metrics via `load_metric(conf)` and adds `BaseLoss`
  learnable variance parameters to the optimizer when `var_weighting:
  learnable`. **Note:** those parameters are not yet checkpointed and re-init on
  resume.
- `train_gen2`'s `-l` PBS self-launch flag was removed (use the `credit` CLI).

## Supporting changes

- **`Reconstruct` postblock** takes `in_key` / `out_key`, so the same block can
  build the `y_target_processed` twin from the flat target.
- **New `to_device` preblock** (`credit/preblock/device.py`) — recursively moves
  every tensor in the nested state dict to a device, preserving containers and
  non-tensor metadata.
- **`credit/parser.py`** skips the legacy flat-key loss assertions for
  `{type, args}` sections.
- **`credit/registry.py`** gains the `metric` object type.

## Bug fixes

- **Corrupted schema files under multi-rank writes** — `ChannelSchema.save` and
  `GridSchema.save` staged into a single shared `<path>.tmp`; concurrent ranks
  and DataLoader workers interleaved their writes (surfacing as `[Errno 13]
  Permission denied` from HDF5 for `GridSchema`). Now per-PID temp files with
  cleanup on failure.
- **torchrun resolution for conda envs outside the base install** —
  `$(conda info --base)/envs/<name>` silently produced a nonexistent path for
  envs under `envs_dirs` (e.g. `/glade/work/$USER/conda-envs`). Now resolves
  `${CONDA_PREFIX:-$(conda info --envs lookup)}` at run time in `credit/pbs.py`,
  `credit/slurm.py`, and `credit/cli/_common.py`.
- **Scaler overwrite** — `preprocess` now moves an existing scaler aside with a
  UTC timestamp before writing, and expands `$VARS` in the scaler path.
- Narrowed bare `except Exception` on config loading to `OSError` /
  `yaml.YAMLError` in `train_gen2` and `preprocess`.

## Tooling / config / docs

- **`pyproject.toml`**: pinned `[tool.ruff.lint] select = ["E4","E7","E9","F"]`.
  Ruff 0.16 greatly expanded its default `select`, so a hook bump would have
  silently rewritten hundreds of files.
- **`requirements.txt`**: added `obstore`.
- Both `config/gen_2/examples/*.yml` updated with the new `loss` / `metrics`
  sections and the target-twin postblocks; also fixes a
  `bridgescaler_transformer` → `bridgescaler_transform` typo in
  `example-v2026.2.yml`.
- Docs and `climate/` READMEs reformatted for the pinned ruff rule set.

## Tests

- `tests/test_losses_base.py` — 27 tests (new)
- `tests/test_metrics_base.py` — 45 tests (new)
- Additions to `test_registry.py` (metric registration + `custom_objects`),
  `test_trainer_components.py` (display metrics, epoch logging, CSV/TensorBoard
  export), `test_preblock.py` (`ToDevice`), `test_preprocess.py` (scaler
  backup), and `test_cli.py` (conda env resolution).

### Related Issues and PRs

Closes #350
Closes #351 
Closes #402

### Checklist


djgagne (commented):
## Review against the PR checklist

Local test run: **450 passed** (`test_losses_base`, `test_metrics_base`, `test_registry`, `test_preblock`, `test_preprocess`, `test_trainer_components`, `test_cli`). CI green on Python 3.11/3.12/3.13; `build` and Read the Docs pending at time of review.

### Positive

Well-engineered PR. The module docstrings in `credit/losses/base.py` and `credit/metrics/base.py` are the best in the repo — they document the config format, the postblock contract, and *why* `inverse_variance` exists (physical-unit scale disparity), not just what it does. Error messages give the user the exact YAML to add rather than just naming what's missing. The three bug fixes each explain the root cause in the docstring; the `GridSchema` one (shared `.tmp` colliding inside HDF5, surfacing as `Errno 13` for every writer but the first) is a genuinely non-obvious diagnosis. The ruff `select` pin with its rationale comment prevents a real future footgun. 72 new tests plus targeted additions to five existing files.

### Required checklist

| Item | Status |
|---|---|
| Clear description | ✅ Thorough, including a self-disclosed limitation (learnable params not checkpointed) |
| Issues tagged | ✅ Closes #350, #351, #402 |
| CI checks and tests pass | ✅ 450 pass locally; 3.11/3.12/3.13 green |
| Tests updated to cover changes | ⚠️ Gap — see F3 |
| Documentation updated | ❌ See F2 |
| Dependency lists updated | ✅ `obstore` added to `requirements.txt`; already in `pyproject.toml` — now consistent |
| Paths user generic | ✅ New paths use `$USER` or `/glade/campaign/cisl/aiml/credit/`. (`aiml/ksha/` appears in new `latitude_weights` lines but pre-exists 9× in both configs — pre-existing, not introduced here) |
| GPU functionality verified on Casper/Derecho | ❓ Not stated — the `trainer_gen2` train/validate branches are GPU-path changes |

---

## Findings

### F1 — Blocking: the `gen_1/` move broke 8 registered loss types

`credit/losses/__init__.py:19-27` and `:44-51` still point at the pre-move module paths. `load_loss` with `logcosh`, `xtanh`, `xsigmoid`, `KCRPS`, `almost-fair-crps`, `spectral`, `power`, or `covmse` now fails, as does `from credit.losses import LogCoshLoss` — which the comment at `:41` calls "kept for backward compatibility". This also breaks `training_loss` / `base_loss_overrides` inside the new `BaseLoss`, since it resolves through the same `_load_loss_entry`.

`_load_loss_entry` catches `ImportError`, so the underlying `ModuleNotFoundError` is masked as *"requires optional dependencies that are not installed"* — actively misleading during debugging.

Verified:

```
_load_loss_entry('logcosh')  ->  ImportError: ... optional dependencies not installed
import credit.losses.logcosh         ->  MISS
import credit.losses.gen_1.logcosh   ->  OK
```

Single-file fix: prefix those 16 entries with `gen_1.`. `VariableTotalLoss2D` and `DownscalingLoss` were updated correctly in `__getattr__`, so only the two registry tables were missed.

### F2 — Documentation checklist item is not met

The entire `docs/` diff is ruff-driven code-block reformatting. Nothing under `docs/source/` mentions `BaseLoss`, `var_weighting`, `load_metric`, or `BaseCombinedMetric`, and no page was added to `index.rst`. The docstrings are excellent but aren't reachable from the user guide. `Losses.md` needs a Gen 2 section, and the metrics framework needs a page (or a section in `Evaluation.md`).

### F3 — Test gap that let F1 through

No test anywhere loads a registry loss by name — `grep` for `logcosh|xtanh|covmse|spectral` across `tests/` returns nothing. A parametrized test over `_LOSS_REGISTRY` calling `_load_loss_entry` would have caught F1 and is roughly four lines.

Separately, `climatology_path` — documented as climatology source #1 for ACC/SDAF — has no test. Only the in-memory `climatology=` dict and the online accumulation path are covered, which is how F4 survived.

### F4 — Blocking: 3-D climatology loaded from file raises

`credit/metrics/anomaly.py:181-182` left-pads with `unsqueeze(0)` until `ndim` matches. A `(level, lat, lon)` field becomes `(1, 1, L, H, W)`, putting `level` on the **time** axis instead of the channel axis; `expand_as` at `:260` then fails. Reproduced with a 4-level netCDF climatology and `(B, L, T, H, W) = (2, 4, 1, 8, 16)`:

```
RuntimeError: The expanded size of the tensor (1) must match the existing size (4)
at non-singleton dimension 2. Target sizes: [2, 4, 1, 8, 16]. Tensor sizes: [1, 1, 4, 8, 16]
```

2-D variables work; every 3-D variable breaks. Fix by inserting the missing axes on the right (before `H, W`) rather than the left, or by aligning explicitly against the channel/level count.

### F5 — `inverse_variance` is dimensionally wrong for non-quadratic metrics

The `1/sigma^2` weight exactly recovers normalized-space MSE — correct for `BaseLoss`. But both example configs ship `var_weighting: "inverse_variance"` with `metrics: {rmse, r2score, bias}`:

- RMSE and bias are first-order in sigma, so `1/sigma^2` over-weights low-variance variables quadratically.
- R² and ACC are already dimensionless, so scaling them by `1/sigma^2` makes the aggregate meaningless.

Per-variable scores are unaffected — only the combined aggregate. Note `load_metric`'s built-in default correctly uses `"none"`; the shipped example configs and the `BaseCombinedMetric` class default disagree with it. Suggest either `1/sigma` for first-order metrics, or defaulting the example configs to `"none"`.

### F6 — `base_metric_overrides` is advertised but not implemented

Both example configs show it in a commented block; there is no implementation in `credit/`. Either implement it (mirroring `base_loss_overrides`) or drop the comment.

---

## Minor

- **Redundancy** (checklist: *Code minimizes redundancy*): the conda-prefix `awk` lookup is now triplicated verbatim in `credit/cli/_common.py:161-162`, `credit/pbs.py:229-230`, and `credit/slurm.py:157-158`. `_conda_prefix_expr` already exists — the other two should import it.
- `last_var_losses` / `last_var_scores` are documented as attributes but only created in `forward`; accessing before the first pass raises `AttributeError`. Initialize to `{}` in `__init__`.
- `credit/losses/base.py:428` uses `p` after the loop that bound it. Works today, but bind the device explicitly.
- Learnable variance params not being checkpointed is honestly disclosed in the description — worth a tracking issue so it isn't lost.

## Recommendation

F1 and F4 are real breakage and both are small fixes; F3's registry test should land with them. F2 is the one required checklist item that can't be checked off as-is. F5 and F6 are worth resolving before merge since they ship in the example configs users copy.

djgagne:
Pushed 5ddb7941 addressing **F1**, **F3**, and **F4** from the review above.

### F1 — loss registry paths (`credit/losses/__init__.py`)

Prefixed the 8 moved modules with `gen_1.` in both `_LOSS_REGISTRY` and `_CLASS_SOURCES`, plus the stale example in the comment above `__getattr__`. `msle` and `ring-crps` are untouched — those modules did not move.

`logcosh`, `xtanh`, `xsigmoid`, `KCRPS`, `almost-fair-crps`, `spectral`, `power`, and `covmse` all resolve again, as do the backward-compatible `from credit.losses import LogCoshLoss` style imports.

### F4 — 3-D climatology from file (`credit/metrics/anomaly.py`)

`_get_clim` now inserts the missing axes at their correct positions for 5-D refs rather than left-padding:

- `(lat, lon)` → `(1, 1, 1, H, W)`
- `(level, lat, lon)` → `(1, L, 1, H, W)`

Added two guards with actionable messages — rejecting an unexpected rank, and rejecting a level count that disagrees with the field. The old left-pad remains as the fallback for non-5-D refs. 2-D variables were never affected.

The reproduction from the review now returns a valid score instead of raising:

```
{'acc/ERA5/prognostic/3d/temperature': 0.502, 'acc': 0.502}
```

### F3 — tests

- `tests/test_loss.py` — `test_all_registry_losses_resolve` (parametrized over every `_LOSS_REGISTRY` entry via `_load_loss_entry`) and `test_all_class_source_losses_importable` (attribute access for every `_CLASS_SOURCES` entry, since F1 broke both tables).
- `tests/test_metrics_base.py` — `test_acc_climatology_from_file`, `test_activity_climatology_from_file_matches_manual` (checks the SDAF values, not just that it runs), and `test_climatology_file_level_mismatch_raises`. Each uses a netCDF fixture with one 3-D (`T`) and two 2-D (`SP`, `precip`) variables.

Both fixes were reverted individually to confirm the new tests fail against the unfixed code, so these are real regression guards rather than passing vacuously.

467 tests pass locally across the affected files; `ruff check` and `ruff format --check` are clean.

### Still open

**F2** (no Sphinx docs for either new framework — the one required checklist item still unmet), **F5** (`inverse_variance` weighting is dimensionally wrong for RMSE / bias / R² / ACC, and ships that way in both example configs), **F6** (`base_metric_overrides` advertised in the example configs but not implemented), and the minor items (conda-logic triplication across `_common.py` / `pbs.py` / `slurm.py`, uninitialized `last_var_losses` / `last_var_scores`, the loop-variable use at `credit/losses/base.py:428`, and a tracking issue for learnable-variance checkpointing).

djgagne:
Pushed d79cb933 addressing **F5** and **F6**.

### F5 — `inverse_variance` now scales by metric order

`BaseVariableMetric` gains a `scale_power` class attribute — the power of sigma a metric's per-variable score carries — and `inverse_variance` weights by `1 / sigma ** scale_power` instead of always `1 / sigma^2`:

| scale_power | Metrics | Rationale |
|---|---|---|
| 2 | `mse` | Quadratic; `1/sigma^2` recovers the normalized-space score exactly (unchanged) |
| 1 | `rmse`, `mae`, `bias`, `activity` | Linear in sigma |
| 0 | `r2score`, `acc`, `log_variance_ratio` | Already dimensionless; the scaler is not consulted at all |

Only the combined aggregate was affected — per-variable scores were always correct.

A `scale_power == 0` metric no longer requires `scaler_path`, since it never reads the scaler. When `scaler_path` *is* required, the error now names the metric and its `scale_power`.

`BaseLoss` is deliberately untouched: it applies `1/sigma^2` to an elementwise MSE-style loss, where the quadratic weight is correct and is documented as approximating normalized-space training.

`example-v2026.2.yml` needed no change — it has no `metrics` section, so it picks up the `load_metric` default of `var_weighting: "none"`.

### F6 — `base_metric_overrides` comment removed

Dropped rather than implemented. The concept doesn't map onto `BaseCombinedMetric`: it holds N metrics, each scoring every variable under its own namespace, so there is no single "base" metric for a per-variable entry to override. (It *is* well defined for `BaseLoss`, which has exactly one base loss — hence `base_loss_overrides` there.) Implementing it would also have meant reporting, say, an MAE value under an `rmse/...` key, and would have conflicted with `scale_power` since the override changes the score's units.

Replaced with a note pointing at the actual mechanism: add the metric to the list and it gets its own namespace (`mae/ERA5/...`).

### Tests

Per-class `scale_power` values; RMSE weighted by `1/sigma` and MSE by `1/sigma^2` checked against explicit formulas; R² aggregating identically under `inverse_variance` and `none`; and a dimensionless metric constructing without `scaler_path`.

The weighting tests needed a new moderate-variance scaler fixture. My first attempt reused the existing fixture, whose variances span 14 orders of magnitude — after `normalize_weights` one variable dominates the aggregate and `1/sigma` vs `1/sigma^2` become numerically indistinguishable, so the test passed even with `scale_power` reverted. With closely-spaced variances all four new tests fail against the unfixed code, which I verified.

472 tests pass locally; `ruff check` / `ruff format --check` clean; the edited YAML parses.

### Still open

**F2** — no Sphinx documentation for either framework, still the one unmet required checklist item — and the minor items: conda-logic triplication across `_common.py` / `pbs.py` / `slurm.py`, uninitialized `last_var_losses` / `last_var_scores`, the loop-variable use at `credit/losses/base.py:428`, and a tracking issue for learnable-variance checkpointing.

djgagne:
Pushed 0aceeb04 and 548b5cf5, which close **F2**.

### `Losses.md` split by generation (0aceeb04)

The existing page moved to `Losses_gen1.md` (via `git mv`, so history follows) and gained a banner explaining it documents the flat-key format, plus the rule `load_loss` actually uses to choose: presence of a `type` key.

The new `Losses.md` documents the Gen 2 `{type, args}` section and `BaseLoss`: why a per-variable physical-units loss exists, the target-twin postblocks it requires, the full `args` table, all four `var_weighting` modes, latitude weighting, per-variable overrides, per-variable logging, and custom losses via `@register_loss` + `custom_objects`. The `learnable` section flags that its parameters are not yet checkpointed.

Two details I verified rather than assumed:

- The postblock YAML is copied from the working `example-end-to-end.yml`, not the abbreviated docstring, so it includes the `exp_transform` mirroring that the docstring omits.
- I checked at runtime which registry losses are actually elementwise. `mse`, `mae`, `msle`, `huber`, `logcosh`, `xtanh`, `xsigmoid` qualify; `spectral` and `power` reduce to a scalar and `covmse` returns a mismatched shape, so all three fail `BaseLoss`'s shape check at the first forward pass. The page says so explicitly.

It also corrects the custom-loss instructions inherited by the Gen 1 page, which told readers to edit a `base_losses.py` that no longer exists and add to a hard-coded dict. Losses resolve through the registry now, so the example gains `@register_loss` and its `forward` signature is corrected to `(target, prediction)`.

### New `Metrics.md` (548b5cf5)

The user-guide counterpart for `credit.metrics`, which had no coverage outside docstrings. `Evaluation.md` is left alone — it covers offline verification of a completed rollout via `rollout_metrics.py`, which is a different thing from metrics computed during training; the two pages cross-link.

Covers the optional `metrics` section and its defaults, the shared target-twin postblock requirement, the eight built-in metrics with their `scale_power`, the output key format, `var_weighting`, the full option table, ACC/SDAF with all three climatology sources, and custom metrics via `@register_metric` + `custom_objects`.

### One thing worth knowing about the anchors

The section links I added in 0aceeb04 were silently dead — `myst_heading_anchors` was not set, so MyST generated no heading slugs at all. `Losses.md`'s own `#variable-weighting` link was broken and I did not catch it, because the warning was buried in the ~170 pre-existing AutoAPI warnings.

548b5cf5 sets `myst_heading_anchors = 3`, which fixes those and makes section cross-referencing available to every Markdown page. Turning it on also surfaced two pre-existing dead anchors in `Model_Architectures.md`'s model list — `#unet` and `#graph-transformer`, neither of which has a section — so those are now plain text rather than broken links.

Verified with a clean `make html`: build succeeds, no warnings on any edited page, no unresolved cross-references anywhere in the docs, and `Losses.html#required-postblocks` resolves from `Metrics.md`.

### Checklist status

With this, **Documentation has been updated to cover code changes and renders properly** can be checked. The remaining open items are all minor: conda-logic triplication across `_common.py` / `pbs.py` / `slurm.py`, uninitialized `last_var_losses` / `last_var_scores`, the loop-variable use at `credit/losses/base.py:428`, and a tracking issue for learnable-variance checkpointing.

The one required item I still can't verify myself is whether the GPU-affecting trainer changes have been run on Casper or Derecho.

djgagne:
Pushed ea8cf05f, closing the four minor items. Every finding from the review above is now resolved.

### Conda resolution de-duplicated

The `conda info --envs` awk lookup had been copied verbatim into `credit/cli/_common.py`, `credit/pbs.py`, and `credit/slurm.py`. It now lives once in a new `credit/conda_env.py`.

That module is deliberately dependency-free (stdlib only, no `credit` imports) rather than reusing `credit/cli/_common.py`'s copy, so the launchers do not have to import the CLI package to resolve an env path. `slurm.py` already imported `credit.cli._common`; `pbs.py` did not, and still does not.

The refactor is behavior-preserving — I checked the three call sites against the previous implementation for both bare names and full prefix paths, and the emitted strings are byte-identical:

```
${CONDA_PREFIX:-$(conda info --envs | awk -v n=credit '$1==n {print $NF; exit}')}/bin/torchrun
```

### `last_var_losses` / `last_var_scores` initialized

Both are documented as attributes but were only created in `forward()`, so reading either before the first pass raised `AttributeError`. Latent rather than live — the trainer always calls the criterion before reading them — but the docstrings promised otherwise.

### Loop-variable device

`BaseLoss.forward`'s learnable branch read `p.device` after the loop that bound `p`. The device is now captured explicitly inside the loop, defaulting to the parameter's own device so the `.to()` is a no-op if the loop never runs. No behavior change on any reachable path.

### Learnable-variance checkpointing

Opened #473 with the impact, what needs doing (save/restore, the FSDP question, whether the validation criterion needs the same state, a round-trip test), and the workaround. Linked from both the `BaseLoss` docstring and the `learnable` section of `Losses.md`, so the limitation is discoverable from the code and the docs rather than living only in this PR's description.

### Tests

The three launchers agree with each other and with the shared helper, plus a guard that scans `credit/` and fails if the inline awk lookup reappears outside `conda_env.py` — the duplication is what caused the drift originally, so it seemed worth making regression-proof rather than just fixing. Both attribute-initialization checks added.

Full suite: **1405 passed, 33 skipped**. `ruff check` and `ruff format` clean.

### Where the checklist stands

| Item | Status |
|---|---|
| Clear description | ✅ |
| Issues tagged | ✅ |
| CI checks and tests pass | ✅ |
| Tests updated to cover changes | ✅ |
| Documentation updated | ✅ |
| Dependency lists updated | ✅ |
| Paths user generic | ✅ |
| GPU verified on Casper/Derecho | ❓ |

The last one is the only item I cannot verify — the `TrainerERA5Gen2` train/validate branches are GPU-path changes, so a run on Casper or Derecho is worth confirming before merge.

One thing I noticed but did not change, since it is outside every finding: `BaseLoss._build_static_weights` divides by `len(weights)`, so a config resolving to zero scored variables would raise `ZeroDivisionError` rather than a useful message. Not reachable through any normal config, and easy to add a guard for later if it ever surfaces.

kevinyang-cky (approved):
I made three light commits to fix the 1)ToDevice test failing on GPU machines, 2)add inference postblocks to the [config/gen_2/examples/example-end-to-end.yml‎](https://github.com/NCAR/miles-credit/pull/472/changes#diff-731d79504a4d2b213a1c964d7d8f8668fd8fddf05f8b900b0d795ff0c9336fac), and 3)make a minor fix to the docstring in [credit/registry.py](https://github.com/NCAR/miles-credit/pull/472/changes#diff-ef2164ba1177aa956675a606067ede3eaea08431b2d1fa387c5fcac268981900).

For 2), I believe this is the only solution to have a config that works for both train and rollout, but please confirm if adding a `postblocks` block under the `inference` section is the correct approach.

Could you add `to_device preblock` into `example-end-to-end.yml`?

Another comment/question that I have is: since the loss is now calculated in the physical space and scaled by `inverse_variance`, does that mean standard scaling is almost always required no matter what the scaling method the user wants to use? (e.g., in the case where the user wants to use minmax or quantile scaling)So running pre-processing at least twice?

I ran preprocess, train, and rollout with [config/gen_2/examples/example-end-to-end.yml‎](https://github.com/NCAR/miles-credit/pull/472/changes#diff-731d79504a4d2b213a1c964d7d8f8668fd8fddf05f8b900b0d795ff0c9336fac) and all passed.

If someone also wants to do a review, then merging can wait; otherwise, I approve the PR.

credit/applications/train_gen2.py
```diff
@@ -80,21 +76,22 @@ def main_cli():
...
         sys.exit(1)
 
     assert "source" in conf["data"], (
         "train.py requires the Gen2 nested data schema (conf['data']['source']). "
         "For the legacy flat schema, use applications/train.py."
     )
-    loss_name = conf["loss"]["training_loss"]
+    # Loss sections are either legacy flat keys (training_loss: ...) or the
+    # new-style {type, args} structure (mirrors preblocks/postblocks).
+    loss_name = conf["loss"].get("training_loss") or conf["loss"].get("type")
```
jsschreck, line 94: Remind me are wer preserving gen1 here? Do you want to kill 'type'?

credit/applications/train_gen2.py
```diff
@@ -235,9 +222,19 @@ def main_cli():
 
     train_criterion = load_loss(conf)
     valid_criterion = load_loss(conf, validation=True)
-    from credit.metrics import LatWeightedMetrics
 
-    metrics = LatWeightedMetrics(conf)
+    # BaseLoss with var_weighting="learnable" owns trainable per-variable
+    # log-variance parameters — they must join the optimizer explicitly.
+    # (Checkpointing of these parameters is not wired; they re-init on resume.)
+    from credit.losses import BaseLoss
```
jsschreck, line 229: Are both imports here supposed to be local or meant to be global (at the top of the file)?

credit/applications/train_gen2.py
```diff
@@ -10,28 +10,25 @@
 For the legacy flat schema (v1), use ``applications/train.py``.
```
jsschreck, line 10: We have four train scripts here. Three of them are gen1, correct? If so, do we need to continue to incrementally update them? I would favor (in a future PR) getting them down into one script (if that is possible) so we only need to edit one going forward for gen2.

credit/cli/_submit.py
```diff
@@ -478,6 +478,61 @@ def _scheduler_ctx(args: argparse.Namespace) -> dict:
     }
 
 
+def _derecho_nodes(args: argparse.Namespace) -> int:
+    """Return the requested node count, defaulting to 1 when unset."""
+    return int(getattr(args, "nodes", 1) or 1)
+
+
+def _use_pbsdsh(args: argparse.Namespace) -> bool:
```
jsschreck, line 512: Nice addition!

jsschreck (commented):
Mostly looks good to me, I note that many of the files changed here are just renaming them.

djgagne:
I'm going ahead and merging. Github actions is borked at the moment but tests pass on casper.

PR #475, opened 2026-08-06
WXFormer updates for gen2: CrossEmbedLayer padding fix, checkerboard-…

jsschreck:
…free decoder by default, wxformer_base alias

Fixes reported on credit.models.wxformer.crossformer.CrossFormer while testing cross_embed_strides=(1, 2, 2, 2) (entering the encoder at native resolution for the first stage):

- CrossEmbedLayer's padding=(kernel - stride) // 2 only gives a true "same" shape when (kernel - stride) is even. At stride=1 with an even kernel that's odd, so integer division drops a cell (kernel=4, stride=1: 480 in, 479 out). Every kernel branch in the multi-scale conv loses the same cell, so cat() still succeeds -- silently, at the wrong size. Fixed with explicit asymmetric zero-padding; stride=2 defaults are unaffected (kernel=4, stride=2, in=480 -> out=240 either way).

Also, since the checkerboard-artifact question came up alongside it: the PixelShuffle decoder (ICNR-initialized weights, zero-init bias -- Aitken et al. 2017, https://arxiv.org/abs/1707.02937; Lasagne/Lasagne#862) was already correctly implemented, but only reachable via upsample_with_ps=True, with the class-level default still the checkerboard-prone ConvTranspose2d path. Every current example/production config that sets this flag already sets it to True; none rely on False. Removed the ConvTranspose2d branch entirely rather than just flipping the default, since the goal is no longer having a checkerboard-prone option to reach for by accident. upsample_with_ps is kept as a no-op constructor argument so existing configs that pass it don't break. Verified at init: post-PixelShuffle intra-2x2-block variance is now exactly 0.0 (was effectively 100% of local variance with ConvTranspose2d, same measurement used on the cube-sphere decoder fix).

Also added "wxformer_base" as a registry alias for CrossFormer -- "wxformer" alone is ambiguous now that "nextgen_wxformer" also exists; "wxformer_base" names this as the original CrossFormer-backed architecture specifically. "wxformer" is kept working for backward compatibility.

Deliberately narrow scope: credit/models/crossformer.py (registry key "crossformer") is a separate, older fork of this same file with neither fix; left untouched pending a separate decision on consolidating it.

### Related Issues and PRs

Closes #XXX

### Checklist

### DIFF
--- a/credit/models/__init__.py
+++ b/credit/models/__init__.py
@@ -32,10 +32,20 @@ _MODEL_REGISTRY = {
         "UnetDiffusion",
         "Loading A DDPM model with UNET Backbone ...",
     ),
+    # "wxformer" is kept only for config backward compatibility -- prefer
+    # "wxformer_base" in new configs. "wxformer" alone is ambiguous now that
+    # "nextgen_wxformer" also exists; "wxformer_base" names this as the
+    # original CrossFormer-backed architecture specifically, not the whole
+    # WXFormer family.
     "wxformer": (
         "credit.models.wxformer.crossformer",
         "CrossFormer",
         "Loading the WXFormer deterministic model ...",
+    ),
+    "wxformer_base": (
+        "credit.models.wxformer.crossformer",
+        "CrossFormer",
+        "Loading WXFormer base (CrossFormer U-Net backbone) ...",
     ),
     "crossformer-ensemble": (
         "credit.models.wxformer.crossformer_ensemble",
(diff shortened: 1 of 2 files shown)


djgagne (approved):
Thank you for submitting the changes. Shoutout to @ILoveSorasakiHina for bringing up these issues and if you see any inconsistencies with the fixes you've made, please submit another PR.

PR #476, opened 2026-08-08
Gen 2 User guide, support for mac training, and bug fixes

### LINKED ISSUE #397: Remove launch sections from gen 2 applications?
The gen 2 applications for training and rollout still have a section for launching a PBS job from the script. However, the new credit cli submit option seems to handle packaging and launching of jobs outside of the script, which would make the script-specific calls redundant. @jsschreck can you review both the credit submit interface and the train_gen2 scripts to see if we can remove the launch code from gen 2 apps?

### LINKED ISSUE #426: Installation guidance - recommend isolated installs
The new CREDIT guidance (lines below) to install packages in user space within a conda environment is likely to cause issue for folks (since it's no longer an isolated install w/ the packages in user space).  See here: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#using-pip-in-an-environment

https://github.com/NCAR/miles-credit/blob/595cf700f1fb415d035263687f032f92c153d327/docs/source/getting-started.md?plain=1#L13-L28

I suspect the motivation for this was to be able to provide folks an environment to build off of and then modify at the user level, but I think we might want to approach that differently.

Some options might be:
* having folks clone the starting environment and then edit that
* recommending everyone build their own environments from scratch

If the concern is about the time and space it takes to build new environments, we might look into something like uv or pixi.

### LINKED ISSUE #432: Gen2 Trainer (and other derivative functions) do not support `levels: null` for selecting all levels, despite LocalDataset supporting it

### LINKED ISSUE #447: User guide updates
Update user guide to cover Gen 2 structure and how new config works.

djgagne:
# THIS IS THE LAST PR BEFORE THE GEN 2 RELEASE

## Summary

This is a large branch that builds out the **Generation 2** stack across models, losses, metrics, datasets, and CLI tooling, removes deprecated Gen 1 code, and adds a substantial documentation set. It accumulates several already-reviewed feature merges (`g2loss`, `gfs`, `credit-solardataset`, `wxformer-updates-gen2`) plus new tooling and docs work.

**Scope:** 33 commits · 107 files · **+5,338 / −5,658** (vs `main`; earlier feature branches in this series were merged separately)

### New CLI tooling
- **`credit check`** (`credit/cli/_check.py`, ~1.1k lines) — static config validator: resolves every registry key, binds each block's `args` against the real constructor signature, cross-checks channel layout vs. model geometry, verifies the BaseLoss target-twin postblock chain, and existence-checks paths.
- **`credit begin`** (`credit/cli/_begin.py`, ~850 lines) — interactive config wizard for bootstrapping a run.

### Gen 2 loss framework
- New **`BaseLoss`** (`credit/losses/base.py`) with per-variable/univariate composition operating on postblock output in physical units.
- Gen 1 losses split into `credit/losses/gen_1/`; expanded loss registry (`crps`, `almost-fair-crps`, `ring-crps`, `covmse`, `spectral`, `msle`, `huber`, `power`, …).

### Gen 2 metrics framework
- **`BaseVariableMetric` / `BaseCombinedMetric`** (`credit/metrics/`), an anomaly module (ACC / forecast activity SDAF, `credit/metrics/anomaly.py`), and common area-weighted metrics. Gen 1 metrics moved to `gen_1/`.

### New datasets
- **GFS / GDAS** and **GEFS ensemble** Gen 2 datasets (`credit/datasets/gen_2/gfs.py`, `gefs.py`, with download helpers).
- **TISR** switched to the CMIP6 total-solar-irradiance table (covers 1850–2299).

### WXFormer / model updates
- CrossEmbedLayer boundary-padding fix, **checkerboard-free decoder by default** (pixel-shuffle / ICNR + optional bilinear `interp`), and a **`wxformer_base`** alias.
- Backward-compatible loading of pre-`ZeroPad2d`-wrapper CrossFormer checkpoints.

### Preprocess / preblocks
- **Preprocess now fits and saves *every* bridgescaler scaler preblock**, not just the first (supports mixing standard + quantile scaling on disjoint variables); rejects duplicate `scaler_path`s and backs up an existing scaler before overwriting.
- New `to_device` preblock.

### Distributed / HPC
- Recognize **fsdp2** in `get_rank_info` so ranks map to distinct GPUs.
- `credit submit` honors `--nodes` in every mode (not just `train`); adaptive Casper GPU/memory requests.

### Docs (large)
- New/expanded Gen 2 pages: **Datasets, Models, Preblocks, Postblocks, Losses, Metrics, CLI, Custom Objects**, plus `gen2_overview` and `quickstart`; toctree reorganized into Gen 2 / Training & Inference / API / Gen 1 sections; autoapi docstring warnings fixed.

### Removals / cleanup
- Deleted deprecated trainers (`credit/trainers/deprecated/*`), standalone `train*.py` / `model_summary.py` scripts, and `transforms/deprecated/_transforms.py` (~5k lines removed).
- Pinned the ruff rule set (`select = ["E4","E7","E9","F"]`) to prevent silent mass-rewrites on hook bumps; added `obstore` dependency.

## Breaking changes / things to watch
- Deprecated Gen 1 trainer/script modules were removed — anything importing them directly will break (use the `credit` CLI / Gen 2 trainers).
- Configs with **two bridgescaler scaler preblocks** must now give each a **distinct `scaler_path`** (`credit check` enforces this).
- Ruff `select` is now pinned; contributors on newer ruff versions will see fewer auto-fixes than the tool's default.

## Where reviewers should focus

Given the size, prioritize the **core runtime logic** over deletions:

1. **Gen 2 loss `BaseLoss` + the target-twin postblock chain** (`credit/losses/base.py`, and how `credit check` validates it) — this is central to training correctness.
2. **Gen 2 metrics framework** (`credit/metrics/base.py`, `anomaly.py`) — verify the area-weighting / climatology alignment and the `metric(full_data_dict)` calling convention.
3. **Preprocess multi-scaler change** (`credit/applications/preprocess.py`, `credit/preblock/__init__.py`) — fitting/saving multiple scalers and the distinct-`scaler_path` rule.
4. **WXFormer padding + checkpoint back-compat** (`credit/models/wxformer/`) — the CrossEmbedLayer padding fix and pre-wrapper checkpoint loading; confirm existing checkpoints still load.
5. **New datasets** (`gfs.py`, `gefs.py`, TISR CMIP6 table) — remote I/O paths, channel typing, and multi-step availability checks.
6. **fsdp2 rank mapping** (`credit/distributed.py`) — multi-GPU correctness.
7. **Overview docs** — make sure the new quickstart guide and overview docs make sense to a new user.

**Lower priority / skim:** deprecated-file deletions, and ruff/tooling config — mostly additive or mechanical.

## Post-review fixes

The [review](https://github.com/NCAR/miles-credit/pull/476#pullrequestreview-4887893534) findings are addressed in the last three commits:

- **`select_device` no longer flips `cudnn.benchmark` on by default** — `BaseTrainer.__init__` runs after `seed_everything`, so the old side effect silently undid determinism; `credit preprocess` opts back in, and explicit `--device cuda:N` calls `set_device` again.
- **`credit begin` correctness**: WB2 TISR grid specs corrected against the *actual* store coordinates (the regridded 240x121/64x32 stores use ascending latitude — the old specs flipped the solar channel north–south besides the spacing drift); local datasets now read real vertical-coordinate values instead of fabricating `1..N` IDs that `method="nearest"` collapses onto one level; the diagnostic/static prompts are actually disableable and diagnostic variables flow into `output_only_channels`; anchored `dec\d+`/`nid\d+` hostname matching; added `credit/cli/__main__.py` so the wizard's `python -m credit.cli check` self-check works.
- **Scaler-path dedup** normalizes with `expanduser`+`realpath` in both `credit check` and preprocess.
- **New `tests/test_cli_begin.py`** (24 tests): every wizard preset round-trips through the YAML dumper and passes `credit check`; grid-spec truth table locked in.
- **Docs**: quickstart now includes the missing `credit preprocess` step and a laptop `credit train` path; Training.md's dead config references replaced with real files; gen2 `forecast_len` 1-indexing stated; every `dataset_type` registry key documented (incl. new GFS/GEFS sections); `wxformer_base` alias direction corrected; Inference.md `--rollout`/`--jobs` claims fixed; plus ~20 smaller clarity fixes from a new-user read-through.
- **`arco_era5_wxformer.yml` example repaired** — fixing its invalid `dataset_type: "arcoera5"` key unmasked six `credit check` errors (`levels: []`, flat `dim` pyramid, `detach` default, invalid postblock scaler args, missing `y_target_processed` twin chain); it now validates clean. Added `arco_era5_wxformer_derecho.yml`, a single-node quick-test variant validated on Derecho.

## Testing
- New test suites for `credit check` (`tests/test_cli_check.py`), the loss framework (`tests/test_losses_base.py`), the metrics framework (`tests/test_metrics_base.py`), and the `credit begin` wizard (`tests/test_cli_begin.py`), plus preprocess multi-scaler coverage.
- Full local run: 1589 passed / 33 skipped / 0 failed. CI runs `pytest` on Python 3.11/3.12/3.13 (CPU-only torch). Note `tests/manual/` (multi-GPU/HPC) and `tests/CI_CIRRUS/` (self-hosted GPU smoke) are not part of the standard run.
- Docs verified with `sphinx-build` (no new warnings on the added pages).
- GPU/HPC validation: the ARCO WXFormer quick-test config was run on Derecho.

## Related Issues and PRs

Closes #397 
Closes #426 
Closes #432
Closes #447 

### Checklist


djgagne (commented):
# Review — Gen 2 User Guide, mac training support, and bug fixes

**Verdict: approve with requested changes.** CI is fully green (pytest on 3.11/3.12/3.13, docs build, pre-commit, build job), the full suite passes locally (1,565 passed / 0 failed), and the core runtime changes are sound — the legacy-checkpoint compat work and the fsdp2 rank-mapping fix are genuinely well done. But there are four should-fix code issues (two of which produce *silently wrong science* from wizard-generated configs), the new 854-line `credit begin` has zero tests, and the docs — while individually strong — have a broken getting-started path for anyone not on Casper/Derecho. Housekeeping: the PR description says 161 files / +14,392, but the actual diff vs `main` is 101 files / +4,425 / −5,625 — worth refreshing before merge since the title feeds release notes.

## Part 1 — Review checklist

| Checklist item | Verdict | Evidence |
|---|---|---|
| Clear description, issues tagged | ✅ Pass | Thorough description; closes #397, #426, #432, #447. Stats are stale (see above). |
| All CI checks pass | ✅ Pass | All 6 checks green, including Read the Docs. |
| Tests updated for code changes | ⚠️ Partial | See below. |
| Docs updated and render properly | ⚠️ Partial | RTD builds clean, but see Part 3 for content gaps and three dead file references. |
| Dependency lists updated | ✅ Pass | No dep-file changes needed: `obstore`, `bridgescaler`, `gcsfs` all already declared; the `xesmf` import in `data_server.py` was made *optional* (try/except) — an improvement. |
| User-generic paths | ✅ Pass | New example config uses `$USER`; the PR *removes* hard-coded personal paths from `example-end-to-end.yml`. One borderline: `docs/source/Datasets.md:87-101` adds `/glade/campaign/cisl/aiml/ksha/...` — technically outside the sanctioned `/glade/campaign/aiml/credit/`, but it matches the existing convention in 10+ configs on main. Reviewer's-judgment call whether to migrate the dataset reference now. |
| GPU functionality verified on Casper/Derecho | ❓ Please confirm | The fsdp2 rank-mapping fix and `select_device` changes touch multi-GPU paths; please confirm these ran on Casper/Derecho and note it here. |

**Test coverage detail.** Good coverage where it matters most: `tests/test_legacy_checkpoint_compat.py` (222 lines) genuinely exercises the compat path — legacy key layout with `strict=True`, forward-output equality, the FSDP2 up-front migration, idempotency, and protection against migrating sibling architectures. The multi-scaler preprocess, `_check_scaler_paths_unique`, channel-layout resolution, and preflight fallback all have matching tests. Gaps:

- **`credit/cli/_begin.py` — 854 new lines, zero tests.** The pure helpers (`_build_config`, `_padding_totals`, `_pbs_config`, `_make_data`) are unit-testable without any interactivity; ideally pipe a canned answer sequence through `_collect_state` and run `credit check` on the emitted config. Findings 2–4 below all live in this file, so this is the highest-value test to add.
- The `era5.py` gcsfs→obstore swap and the fsdp2 branch of `get_rank_info` are untested (both hard to unit-test; noting for the record).

## Part 2 — Code findings (verified against the diff; most severe first)

**Should-fix:**

1. **`credit/distributed.py:56-60` + `credit/trainers/base_trainer.py:237` — `select_device()` silently undoes `seed_everything`'s determinism.** It sets `torch.backends.cudnn.benchmark = True` as a side effect, and `BaseTrainer.__init__` now calls it *after* `train_gen2.py:212` has run `seed_everything` (which sets `benchmark = False` and `use_deterministic_algorithms(True)`). Every CUDA training run now gets nondeterministic cuDNN autotuning while still claiming determinism. Same ordering in `train_gen1.py`. Suggest a `set_benchmark: bool = False` parameter, or don't touch cudnn state in a device-selection helper at all.

2. **`credit/cli/_begin.py:36-42` — wrong TISR grid specs for two WeatherBench2 resolutions.** The specs feed `torch.linspace` with inclusive endpoints (`tisr.py:369-373`): `"240x121"` lon `[0, 359, 240]` gives 1.5021° spacing vs the real 1.5° grid (0.5° drift at the dateline), and `"64x32"` lon `[0, 360, 64]` duplicates the 0°/360° meridian while the real 5.625° grid ends at 354.375°, with the lat axis also wrongly pole-inclusive. The SOLAR forcing channel is then misaligned with the ERA5 channels by up to half a cell — silently, because grid *sizes* match so neither `credit check` nor the trainer catches it. The `1440x721`/`full` specs are exact.

3. **`credit/cli/_begin.py:321-322` — fabricated level IDs + `method="nearest"` = N copies of one level.** For local datasets the wizard asks for a level *count* and writes `levels: [1..n]`, but `LocalDataset` selects with `da.sel(..., method="nearest")` (`local.py:658`). On a file with a physical vertical coordinate (e.g. pressure 100–1000 hPa), every fabricated level 1…18 snaps to 100 hPa — all model levels contain identical data, with no error anywhere (`credit check` passes; training runs). The wizard already globs and opens the data directory; it should read the actual coordinate values instead.

4. **`credit/cli/_begin.py:316-317, 409-430` — the "blank disables" prompts can't actually be blanked, and the diagnostic answer is discarded.** `_prompt` returns the default on empty input and `_glob_suggestions` never yields an empty default, so disabling diagnostics/statics is impossible; the diagnostic path the user enters is never used (`preset.get("diagnostic")` is never set for any preset); and the `elif static:` fallback would emit a static group without a `path`, which `LocalDataset` rejects.

**Nits:**

5. `select_device` with an explicit device skips `torch.cuda.set_device`, so `credit preprocess --device cuda:1` leaves the current device at `cuda:0` (the code it replaced did set it). Low impact since the pipeline mostly uses explicit `.to(device)`.
6. Duplicate-`scaler_path` detection uses only `expandvars` — `~/x.json` vs `$HOME/x.json` vs relative spellings of the same file evade both the static check (`_check.py:500`) and the runtime guard (`preprocess.py:250`); add `expanduser` + `realpath`.
7. `_detect_system`'s hostname substring match on `"dec"` will classify any hostname containing "dec" (e.g. `decode-box`) as NCAR and demand PBS settings on a laptop; and `_run_check`'s primary `python -m credit.cli` invocation always fails (no `__main__.py`), so every check runs through the subprocess fallback.

**Done well (verified, not just skimmed):**

- The legacy-checkpoint migration is a model of backward compat: structural detection by walking real `CrossEmbedLayer` instances rather than regex-guessing key names, idempotent rewrites, and a loud, actionable error for the unmigratable decoder case instead of a silently random-initialized decoder — with tests covering the FSDP2 path where module hooks fire too late.
- The fsdp2 addition to `DISTRIBUTED_MODES` fixes a real all-ranks-on-`cuda:0` bug, and the constant's comment documenting that exact failure mode turns a recurring bug class into an enforced invariant.
- `credit begin` closes its own loop by running `credit check` on the config it writes — all four ERA5/WB2 presets it generates pass the checker cleanly in headless runs, including the target-twin postblock chain and model-geometry cross-checks.
- The multi-scaler preprocess claim that "preblocks never mutate the caller's batch" was verified across `scaler.py`, `log.py`, `norm.py`, `regrid.py` and holds; duplicate-path rejection is enforced both statically and at runtime, with tests.

## Part 3 — Docs review (persona: meteorology grad student, new to CREDIT)

**Overall:** the gen2 pages are individually excellent — annotated YAML with "why" comments, honest failure-mode documentation, and `credit check` woven through everything. But the *getting-started path* fails for anyone off NCAR machines, and has one hard stop even for NCAR users.

**Can a new user get from `pip install` to a trained model? No — three blocking gaps:**

1. **The Quickstart never mentions `credit preprocess`.** It jumps from "generate a config" straight to "submit a training job," but `cli.md:97` says the scaler "must run once before training," and `credit begin` itself prints `credit preprocess -c <config>` as a next step. A user following the Quickstart literally trains against a scaler JSON that doesn't exist. Single most important docs fix.
2. **No laptop training path.** Quickstart step 3 is only `credit submit --cluster {casper,derecho}`; plain `credit train -c config.yml` appears only in cli.md. For a PR whose title includes "support for mac training," the getting-started flow never shows a non-PBS training command. The laptop-friendly `weatherbench2_era5_wxformer_tiny.yml` config that already exists in the repo is never referenced anywhere.
3. **Broken references:** Training.md's "Available configs" table lists `config/wxformer_1dg_6hr_v2.yml`, `config/wxformer_025deg_6hr_v2.yml`, and `config/starter_v2.yml` — none exist (verified). The Quickstart's "More detail" link points to the **gen1** `config.md`, which teaches `forecast_len: 0` = single step — the opposite of gen2's 1-indexing, and **no gen2 page anywhere states the 1-indexing convention**. That off-by-one is the most likely silent misconfiguration for a migrating user.

**Per-page issues (selected):**

- **Datasets.md:** copy-pasteable typos in the ARCO example — `specific_humitidy` and `v_component of wind` (missing underscore) at ~line 120. The intro promises example configs for every dataset, but WeatherBench2, TISR, GOES, MRMS, and HRRR sections have none, and no section states its `dataset_type` registry key (they live in `multi_source.py:65-78`; `gfs`/`gefs` are entirely undocumented). `temporal_mode` documents only `"persist"` but the code also supports `"cyclic"`. Also: `DataSets.md` → `Datasets.md` is a case-only rename that will bite macOS/Windows checkouts, and `CLAUDE.md` still cites the old name.
- **Models_gen2.md:** best page in the set, but it says `wxformer` is the type with alias `wxformer_base` — the source says the opposite (`wxformer` is the back-compat alias; prefer `wxformer_base`, `credit/models/__init__.py:35-45`). "ICNR" is never expanded. The page twice says "the examples show the arithmetic in comments" but its own example has no such arithmetic and never links `example-v2026.2.yml`.
- **Preblocks.md:** never says what *bridgescaler* is (an external library?) — a newcomer's first contact is the config block name. Whether `$SCRATCH`-style env expansion works in config paths is never stated (an example uses it). The `era5_normalizer` (mean/std) vs `bridgescaler_transform` (scaler_path) fork — when to use which — is never explained; meanwhile the Quickstart's troubleshooting table says "check `mean_path`/`std_path`" when wizard-generated gen2 configs use `scaler_path`.
- **Inference.md:** leads with gen1 machinery and contains literal personal paths (`/Users/dgagne/data/wxformer_6h_test/`); it claims `--jobs N` "splits init times across N" while the parser help admits N>1 submits N redundant full-range jobs; and it references a `--rollout` flag that doesn't exist (it's `--mode rollout`).
- **Training_gen1.md:** curly typographic quotes in the YAML (`mem: ‘480GB’`, `queue: ‘main’`) that become part of the value and produce an invalid PBS spec on copy-paste; and the claim that `credit_train` "generates a PBS script and submits via qsub" is wrong for a train entrypoint. Most of the useful `pbs:`/`credit submit` reference content (memory scaling table, resolution order) is stranded in the Generation 1 section where gen2 readers won't find it.
- **installation.md:** references `create_env_derecho.sh` (doesn't exist; it's `create_derecho_env.sh`), and `credit_train config/example-v2026.1.0.yml` is missing `-c`. It never says "most users should use the Quickstart install instead" before the build-PyTorch-from-source path.
- **quickstart.md:** orphaned `:::` fence at line 61 will render literally.

**What the docs do well:** the Losses page's "three errors to recognize," the tensor-parallel `NotImplementedError` explanation with issue links, Metrics.md's `scale_power` treatment (turns a dimensional-analysis trap into something a first-timer can follow), the target-twin worked example in postblocks_gen2, and the Quickstart's healthy-training heuristics ("train_loss ≈ 1–3 after epoch 1") are exactly the operational wisdom new users lack.

**Top 5 docs fixes by impact:**
1. Add `credit preprocess` to the Quickstart between config and training, plus a laptop `credit train` alternative to `credit submit`.
2. Fix the three dead config paths in Training.md and link `config/gen_2/examples/example-v2026.2.yml` from the pages that promise it.
3. State gen2's `forecast_len` 1-indexing explicitly in Datasets.md and stop linking the Quickstart to the gen1 config page.
4. Add a complete laptop-runnable WB2 example (the tiny config already exists) and document every `dataset_type` registry key.
5. Reconcile Inference.md's gen1 content and its `--jobs`/`--rollout` claims with actual `credit submit --mode rollout` behavior.

---

**Bottom line:** merge-worthy after fixing findings 1–4 (the cudnn-benchmark regression and the three `_begin.py` correctness issues — the TISR grid one especially, since it silently corrupts the solar forcing for two advertised resolutions), adding at least a smoke test for `credit begin`'s pure helpers, and patching the docs quickstart path (preprocess step + laptop training + dead links). Since this is the last PR before the Gen 2 release, the docs fixes matter more than usual — the quickstart is the release's front door.

kafitzgerald:
@djgagne, is there anything in particular it'd be useful for me to look over?  I'm a bit light on time at the moment.

djgagne:
@kafitzgerald read through new docs and let me know if there are ambiguities or big missing pieces?

charlie-becker:
@kafitzgerald @djgagne 

I'm also, just starting to dig into this.

> read through new docs and let me know if there are ambiguities or big missing pieces?

The first thing that popped out to me, was the "how to create your own Dataset" is for Gen 1 and I don't see a Gen 2 eqivilent.

charlie-becker (changes requested):
A lot of great stuff in here. I have tested `config/gen_2/examples/example-end-to-end.yml` successfully, out-of-the-box with preprocess, train, and rollout. The new loss structure is fabulous. I also really like the yaml arg parsing options you added in the configs to slim them down a little. 

I also built the new docs and spent an hour moving through it. There are a couple of incorrect explanations. I'll also provide some comments for additional content and readability. That said, I think it's in really good shape. 

### Needs fixing 

1. https://github.com/NCAR/miles-credit/blob/6376253079ce8adfefe378f43fb1f015c48c1f54/docs/source/Losses.md?plain=1#L277-L283

I took a peek at `training_log.csv` and it seemed to only log the combined metrics, not per-variable as described above. 

2. https://github.com/NCAR/miles-credit/blob/6376253079ce8adfefe378f43fb1f015c48c1f54/docs/source/Inference.md?plain=1#L37-L50

This is likely the biggest error within the documentation. It demonstrates our old `predict:` key structure and not the new `inference:` key structure, which has changed a lot. It would also be good to provide some usage descriptions for all the other post-processing keys such as: `output_interval:`, `group_by:`, etc.

3. https://github.com/NCAR/miles-credit/blob/6376253079ce8adfefe378f43fb1f015c48c1f54/docs/source/Datasets.md?plain=1#L62-L64

This is not correct. The prediction / target will always only contain 1-time step. It's specifically for how many rollout steps will be performed in a sequential manner.

4. https://github.com/NCAR/miles-credit/blob/6376253079ce8adfefe378f43fb1f015c48c1f54/docs/source/quickstart.md?plain=1#L208-L209

Is this still accurate given loss in physical space?

### Organization / Additions / Nice-to-have

1. I feel like the docs readibility caters to people already familiar with CREDIT gen 1. It might be worth having the landing page (or another page) on "What is CREDIT", in addition to what is CREDIT Gen 2?

2. The Loss documentation seems like it's the most fleshed out (which makes sense as that was the original target of this PR). It might be worth expanding some of the other sections to be a bit more inline with this section. Specifically, having a section on adding your own dataset and preblock / postblock (which are entirely missing), similar to how the loss page is structured.

3. Some of the individual pages are a little lengthy and don't have drop downs. It would be nice is there was finer granularity on the sidebar drop-downs, especially for specific preblocks and postblocks.

4. Explicit declaration of the requirement for the concat preblock and reconstruct postblock would be useful (similar to how the target_reconstruct is declared as mandatory in some situations).

5. I think the CLI tab is awkwardly placed in the drop-down (near the bottom). This should be up higher, if not at the top.

6. Some config naming schemes could be more explicit in the docs. For example what's the human naming convention and what's mapped to a registry.

7. Bridgescaler should have more pointers to the package. It does in some places, but should be included in the preprocess section of the quickstart too. 

8. I would recommend in the quick-start, potentially pointing people to the generic `config/gen_2/examples/example-end-to-end.yml` running the full sequence:

`credit preprocess`
`credit train`
 `credit rollout`

To familiarize new folks with an example that runs in a matter of a couple of minutes just for demonstration purposes.

jsschreck (changes requested):
I mostly was looking at the two new CLI options and wxformer changes. Two minor things overall, I would say neither is blocking:
  - _padding_totals() reimplements the CrossEmbedLayer output-size formula rather than importing it from crossformer.py, so if that formula changes again the two could quietly drift out of sync. Not currently wrong, it happens to use even (kernel - stride) everywhere it's called, so the asymmetric-padding fix doesn't change its answer but it's a duplicated-logic risk worth a comment or a shared helper.
  - In _pbs_config, an unrecognized Casper GPU type typed at the prompt is silently dropped rather than flagged. Its a minor UX gap, not a correctness issue
  - All of the wxformer updates post my changes (Dian's catch) all looks good!

ahijevyc:
1. The valididation time range overlaps the training time range in `config/gen_2/examples/example-v2026.2.yml` and `config/example-v2026.2.yml`.
2. Why are there two `example-v2026.2.yml`'s? I don't think they are identical.

docs/source/quickstart.md
```diff
@@ -1,99 +1,131 @@
...
+credit check -c my_experiment.yml --strict   # exit non-zero on warnings too
+credit check -c my_experiment.yml --json     # machine-readable output
+```
+
+Note that gen2 `forecast_len` is 1-indexed: `forecast_len: 1` means a
+single-step prediction, unlike gen1's 0-indexed convention where `0` meant a
+single step.
+
+> **More detail**: the fully annotated gen2 reference config
+> [`config/gen_2/examples/example-v2026.2.yml`](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/examples/example-v2026.2.yml)
```
kevinyang-cky: Suggest pointing to `example-end-to-end.yml` instead.

docs/source/Datasets.md
```diff
@@ -0,0 +1,223 @@
...
+  timestep: "6h"
+  history_len: 1
+  forecast_len: 1
+```
+
+### TISRDataset
+*API reference: {py:class}`credit.datasets.gen_2.tisr.TISRDataset`* · `dataset_type: tisr`
+
+This dataset calculates total integrated top of atmosphere solar irradiance entirely in Pytorch based on lat-lon and 
+time information. It follows the design patterns and calculations of the [Graphcast solar radiation module](https://github.com/google-deepmind/weathernext/blob/main/graphcast/solar_radiation.py).
```
kevinyang-cky: The link is broken; use this one instead.

https://github.com/google-deepmind/weathernext/blob/08cf73625c9d12bd9aaa038868bcb2fe488f2a22/graphcast/solar_radiation.py

docs/source/Preblocks.md
```diff
@@ -0,0 +1,335 @@
...
+`bridgescaler_transform` for new gen2 work; `era5_normalizer` mainly exists to
+reuse existing gen1 statistics files.
+
+Normalizes per-variable ERA5 tensors using pre-computed mean/std NetCDF files.
+Normalization `(x - mean) / std` is applied per variable; variables not found
+in the statistics file are passed through unchanged. The mean/std may be scalar
+(2D variables) or 1-D per-level vectors.
+
+```yaml
+type: "era5_normalizer"
```
kevinyang-cky, line 132: Are we keeping this? (I thought this was more of a Gen1.5?)

docs/source/Preblocks.md
```diff
@@ -0,0 +1,335 @@
...
+
+```yaml
+# Scale specific variables in both input and target
+type: "bridgescaler_transform"
+args:
+  scaler_path: "/path/to/scaler.json"
+  variables:
+    - "era5/prognostic/3d/T"
+    - "era5/prognostic/3d/U"
+  method: "transform"
```
kevinyang-cky, line 175: A common pitfall is forgetting to set `channels_last=False` (Bridgescaler's default is `channels_last=True`)

docs/source/Preblocks.md
```diff
@@ -0,0 +1,335 @@
...
+
+**AutoAPI:** {py:obj}`credit.preblock.concat.ConcatToTensor`
+
+End-of-chain preblock that collapses a nested batch dict of tensors into a
+single input tensor (and optionally a target tensor), concatenating along the
+channel dimension. Input tensors are sorted by a canonical channel key derived
+from the `{source}/{field_type}/{dim}/{varname}` structure so the channel
+order is deterministic regardless of insertion order. In addition to the
+tensors, channel maps are attached to metadata mapping each variable to its
+slice in the concatenated tensor.
```
kevinyang-cky, line 246: Should mention that this preblock is always required.

kevinyang-cky (changes requested):
I just have several minor comments and the docs overall looking great! Another suggestion that I have is to reorder the sections to Datasets -> Preblock/Postblocks -> Model -> Loss -> Metrics following how the different sections are typically organize in the configs. But this is rather minor.

kafitzgerald (commented):
I haven't had time to look through this in any depth, but in reviewing the docs briefly I noticed a few organizational things that might trip up folks who are new to CREDIT:
 
- Without context, the Gen 1 vs Gen 2 language is a bit confusing.  It could be helpful to clarify this a bit and perhaps an explainer section or note about this somewhere prominent.
- If the CLI is the primary way to run CREDIT, I think that should come much sooner in the docs perhaps as its own section w/ configuration details and component info later on.
- The Get Started / Quickstart section has quite a lot of content.  I'd recommend stripping this down a bit and linking out to other sections more.  
- It may also be helpful to separate information for running on different systems under separate tabs for clarity and to keep things more concise.  Here's some documentation regarding how to do this: https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/web-components.html#tabs

Sorry, I wish I had more time to spend on this, but hopefully these comments are useful.

docs/source/quickstart.md
djgagne: Done — the quickstart's "More detail" callout now leads with `config/gen_2/examples/example-end-to-end.yml` (noting it runs the full `credit preprocess` → `credit train` → `credit rollout` sequence out of the box), with `example-v2026.2.yml` kept as the fully annotated reference.

docs/source/Datasets.md
djgagne: Fixed — the link now points at the SHA-pinned weathernext `graphcast/solar_radiation.py` you suggested, so it can't break again when `main` moves.

docs/source/quickstart.md
djgagne: Done in 9687b6e7 — the quickstart's "More detail" callout now leads with `config/gen_2/examples/example-end-to-end.yml` (noting it runs the full `credit preprocess` → `credit train` → `credit rollout` sequence out of the box), with `example-v2026.2.yml` kept as the fully annotated reference.

docs/source/Datasets.md
djgagne: Fixed in 9687b6e7 — the link now points at the SHA-pinned weathernext `graphcast/solar_radiation.py` you suggested, so it can't break again when `main` moves.

docs/source/Preblocks.md
djgagne: Keeping it for now, but the docs steer people away: the section opens with a "Which normalization route?" note recommending `bridgescaler_transform` for all new gen2 work and framing `era5_normalizer` as a compatibility path for reusing existing gen1 mean/std statistics files. Happy to deprecate it properly in a follow-up if we agree it should go.

docs/source/Preblocks.md
djgagne: Good catch — added a prominent warning box to the `bridgescaler_transform` section in 9687b6e7: bridgescaler defaults to `channels_last=True` while CREDIT tensors are channels-first, so any preblock that fits a new scaler must set `scaler_params: {channels_last: False}`. The first YAML example now includes it, and `example-v2026.2.yml` (the one bridgescaler example that was missing it) now sets it too (2db21fed).

docs/source/Preblocks.md
djgagne: Done in 9687b6e7 — the `concat` section now states it is always required at the end of every gen2 preblock chain, and cross-references its counterpart, the `reconstruct` postblock (whose doc now carries the matching "required as the first postblock" statement).

djgagne:
Review feedback is addressed in the latest three commits (2db21fed configs, 9687b6e7 docs, c0666976 code). Point-by-point:

## @charlie-becker

**1. Per-variable metrics in `training_log.csv`** — fixed in 646fae26: `trainer.save_metric_vars` now defaults to `True`, so per-variable `train_loss_var/…` / `valid_loss_var/…` and metric columns are written by default. Losses.md now documents the switch (`False`/list to suppress or filter).

**2. Inference.md shows the old `predict:` structure** — rewritten. The gen2 rollout section now documents the real `inference:` schema from `rollout_gen2.py`/`output_gen2.py`: `run_mode` (batch/single), `batch_forecast`/`single_forecast`, the optional `inference.data/preblocks/postblocks` overrides, the CLI overrides (`--run-mode`, `--init-time`, `--save-dir`), and every `output:` key (`format`, `output_interval`, `group_by`, `variables`, `metadata`, `encoding`, `max_pending`) with usage descriptions. Also removed a stale claim that `credit rollout` takes `--mode ddp` (it doesn't — mode comes from `inference.mode`), and noted that `credit realtime` still reads `predict:` (that's what the code does today).

**3. `forecast_len` explanation** — corrected in both Datasets.md and the quickstart: prediction/target always cover a single time step; `forecast_len` is the number of sequential autoregressive rollout steps per training sample (1-indexed in gen2).

**4. "train_loss ≈ 1–3 after epoch 1"** — no longer accurate, you're right. The quickstart now emphasizes trends (steady decrease, validation tracking training) and explains that the absolute magnitude depends on units and `var_weighting` (e.g. `inverse_variance` brings the initial loss to order 1). Also fixed the stale column list (`valid_loss`, per-variable columns; there is no `epoch_time_s`).

**Custom gen2 datasets** (your issue comment) — the gen2 recipe lives in the Custom Objects guide (subclass `credit.datasets.gen_2.base_dataset.BaseDataset`, register via `custom_objects:`), but nothing linked to it — added a "Writing Your Own Dataset" section at the end of Datasets.md pointing there, plus matching cross-links from the Preblocks and Postblocks pages.

**Nice-to-haves** — done in this pass: quickstart leads with `example-end-to-end.yml` and the full preprocess/train/rollout sequence (#8); concat/reconstruct requirement stated explicitly (#4); CLI moved up in the sidebar (#5); bridgescaler pointer added to the quickstart preprocess step (#7); "write your own X" cross-links (#2, partially — the full recipe lives in Custom Objects rather than duplicated per page). Deferred to follow-ups: a "What is CREDIT" landing page (#1), finer sidebar granularity (#3), and config naming-convention docs (#6).

## @jsschreck

**`_padding_totals()` duplication** — fixed properly rather than just commented (c0666976): `crossformer.py` now exposes `crossembed_pad_total()`/`crossembed_out_size()` next to `CrossEmbedLayer` (the layer derives its padding from the former), and `_padding_totals()` imports `crossembed_out_size` instead of reimplementing the formula — so the two can no longer drift. In the process the wizard picked up the asymmetric-padding fix (it had still been using the old symmetric formula — identical answers for the current even `(kernel - stride)` values, as you noted). The `tests/test_cli_begin.py` grid-spec truth table passes unchanged.

**Silent GPU-type drop in `_pbs_config`** — fixed: the prompt now re-asks on unrecognized input (same pattern as the parallelism prompt) and advertises the full valid set from `_CASPER_GPU_NODES` (`a100`, `a100_80gb`, `gp100`, `h100`, `l40`, `v100`), which had been partly hidden.

## @kevinyang-cky

Section reorder — done: the "Generation 2 Components" toctree now follows config order (Overview, Datasets, Preblocks, Postblocks, Models, Loss Functions, Verification Metrics, Custom Objects), and CLI moved up into Getting Started. Your inline comments are answered on their threads.

## @kafitzgerald

All four points were useful, thanks. Done in this pass: a "Gen 1 vs Gen 2: which one am I using?" explainer (comparison table + `credit convert` migration note) now sits near the top of the Gen 2 Overview page and is linked from the first paragraph of the quickstart; the CLI page moved up into the Getting Started section of the sidebar. Deferred to a follow-up: slimming the quickstart down further and per-system sphinx-design tabs (the tabs need a new `sphinx-design` dependency, so I'd rather do that as its own commit).

## @ahijevyc

**1. Train/validation overlap** — fixed (2db21fed): validation in `config/gen_2/examples/example-v2026.2.yml` is now 2019-01-01 → 2020-12-31 (training unchanged at 1979 → 2018), matching `example-end-to-end.yml`.

**2. Two `example-v2026.2.yml`s** — the top-level `config/example-v2026.2.yml` was a stale pre-refactor copy (old `predict:`/`data_valid:` schema). It's now a symlink to `config/gen_2/examples/example-v2026.2.yml`, so there is only one real file.

kevinyang-cky (approved):
Thanks for addressing my comments and LGTM!

charlie-becker:
@djgagne 

Thanks for making some changes. However, the "Running inference section" still contains numerous inconsistencies. 

> **2. Inference.md shows the old `predict:` structure** — rewritten. The gen2 rollout section now documents the real `inference:` schema from `rollout_gen2.py`/`output_gen2.py`: `run_mode` (batch/single), `batch_forecast`/`single_forecast`, the optional `inference.data/preblocks/postblocks` overrides, the CLI overrides (`--run-mode`, `--init-time`, `--save-dir`), and every `output:` key (`format`, `output_interval`, `group_by`, `variables`, `metadata`, `encoding`, `max_pending`) with usage descriptions. Also removed a stale claim that `credit rollout` takes `--mode ddp` (it doesn't — mode comes from `inference.mode`), and noted that `credit realtime` still reads `predict:` (that's what the code does today).

Everything BELOW the "**Running Rollouts with the v2 Data Schema**" subtitle within the "**Running Inference**" tab, seems to be correct (what the comment above describes). However, virtually everything BEFORE that, seems to be referring (mostly) to outdated and incorrect Gen1 structure (including the "ingredients for prediction"). As the quickest fix, I would recommend just deleting everything before the that subsection. We can add a bit of additional context later. 

Lastly, we may consider adding the "data" portion into the example yml within this section pointing to the GFS data, for example (showing that the inference section uses that data section for inference, but falls back to the "train" dataset if not present.

charlie-becker:
@djgagne fixed inference doc

charlie-becker (approved):
GTG

djgagne:
I made a few more doc changes and made sure ring CRPS works with the new gen 2 base loss. Should be ready to merge and release once @jsschreck approves.

jsschreck (approved):
It looks good, Im sure we will have more coming :D :D

djgagne:
Thank you everyone! Merging now.

PR #478, opened 2026-08-14
Cubed-sphere WXFormer: CubedWXFormer + WXFormerColumn

jsschreck:
## Summary

Adds a cubed-sphere (SE grid) variant of WXFormer alongside a renamed, cleaned-up
lat/lon "next-gen" variant. Both are research models, not yet replacing the v1
production WXFormer (`wxformer`, unchanged in this PR).

- **`CubedWXFormer`** (`cube_sphere_wxformer` / `cube_sphere_wxformer_next` collapsed
  into one class) — runs WXFormer natively on the cubed-sphere SE grid instead of
  lat/lon, with true ghost halo exchange across cube-face boundaries (real neighbor
  data, not lat/lon-style circular/polar padding — see `wxformer_family.md`), optional
  cross-face attention, and an opt-in `use_column_attn` flag that folds in level
  embeddings + column (vertical) attention + a spectral GNN bottleneck. The two
  previous classes (`CubeSphereWxFormer`, `CubeSphereWxFormerNext`) were collapsed into
  one to remove ~150 lines of duplicated forward-pass logic between them — verified
  **bit-identical** (`torch.equal`) against both prior classes in every mode (real
  ne120 static files, halo + edge + cross-face + global attention all enabled).
- **`WXFormerColumn`** (renamed from `NextGenWXFormer`/`nextgen_wxformer`) — the
  lat/lon-grid sibling with the same level-embedding/column-attention/spectral-bottleneck
  additions, pure rename only (internals byte-identical to the pre-rename file).

### Naming

`nextgen_wxformer` → `wxformer_column`, `cube_sphere_wxformer(_next)` → `cubed_wxformer`
(with `use_column_attn: true|false`). No deprecated aliases — nothing depends on the old
names yet, clean cutover.

### Halo exchange / `F.interpolate` audit

Per explicit request: audited every `F.interpolate` call in both models.
`cubed_wxformer.py` has none — halo/boundary fill goes entirely through `halo.py`'s real
gather map (points on the padded tile are inverted back to the sphere and resolved to
their true owning SE cell, not interpolated). `wxformer_column.py`'s calls are all
legitimate native-target-grid-shape correction (e.g. reconciling internal processing
resolution to ERA5's exact 721×1440-ish grid) plus one unrelated column-attention memory
optimization — none of them substitute for boundary handling.

### Also in this PR

- Fixed an inaccurate docstring/README claim that `CubedWXFormer` gets a native
  tensor-parallel plan "by inheritance" from `wxformer_column` — false, they're unrelated
  sibling classes (no subclassing) and `CubedWXFormer`'s encoder reuses `crossformer.py`'s
  conv-based blocks, which declare no `_tp_plan`. `supports_native_tp()` always returns
  `False` for it today.
- Merged current `main` in (this branch had drifted ~86 commits) — fixed real breakage
  surfaced by real end-to-end runs, not just a clean merge: mandatory
  `trainer.parallelism` block, `dataset_type` requirement, two-section preblocks format,
  the `reconstruct` postblock requirement, etc.
- Reference to Feng et al. 2026 ("Global-Regional AI Weather Forecasting on a Stretched
  Cubed Sphere," GRL, 10.1029/2026GL124199) added near the halo-exchange docs — related
  work tackling the same boundary-artifact problem from the mesh-design side.

## Verification

- Unit tests: `test_native_tp.py`, `test_cube_rope.py`, `test_cubed_wxformer.py`,
  `test_latlon_to_se_preblock.py` — 58/58 pass.
- Full `pytest` suite post-merge: 1562 passed, 8 failed + 3 errors — all pre-existing
  `obstore`/`s3fs` env gaps (HRRR remote / GFS / GEFS tests), unrelated to this PR,
  reproducible on unrelated files.
- **Real single-GPU end-to-end smoke tests on Derecho** (not just unit tests):
  - `config/gen_2/smoke/smoke_wxformer_column_derecho.yml` — local ERA5, full
    train+validate+checkpoint cycle. **PASS** (valid_rmse 0.38, valid_r2score 0.84).
  - `config/gen_2/smoke/smoke_cubed_wxformer_derecho.yml` — WeatherBench2 ERA5,
    lat/lon→SE regrid (ne120, 3.1M weights), `use_column_attn: true`. **PASS** through
    the full pipeline (regrid, model construction, forward/backward, optimizer step,
    checkpoint save). Uses a smoke-test-only scaler fixture (documented in-config) since
    the real WB2 example's fitted scaler predates a schema change brought in by the
    main merge — flagged as a follow-up, not blocking.

## Known follow-ups (not blocking this PR)

- `wxformer_cubesphere_next_wb2_post_scaler.json` (the real WB2 example config's fitted
  scaler) needs re-fitting via `credit preprocess` to match the postblock's current
  `{"input": ..., "target": ...}` schema.


ILoveSorasakiHina:
I'll take a look at this PR. There's quite a bit here, so I'll need some time. I'll also be testing on the CWA HPC — expect comments within the week.

jsschreck:
@ILoveSorasakiHina thanks for looking at it -- I am on PTO until week after next, so take your time!

ILoveSorasakiHina:
I have completed the code review and real-data testing for `WXFormerColumn`. The `CubedWXFormer` review is still in progress, and I will provide those results separately.

## `WXFormerColumn` test results

I tested the model using real 640×1280 ERA5 data with the following configuration:

```yaml
frames: 1
levels: 16
channels: 4
surface_channels: 7
input_only_channels: 3
output_only_channels: 0
col_attn_stride: 8
decoder_col_attn: false
```

This corresponds to:

- 74 input channels
- 71 output channels

Training completed normally, including data loading, forward and backward passes, optimizer steps, and checkpoint saving.

I also verified inference end to end:

- Checkpoint reload succeeded.
- A single 6-hour forecast succeeded.
- A 4-step/24-hour autoregressive rollout succeeded.
- Reconstruction, inverse scaling, and feedback preprocessing for the next step all worked.
- The model successfully wrote 640×1280 NetCDF output.
- The output structure correctly contained four 3D variables on 16 levels and seven 2D prognostic variables.
- The output contained no NaN or Inf values and was restored to reasonable physical units.

Therefore, for the primary use case I tested—`frames: 1` and `output_only_channels: 0`—I did not encounter any issue blocking training or inference.

## Residual correctness bug

I found one reliably reproducible correctness issue when both of the following are true:

```text
input_only_channels > 0
output_only_channels > 0
```

In this case, the residual for output-only diagnostic channels incorrectly uses input-only forcing/static channels.

The input and output channel layouts are:

```text
input  = [prognostic, input-only]
output = [prognostic, output-only]
```

However, the current residual is constructed by slicing `output_channels` directly from the input:

```python
x_res = x[
    :,
    last_frame_offset:last_frame_offset + self.output_channels,
]
```

As a result, the input-only channels following the prognostic channels are interpreted as residuals for the output-only diagnostic channels. If the number of output channels exceeds the number of input channels, only the remaining channels are zero-padded with `F.pad`. Consequently, this issue does not necessarily produce a shape error and may silently affect training and inference.

I confirmed this with the following minimal test:

```text
prognostic channels = 6
input-only channels = 1
output-only channels = 2

Cin  = 7
Cout = 8
```

I set every model parameter to zero so that the decoder-predicted delta was guaranteed to be zero. I then set the single input-only channel to the constant value `17`.

The correct diagnostic outputs should have been:

```text
[0, 0]
```

The actual result was:

```text
[17, 0]
```

The `assert_close` check failed with 2,048 mismatched elements out of 4,096. Since `2,048 = 32 × 64`, this corresponds exactly to the entire first diagnostic channel, confirming that it was contaminated by the input-only channel.

The residual should instead copy only the prognostic channels from the final input frame and explicitly use zero residuals for all output-only channels:

```text
x_res = [
    last-frame prognostic channels,
    zeros for output-only channels,
]
```

This bug does not affect my tested configuration because it uses `output_only_channels: 0`. It will, however, affect configurations that use both forcing/static inputs and diagnostic outputs.

The `CubedWXFormer` review and real-data testing are still in progress. I will provide a separate update after verifying the required ne120 static/regridding files and halo behavior.

ILoveSorasakiHina:
I have completed an initial functional review of `CubedWXFormer`. I used local 640×1280 Gaussian ERA5 data because the current HPC environment cannot access the WeatherBench2 GCS dataset.

I found the required files on Derecho. I generated the 640×1280↔721×1440 bilinear weights and composed them with the 721×1440↔ne120 weights for this smoke test. This is sufficient for functional testing but not scientific validation.

Test configuration:

- 4 3D prognostic variables × 16 levels
- 7 2D prognostic variables
- 1 dynamic forcing variable
- 2 static variables
- 74 input channels and 71 output channels
- `output_only_channels: 0`
- `use_column_attn: true`
- `col_attn_stride: 8`
- `decoder_col_attn: false`
- halo exchange, face-edge attention, and cross-face tile attention enabled

Across training and the workaround rollout described below, the following passed on real GPU hardware:

- 640×1280 → ne120 sparse preprocessing
- model construction
- forward, backward, and optimizer steps
- validation
- checkpoint save and reload (`All keys matched successfully`; 1,243 tensors)
- single-step DDP inference
- ne120 → 640×1280 reverse regridding
- 640×1280 NetCDF output

I found one issue blocking the documented Reconstruct-first native-grid rollout path.

After `Reconstruct`, each SE variable has shape:

```text
(B, levels, time, 1, ncol)
```

`SEToLatLonPostBlock._regrid_tensor()` preserves every leading dimension and appends `(nlat, nlon)`, producing:

```text
(B, levels, time, 1, nlat, nlon)
```

instead of:

```text
(B, levels, time, nlat, nlon)
```

The Gen2 NetCDF writer then fails with:

```text
ValueError: different number of dimensions on data and dims: 6 vs 4
```

This occurs with the postblock sequence declared in `config/gen_2/examples/wxformer_cubesphere_next_wb2.yml`:

```text
Reconstruct → inverse scaling → SEToLatLon
```

For continued testing, I used this configuration-only workaround:

```text
SEToLatLon(keys=[y_pred]) → Reconstruct → inverse scaling
```

Regridding the flat tensor first produces the correct `(B, C, T, H, W)` shape, and the single-step 640×1280 NetCDF rollout completes successfully.

I plan to repeat the test with a locally downloaded copy of the 721×1440 WeatherBench2 ERA5 dataset and the direct 721×1440↔ne120 weights. This will remove the composed 640×1280 interpolation, although the shape issue is independent of the source-grid weights.

ILoveSorasakiHina:
@jsschreck 

Preparing the native 721×1440 ERA5 data mentioned earlier would take more time, so I used my existing Gaussian ERA5 data for these functional checks. This covered the training, checkpoint reload, and multistep output workflows I wanted to test. I have not run the original example end to end with native 721×1440 data.

I used local 640×1280 Gaussian ERA5 data mapped to ne120 through composed weights, along with `se_index_ne120.npy` and `se_face_adjacency_ne120.npz` from John's directory. I completed the following:

- All 21 tests across three relevant test files passed, including the ghost-map ownership test with the real static files.
- With `dim=[64,128,256,512]` and `depth=[2,2,8,2]`, two-step autoregressive training and validation completed for 5 epochs using 8-GPU DDP. The config enabled column attention and global bottleneck attention, and the log confirmed that the face-boundary features were enabled.
- A separate single-step run successfully loaded a checkpoint and continued from epoch 5 through training, validation, and checkpoint saving.
- With adjusted inference postblocks, a four-step, 24-hour rollout completed for one initial condition. I checked all four NetCDF files: the valid times, the 640×1280 latitude/longitude coordinates and their ordering against the source data, the 16 levels, and the dimensions of all 11 variables. No NaN or Inf values were found.

These checks focused on whether training, reloading, and inference output work correctly.

Before merging, I suggest fixing the first two issues below and clarifying the intended behavior of overlapping corner updates.

1. **Both models should copy only prognostic channels into the residual.**

   The HPC runs above used `output_only_channels=0`, so they did not cover diagnostic outputs. I reproduced the issue below separately.

   Both `cubed_wxformer.py:659` and `wxformer_column.py:670` slice `output_channels` from the input for the residual, but only the first `channels * levels + surface_channels` channels are the matching prognostic state. The remaining input channels contain forcing/static fields and should not be used as the residual for output-only diagnostics.

   For example, `smoke_cubed_wxformer_derecho.yml` has 133 prognostic channels and 135 output channels. The two diagnostic outputs therefore receive residuals copied from `geopotential_at_surface` and `land_sea_mask`, respectively.

   In other configurations where `Cout > Cin`, `CubedWXFormer` raises a shape error because the residual has fewer channels than the output.

   Please update both models to copy only the prognostic channels from the last input timestep and use a zero residual for output-only channels. `test_zero_init_head_gives_persistence` currently builds its expected result with the same `:c_out` slice, so it treats this behavior as correct. That test should be updated alongside the fix and cover different input-only/output-only channel counts.

2. **The conversion back to the lat-lon grid still needs fixing.**

   This is the output issue I mentioned earlier. The example config `wxformer_cubesphere_next_wb2.yml` uses `prediction` and `target`, but `Reconstruct` produces `y_processed`. Because the names do not match, the conversion is silently skipped.

   After changing the setting to `keys=["y_processed"]`, the converted output has an extra dimension, which causes NetCDF writing to fail.

   NetCDF writing worked after I converted `y_pred` first, then ran `Reconstruct`. Please fix both the keys and the extra dimension so the original example also works, and add a test for this path.

3. **Is overwriting corner updates in cross-face attention intentional?**

   In `FaceEdgeAttention` (`edge_attn.py:349–350`) and `CrossFaceTileAttention` (`global_attn.py:481–482`), corner regions are updated by two face pairs. Both read the original input, so the second update overwrites the first.

   I confirmed on CPU that reversing the pair execution order changes the output only in these overlapping regions. Is keeping only the last update at corners the intended behavior?

jsschreck:
@ILoveSorasakiHina thank you a ton for these comments, I will get on this shortly!

AidanJanney:
@jsschreck I've been working on mesaclip for the workshop, but I have some relevant changes that I will open a pr into to your branch in the next couple days. Some of the changes echo @ILoveSorasakiHina 's comments above, and I also explore more artifact issues and some possible adjustments at the cubesphere boundary. I'm experimenting with some location embedding elements that may (not) help! 

Another note, I'm finding that local perturbations follow diffusive spreading (not physical, but to be expected) but longer range impacts of perturbations seem to follow cubesphere boundaries. I would be curious about how we could target against this type of behavior in training?

I would love to talk with you about combining grid output types too! Could we have an initial separated layer that processes CubeSphere and another grid type (e.g. rectilinear) to a shared latent space, then concats both tensors together and proceeds with the remaining encoding layers. Maybe we just want separate heads anyways and fine tuning together - it would be cool to have attention between the atm and ocn (on native grid)? I would love to figure out a way to train on a full stack of atm and ocn data at once without needing to regrid ocn data to the cubesphere grid because we lose a lot of fidelity.

(Got a bit carried away in the second half, and those notes won't be part of the review, just wanted to mention it here).

jsschreck:
@ILoveSorasakiHina pushed fixes for the first two, both with tests. Thanks for tracking these down so precisely, the repro steps made this fast.

**1. Residual copying wrong channels** — fixed in both models. They now build the residual from only the prognostic channels of the last input frame (`channels*levels + surface_channels`) and zero-pad the rest, so diagnostic outputs get a proper zero residual regardless of how input-only/output-only channel counts relate. `CubedWXFormer` no longer raises when Cout > Cin either. Updated `test_zero_init_head_gives_persistence` and added parametrized coverage across several input/output-only combinations (including Cout > Cin), plus a new test file for `WXFormerColumn`, which didn't have residual-specific coverage before.

**2. Lat-lon conversion** — root cause was actually in `Reconstruct`, not just the config. It folds a model's frame axis into the channel dim by checking `dim() == 5`, which assumes a 2D (H, W) grid. Cube-sphere models return `(B, C, 1, ncol)` — 4D, since the spatial dim is 1D — so that frame axis survived as a spurious extra dim through `y_processed` and then through the regrid. Generalized the check to fold the frame axis whenever it's a singleton at dim=2 regardless of total rank (a real spatial dim is never size 1, so this doesn't touch existing 2D-grid models). Fixed `FlattenToTensor`'s matching assumption too. Also fixed the config's `keys: [prediction, target]` → `keys: [y_processed]` (neither `prediction` nor `target` is ever written by `reconstruct`, so the regrid was a silent no-op before). Added tests for both the rank fix and the full `Reconstruct -> SEToLatLonPostBlock` path so the original example config now works as written, no reordering needed.

**3. Corner overwrite** — good catch, and no, I don't think last-write-wins was a deliberate choice, it's just what falls out of both pairs reading from the same pristine input and scattering independently. I don't want to guess at the right fix here (average the two updates? attend jointly across all faces touching a corner?) without thinking through what that does to gradient flow and to the already-tested boundary behavior, so I'd rather not rush a change into this PR. Let's take it up as a follow-up — @AidanJanney mentioned boundary/artifact work along these lines already in progress, might be a natural place for it to land together.

Pushed as 8458f851 and 4898ffd7 on this branch. Full suite still green (1576 passed, only pre-existing unrelated failures from missing optional `s3fs`/`obstore` deps).

AidanJanney:
@jsschreck I can keep boundary effect changes/suggestions to the credit-mesaclip repo for the time being. I also want to continue increasing model dim size to see how that affects these boundary artifacts with such a large dataset, but the propagation of perturbations along face boundaries is making me think it's weirdness with boundaries that could be improved - not just model size/undertraining.

ILoveSorasakiHina:
@jsschreck 

Thanks for the fixes! I reviewed the updated code, and the fixes for the first two issues look good. The relevant CPU tests also passed. The GPUs on CWA’s HPC system are currently busy, so I haven’t rerun the updated version on GPUs yet. I’m happy to leave the corner attention updates for the follow-up boundary work.

Regarding the boundary artifacts you and @AidanJanney discussed, I noticed that when `se_to_cube()` in `cubed_wxformer.py` arranges the SE data into six faces, some shared boundary locations remain zero, and `HaloExchange` in `halo.py` does not fill them in. Could these locations be filled using the corresponding data from the neighboring faces?

In a CPU test with an all-ones SE input, 4,324 boundary locations remained zero after these operations. I’m not sure whether this is related to the artifacts Aidan observed, and I haven’t compared the behavior of the full model with and without filling these locations. I just wanted to suggest a possible direction to investigate.

ILoveSorasakiHina:
@jsschreck 

Thanks for the fixes. I re-ran training on the updated branch (`4898ffd`) for both models, and training, validation, and checkpoint saving completed without problems in each case.

From my side the review items are addressed and I don't see anything else blocking.

djgagne (commented):
Thanks @jsschreck, and thanks @ILoveSorasakiHina for the very thorough testing. I went through this against the PR review checklist and also checked whether the zero-edge issue from 09-17 is resolved. It isn't yet; details and a suggested fix are below.

## Zero-valued edge cells in `CubedWXFormer`: not fixed yet

The PR head (`4898ffd7`) came before the 09-17 report, and none of the code involved has changed since.

**Why it happens:**
- `se_to_cube` (`cubed_wxformer.py:604`) creates the cube with `new_zeros` and fills only SE-owned cells. The 6×361² − 777,602 = **4,324** duplicate edge cells therefore start as zero.
- `HaloExchange.forward` (`halo.py:393`) does `torch.where(self.native_mask, x_native, interpolated)`.
- `native_mask` covers the whole 361×361 face (`halo.py:142-143`), so every cell inside the face, zeros included, is copied through unchanged. Only cells outside the face get ghost values.
- The docstring at `halo.py:18` says the halo "fills duplicated face edges, missing native face cells". Only the legacy `source_flat_index` map does that, and `forward()` doesn't use it.

**Repro:** I built a synthetic ne120 `se_index` with the same dominant-axis ownership rule `halo.py` uses. It gives exactly ncol = 777,602. I then ran an all-ones input through `se_to_cube` and `_pad_face` with halo exchange on:

| | zero cells |
|---|---|
| after `se_to_cube` | 4,324 |
| after `HaloExchange`, inside the face | **4,324** |
| faces 0 and 1 | 0 |
| faces 2 and 3 | 722 each (left and right edge columns) |
| faces 4 and 5 (poles) | 1,440 each (all four edges) |

The encoder therefore sees one-pixel lines of zeros along the seams on four of the six faces. That's a plausible contributor to the artifacts along face boundaries @AidanJanney described.

**Suggested fix:** only copy through cells the SE grid actually owns, and let the rest take the halo value:

```python
# per-face mask, shape (6, 1, p, p); forward() needs a small reshape to use it with faces packed in the batch dim
native_mask[:, :, crop_top:crop_top+E, crop_left:crop_left+E] = owned.reshape(NFACE, 1, E, E)
```

I checked this in memory with a smooth test field (the unit-sphere xyz coordinates):
- All 4,324 unowned cells get exactly the right value (max abs error 0.0).
- Every owned cell is bit-identical to before.

Without an adjacency file there's no halo, so these zeros remain. That fallback could scatter-fill the duplicate cells from their owning face instead.

Please also add a CI test with a small synthetic cube index, so halo exchange, edge attention and cross-face attention are actually tested. At the moment every `CubedWXFormer` test uses `adjacency_path="/does/not/exist.npz"`, and the one ne120 halo test is skipped on CI.

## Required checklist

| Item | Status |
|---|---|
| Clear description | ✅ Thorough summary, verification and follow-ups |
| Issues tagged | ❌ No linked issues |
| CI passes | ✅ build, py3.11/3.12/3.13, RTD and pre-commit all pass |
| Tests updated | ⚠️ New tests pass (103 passed, 1 skipped locally), but CI never tests halo exchange, edge attention or cross-face attention (see above) |
| Docs updated | ❌ See below |
| Dependencies | ✅ Nothing new needed |
| Generic paths | ❌ See below |
| Positive and constructive feedback | ✅ This review |
| GPU run on Casper/Derecho | ✅ Author ran Derecho smoke tests; ILoveSorasakiHina ran 8-GPU DDP and re-ran training at `4898ffd` |

**Docs:**
- No page under `docs/source/` changed.
- `docs/source/Models_gen2.md:56,98` still documents `nextgen_wxformer` / `wxformer_next.NextGenWXFormer`, which no longer exist, so that autoapi link will break.
- `cubed_wxformer`, `wxformer_column`, the `tripole_to_se` preblock and the `se_to_latlon` postblock need entries in `Models_gen2.md`, `Preblocks.md` and `postblocks_gen2.md`.
- `wxformer_family.md` and the README aren't in the Sphinx toctree.
- The `nextgen_wxformer` key was removed with no alias, so any existing config using it breaks. Please mention that in the docs or changelog.

**Paths:** these hardcode `/glade/work/schreck/...` or `/glade/derecho/scratch/schreck`:
- `config/gen_2/examples/wxformer_cubesphere_next_wb2.yml:40,41,96,130`
- `config/gen_2/smoke/smoke_cubed_wxformer_derecho.yml:1,38,39,90`
- `config/gen_2/smoke/smoke_wxformer_column_derecho.yml:1`
- `credit/models/wxformer/cubed_wxformer.py:115,746`
- `credit/preblock/latlon_to_se.py:11,19`

The ne120 static files (`se_index`, adjacency) and the regrid weights should move to `/glade/campaign/aiml/credit/`, and the configs should point there.

## Other issues

1. **Checkpoint loading fails in two rollout modes.**
   - `CubedWXFormer` subclasses `nn.Module`, not `BaseModel` (`cubed_wxformer.py:221`).
   - `load_model(conf, load_weights=True)` calls `model.load_model(conf)`, which only `BaseModel` has.
   - So `predict.mode: none` and `fsdp` in `rollout_to_netcdf_gen2.py:331,341` and `rollout_realtime_gen2.py:296,306` will raise AttributeError. Only DDP mode, the path that was tested, loads the state dict directly.
   - `WXFormerColumn` already subclasses `BaseModel`; `CubedWXFormer` should too.
2. **The example config fails `credit check`.** `wxformer_cubesphere_next_wb2.yml` has no `trainer.parallelism` block, and `validation_data` is missing `forecast_len` and `timestep`. Both smoke configs pass.
3. **"Resolution-agnostic" only holds with adjacency off.**
   - `halo.py`, `edge_attn.py` and `global_attn.py` hardcode `NFACE_EDGE = 361`.
   - `HaloExchange.forward` raises for any other face size.
   - The auto-detect in `cubed_wxformer.py:376` looks only for `se_face_adjacency_ne120.npz`.

   Either derive E from the index, as `cubed_wxformer.py` does, or document that the boundary features are ne120-only.
4. **A missing adjacency file only produces a warning.** A mistyped `adjacency_path` logs a warning and quietly turns off halo and edge attention (`cubed_wxformer.py:378-383`); I'd make it an error. `HaloExchange` loads that file only to confirm it exists and never uses it (`halo.py:116`).
5. **Shared postblock and preblock changes are broader than they need to be.**
   - `Reconstruct` now flattens only when `shape[2] == 1`, so a 5D output with T > 1 is no longer flattened (the old check was `dim() == 5`).
   - `FlattenToTensor` now flattens every tensor.
   - `norm.py` skips normalization for any 5D `(..., 1, N)` tensor.

   These are fine for today's models, but they guess from tensor shape in shared code. Please add a comment or test pinning the intended behavior.
6. **Corner overwrite (item 3 in @ILoveSorasakiHina's review).** It was deferred to follow-up work. Please open a tracking issue so it isn't lost.

## Recommended checklist / minor

- **Docstrings:** good overall.
  - `FaceTransformerBlock.forward` has none.
  - The `SEToLatLonPostBlock` docstring says it runs once after the rollout (`per_step=False`), but nothing reads that attribute and the example config puts it under `per_step`.
- **Type annotations:** mostly present. `cube_rope.py` leaves some `device`/`dtype` parameters unannotated.
- **Redundancy:** the `TOPOLOGY` table and the `NFACE`/`NFACE_EDGE` constants are copied between `edge_attn.py` and `global_attn.py`, and should live in one shared module. The `__main__` smoke block in `cubed_wxformer.py` modifies `sys.path` and belongs in a script or manual test.
- **Comments:** excellent. The decoder-stride rationale and the residual explanation are especially clear.

## What's great

- Merging the two cube-sphere classes into one, with `torch.equal` parity checked against both. That's a nice reduction in duplicated code.
- The ghost-cell geometry, where each halo cell is traced back to the SE cell that really owns it rather than using lat/lon-style padding.
- Quick, well-tested fixes for the residual and `Reconstruct` issues, including the parametrized input-only/output-only tests.

## Before merging

1. The edge-zeros mask fix, plus a synthetic-index halo test on CI.
2. Generic data paths, with the static files and weights in `/glade/campaign/aiml/credit/`.
3. Updated docs, including removing `nextgen_wxformer` from `Models_gen2.md`.
4. A fix for checkpoint loading: subclass `BaseModel`, or document that only DDP rollout works.

AidanJanney:
@djgagne I talked with @jsschreck this morning and have a PR incoming today. 

It addresses: 
- The halo issue you mention above originating from `se_to_cube`
- Part of Additional issue 1 - I don't think the current code saves out checkpoints for cross-face attention which contributes to this.

I can try and take a pass at these other issues too.

PR #479, opened 2026-08-17
Add register_trainer: custom_objects support for trainers

jsschreck:
## Summary

This is here to support Kyle's regional trainer. His WRF/downscaling model needs a training loop that's different enough from the standard gen2 trainer (it takes interior + boundary + time-encoding tensors, not a single `x`) that it doesn't make sense to bolt onto `trainer_gen2.py` directly. The cleaner path is a new trainer class that subclasses `TrainerERA5Gen2` and overrides just what it needs, living in credit-zoo rather than in CREDIT itself. But there was no way to actually plug a trainer like that into a config without hand-editing `credit/trainers/__init__.py`, so this PR adds that missing piece first.

Trainers were the one part of the registry system you couldn't extend without editing CREDIT's source directly. Models, datasets, preblocks, postblocks, losses, and metrics all support a `custom_objects:` block in the config, so you can write your own class, point a config at it, and it behaves exactly like a built-in. Trainers never got that. `_TRAINER_REGISTRY` in `credit/trainers/__init__.py` was just a closed dict, and the only way to add one was to hand-edit that file.

This PR closes that gap. We added `register_trainer`, which works the same way `register_model` already does: decorate a class, it gets checked against `BaseTrainer`, and it lands in the registry. We also added `"trainer"` to `registry.py`'s `_REGISTRY_MAP`, so `custom_objects: {..., object_type: trainer}` works right away, no other wiring needed.

Worth noting: `credit check` already runs `load_custom_objects` before it does any registry lookups, so it picks up custom trainers for free. Nothing to change there.

This doesn't touch `trainer_gen2.py` or any trainer's actual training logic. It's purely the registration/dispatch layer, the same file where every built-in trainer already lives. Docs are updated too (`CLAUDE.md`, `docs/source/Custom.md`), since we're now at seven registrable types instead of six.

## Verification

Tested the decorator path, the `custom_objects` path (the one most people will actually use), and confirmed a wrong base class raises `TypeError` like it should. Built-in trainers still resolve fine. `test_trainer_components.py` passes 89/89, and the full suite passes clean: 1602 passed, 33 skipped, 0 failed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,9 +94,11 @@ Seven subsystems — models, trainers, datasets, preblocks, postblocks, losses, 
 by a string key in the config (`model.type`, `trainer.type`, `data.source.<name>.dataset_type`, etc.),
 resolved through per-package `_REGISTRY` dicts with lazy imports:
 - `credit/models/__init__.py` → `load_model(conf)`, keys like `crossformer`, `wxformer`, `unet`, `fuxi`, `swin`, `graph`
-- `credit/trainers/__init__.py` → keys like `era5-gen1`/`era5`, `gen2`/`era5-gen2`, `era5-diffusion`, `era5-ensemble`, `ic-opt`
-- `credit/datasets/`, `credit/preblock/`, `credit/postblock/`, `credit/losses/`, `credit/metrics/` follow the
-  same `register_*` decorator pattern
+- `credit/trainers/__init__.py` → `load_trainer(conf)`, keys like `era5-gen1`/`era5`, `gen2`/`era5-gen2`,
+  `era5-diffusion`, `era5-ensemble`, `ic-opt`
+- `credit/datasets/`, `credit/preblock/`, `credit/postblock/`, `credit/losses/`, `credit/metrics/`, `credit/trainers/`
+  all follow the same `register_*` decorator pattern (`register_trainer` mirrors `register_model`, requiring the
+  class subclass `credit.trainers.base_trainer.BaseTrainer`)
 
 `credit/registry.py` (`load_custom_objects`) is the meta-layer: a config's `custom_objects:` block lets users
 plug in their *own* classes (must subclass the relevant `Base*` class) by dotted import path, without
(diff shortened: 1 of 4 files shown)


djgagne:
I do see the value of having a trainer registry since we do have a base trainer class and there may be optimizations to be had by customizing the training loop. That said, I do think Kyle's regional model could be implemented with the existing training by taking advantage of the state dict and multi source dataset by composing some of the operations in pre/post blocks and the forward method of the regional model.

kevinyang-cky:
> I do see the value of having a trainer registry since we do have a base trainer class and there may be optimizations to be had by customizing the training loop. That said, I do think Kyle's regional model could be implemented with the existing training by taking advantage of the state dict and multi source dataset by composing some of the operations in pre/post blocks and the forward method of the regional model.

I second this and would suggest adding guidance in the custom registry docs stating that one should first explore ways to use existing/custom datasets or pre/postblocks objects before creating a custom trainer, to avoid introducing a completely new pipeline (in some cases it is fine) that deviates completely from the Gen2 structure.

jsschreck:
I can stage the PR so that regional work is contained with trainer_gen2.py ... I did this initially. My hesitation surrounding Kyle's specific model is due to FiLM. That will require some changes a little more broadly. If we ignored adding FiLM then its pretty straight forward.

jsschreck:
Kevin, I will write some details into the documentation about ways to use the existing infrastructure with some tips before one registers an alternative trainer. I will get the backwards optimization example registered eventually (for that piece of work, I do a separate PR later after I talk to Greg in more detail). I have a "coupled" trainer that is not in CREDIT yet which is a sketch of how to train a bunch of models together. I am not sure this is GEN2 though :D might be GEN3 :D :D)

kevinyang-cky:
> Kevin I will write some details into the documentation about ways to use the existing infrastructure with some tips before one registers an alternative trainer. I will get the backwards optimization example registered eventually (for that piece of work, I need a separate PR that, but will not do until later until after I talk to Greg in more detail).

Sounds good! Thank you!

PR #480, opened 2026-08-21
feat: gen2 GlobalNetEnergyFixer postblock for net TOA/surface fluxes

djgagne:
## Summary

Closes #467.

- **`GlobalNetEnergyFixer`** (`credit/postblock/conservation.py`): gen2, name-based port of the gen1 `GlobalEnergyFixer`. Conserves global total energy using **net** TOA/surface fluxes (ERA5 `top_net_*`, `surface_net_*`, `surface_*_heat_flux`), with temperature carrying the correction. ERA5 positive-downward sign convention: `R_T = toa_net_solar + toa_net_thermal`, `F_S = surf_net_solar + surf_net_thermal + SH + LH`.
  - `flux_units: "J m-2"` (default, gen1 semantics — accumulated over the step) or `"W m-2"` (mean rate).
  - `sp_var` optional: required for `grid_type: sigma`, ignored on the pressure-level grid (supported, as in gen1).
  - Out-of-place ops so the correction stays on the autograd graph.
- **Registry**: new key `global_energy_fixer_net`; `global_energy_fixer` remains an alias of `GlobalEnergyFixerUpDown` (no behavior change for existing configs). `credit check` validates the new block's args automatically via signature binding.
- **Docs**: `postblocks_gen2.md` now has a "which energy fixer" note plus separate `global_energy_fixer_net` (ERA5 example) and `global_energy_fixer_updown` sections; gen1 `postblock.md` points to the gen2 equivalents.
- **Tests**: run/grad, budget closure (`TE_t1 == TE_t0 + N_seconds·(R_T − F_S)` after the fix), J m⁻² ≡ W m⁻² equivalence, error cases, registry resolution.

This branch also carries three earlier commits not yet on `main`: the ARCO ERA5 `co/` data source + `spectral_to_grid` preblock (1a1c3598), the Casper smoke-test runbook (6d6cfbf5), and config updates (1d53fb73).

## Test plan

- [x] `pytest tests/test_conservation_gen2.py tests/test_preblock.py` — 87 passed
- [x] `pytest tests/test_postblock.py` (gen1 fixers untouched) — 29 passed
- [x] `_bind_error` from `credit check` accepts a correct `global_energy_fixer_net` arg set and reports a misspelled arg with the accepted list
- [x] `ruff check` / `ruff format` (pre-commit)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #481, opened 2026-08-21
feat: gen2 GlobalNetEnergyFixer postblock for net TOA/surface fluxes

### LINKED ISSUE #467: global_energy_fixer postblock (gen2) only supports explicit up/down fluxes — no net-flux option, and docs are gen1-only
### Summary

The gen1 global energy fixer postblock (credit/postblock/gen1.py::GlobalEnergyFixer) accepts net TOA and surface radiation fluxes (e.g. FSNT, FLNT, FSNS, FLNS), matching what's documented at https://miles-credit.readthedocs.io/en/latest/postblock.html. The gen2 implementation (credit/postblock/conservation.py::GlobalEnergyFixerUpDown) only supports explicit up/down flux decomposition (toa_up_solar_var, toa_up_olr_var, surf_down_solar_var, surf_up_solar_var, surf_down_lw_var, surf_up_lw_var, etc.) — there's no net-flux equivalent registered for gen2.

This means:
1. Users can't use net-flux output with the gen2 energy fixer without manually decomposing into up/down components first.
2. The registry key global_energy_fixer in credit/postblock/__init__.py is aliased to GlobalEnergyFixerUpDown, so a gen1-style config using that key (with net-flux variable names) will fail or silently misconfigure rather than clearly erroring

### Proposed fix
- Port a net-flux-compatible energy fixer (GlobalEnergyFixer, mirroring gen1's class) into credit/postblock/conservation.py for gen2, alongside the existing GlobalEnergyFixerUpDown.
- Register it under a distinct key (e.g. global_energy_fixer_net) so it doesn't collide with the up/down variant, and consider renaming/aliasing the existing global_energy_fixer key more explicitly

### Docs
The [postblock readthedocs page](https://miles-credit.readthedocs.io/en/latest/postblock.html) currently only shows the gen1/net-flux config example under "Global Energy Fixer." Once gen2 supports both methods, the docs should be updated to show both config examples (net-flux vs. up/down) and clarify which config keys map to which method for gen2.

### Environment
- Repo: NCAR/miles-credit, main branch
- Relevant files: credit/postblock/gen1.py, credit/postblock/conservation.py, credit/postblock/__init__.py

djgagne:
## Summary

Closes #467. (Supersedes #480, which also carried unrelated branch history.)

- **`GlobalNetEnergyFixer`** (`credit/postblock/conservation.py`): gen2, name-based port of the gen1 `GlobalEnergyFixer`. Conserves global total energy using **net** TOA/surface fluxes (ERA5 `top_net_*`, `surface_net_*`, `surface_*_heat_flux`), with temperature carrying the correction. ERA5 positive-downward sign convention: `R_T = toa_net_solar + toa_net_thermal`, `F_S = surf_net_solar + surf_net_thermal + SH + LH`.
  - `flux_units: "J m-2"` (default, gen1 semantics — accumulated over the step) or `"W m-2"` (mean rate).
  - `sp_var` optional: required for `grid_type: sigma`, ignored on the pressure-level grid (supported, as in gen1).
  - Out-of-place ops so the correction stays on the autograd graph.
- **Registry**: new key `global_energy_fixer_net`; `global_energy_fixer` remains an alias of `GlobalEnergyFixerUpDown` (no behavior change for existing configs). `credit check` validates the new block's args automatically via signature binding.
- **Docs**: `postblocks_gen2.md` now has a "which energy fixer" note plus separate `global_energy_fixer_net` (ERA5 example) and `global_energy_fixer_updown` sections; gen1 `postblock.md` points to the gen2 equivalents.
- **Tests**: run/grad, budget closure (`TE_t1 == TE_t0 + N_seconds·(R_T − F_S)` after the fix), J m⁻² ≡ W m⁻² equivalence, error cases, registry resolution.

## Test plan

- [x] `pytest tests/test_conservation_gen2.py tests/test_preblock.py` — 87 passed
- [x] `pytest tests/test_postblock.py` (gen1 fixers untouched) — 29 passed
- [x] `_bind_error` from `credit check` accepts a correct `global_energy_fixer_net` arg set and reports a misspelled arg with the accepted list
- [x] `ruff check` / `ruff format` (pre-commit)


docs/source/postblocks_gen2.md
```diff
@@ -347,15 +347,61 @@ args:
...
+Gen2 port of the gen1 `GlobalEnergyFixer`: the column total-energy tendency is
+forced to match the net TOA + surface energy fluxes, with temperature carrying
+the correction. All six fluxes are read from the prediction by name and follow
+the ERA5 sign convention (**positive downward**):
+`R_T = toa_net_solar + toa_net_thermal` and
+`F_S = surf_net_solar + surf_net_thermal + surf_sh + surf_lh`; the atmosphere
+gains `R_T - F_S`. Datasets with upward-positive fluxes (e.g. CESM `FLNT`,
+`FLNS`, `SHFLX`, `LHFLX`) must be sign-flipped first. `flux_units` is
+`"J m-2"` (default — energy accumulated over the step, divided by
+`lead_time_periods` internally) or `"W m-2"` (mean rate over the step).
```
kjmayer, line 369: suggest changing to "...or `"W m-2"` (instantaneous or mean rate over the step)."

credit/postblock/conservation.py
```diff
@@ -236,6 +236,180 @@ def forward(self, batch_dict: dict) -> dict:
...
+        F_S = surf_net_solar + surf_net_thermal + surf_sh + surf_lh        # net energy out of the atmosphere at the surface
+
+    and the atmosphere gains ``R_T - F_S`` per unit area.  Datasets that store
+    upward-positive fluxes (e.g. CESM ``FLNT``/``FLNS``/``SHFLX``/``LHFLX``)
+    must be sign-flipped before this block sees them.
+
+    Units: with ``flux_units="J m-2"`` (default, gen1 behavior) every flux is
+    the energy accumulated over the forecast step and is divided by
+    ``N_seconds`` internally; with ``flux_units="W m-2"`` every flux is the
+    mean rate over the step and is used as-is.
```
kjmayer, line 265: suggest adding "...every flux is the instantaneous or mean rate..."

credit/postblock/conservation.py
```diff
@@ -236,6 +236,180 @@ def forward(self, batch_dict: dict) -> dict:
...
+        surf_sh_var, surf_lh_var: surface sensible / latent heat flux keys.
+        lead_time_periods: forecast step length in hours (sets ``N_seconds``).
+        sp_var: surface pressure key; required for ``grid_type="sigma"``,
+            ignored for ``grid_type="pressure"``.
+        flux_units: ``"J m-2"`` or ``"W m-2"`` (see above).
+        input_source_key: batch-dict key holding the t0 physical input state.
+        **physics: physics-core setup keys (see ``GlobalMassFixer``).
+    """
+
+    _FLUX_UNITS = ("J m-2", "W m-2")
```
kjmayer, line 282: I wonder if you should include J/m2 and W/m2 as an option as well

tests/test_conservation_gen2.py
```diff
@@ -337,6 +343,119 @@ def test_energy_fixer_updown_runs_and_grad(statics_file):
...
+    src = batch["y_processed"][SRC]
+    inp = batch["x_physical"][SRC]
+
+    def lvl(d, k, t):
+        return d[k][:, :, t, ...].detach()
+
+    def sfc(d, k, t):
+        return d[k][:, 0, t, ...].detach()
+
+    T0, q0, U0, V0 = (lvl(inp, key("prognostic", "3d", n), -1) for n in ("T", "Qtot", "U", "V"))
```
kjmayer, line 382: I know Qtot is slightly different than Q, but I wonder if Q is still a more universal option here.

kjmayer (approved):
I added minor suggestions.

credit/postblock/conservation.py
```diff
@@ -236,6 +236,180 @@ def forward(self, batch_dict: dict) -> dict:
...
+        global_TE_t0 = self.core.weighted_sum(TE_t0, axis=(-2, -1))
+        global_TE_t1 = self.core.weighted_sum(TE_t1, axis=(-2, -1))
+
+        # total energy correction ratio; broadcast to (B, 1, 1, 1)
+        E_correct_ratio = (self.N_seconds * (R_T_sum - F_S_sum) + global_TE_t0) / global_TE_t1
+        E_correct_ratio = E_correct_ratio.unsqueeze(-1).unsqueeze(-1).unsqueeze(-1)
+
+        # let thermal energy carry the corrected total energy amount
+        E_t1_correct = E_level_t1 * E_correct_ratio
+        T_pred = (E_t1_correct - E_qgk_t1) / CP_t1
```
charlie-becker, line 406: I think this code calculates the ratio based on all variables (including geopotential and kinetic winds), which I think would disproportionally add more energy (temperature) to high elevations, for example. 

We should be able to create the ratio to distribute uniformly based on mass or scaled by the raw temperature only (I believe), but I don't know which one makes more sense.

charlie-becker (changes requested):

PR #484, opened 2026-09-02
Add SDL fine-tuning config + fix stale Ensembles.md docs

jsschreck:
## Summary

This adds a real, checked config for fine-tuning SDL (the noise-injection ensemble layers, `StochasticDecompositionLayer`) on top of a trained wxformer checkpoint, plus a docs pass to go with it.

While writing the config I found `docs/source/Ensembles.md` already existed and already tried to document this, but it was out of date with the actual code. It described a `PixelNoiseInjection` class that doesn't exist (the real class is `StochasticDecompositionLayer`) and a gen1-style `trainer.type`/`KCRPS` config example that doesn't match how any of this is actually wired up today. Rather than add a second, contradicting page, I rewrote that section in place: what's actually learned per injection point versus what's a fixed hyperparameter (the `noise_factor` lives in an `nn.Parameter` but is built with `requires_grad=False`, so it doesn't train despite looking learnable), and a concrete numbered walkthrough instead of an architecture overview with no runnable steps. Also cleaned up a leftover unfinished note at the bottom of the file into a proper section on the two different ways to build an ensemble in CREDIT (perturbed IC vs. fixed IC/SDL), since that distinction was worth keeping, just not as a stray fragment.

New config: `config/gen_2/examples/wxformer_sdl.yml`. Same data/preblocks/postblocks/loss as the base gen2 starter config, `model.type: crossformer-ensemble` with the SDL params added on top of the frozen base architecture, and `trainer.load_weights: True` pointing at a copied-in base checkpoint (there's no separate "warm start from elsewhere" field, so the config comments walk through copying the checkpoint into the new `save_loc` first).

## Verification

Ran `credit check` (plain and `--deep`) against the new config, both clean. Also did a direct test outside the config path: built a plain `CrossFormer`, saved its state dict, loaded it into a `CrossFormerWithNoise` with `strict=False`, and confirmed the only missing keys are the new noise layers (zero unexpected keys), that only the noise-layer parameters have `requires_grad=True`, and that a forward/backward pass produces a nonzero gradient on those parameters. That's the exact mechanism the docs describe, not just a description of it.


docs/source/Ensembles.md
```diff
@@ -11,97 +11,83 @@ CREDIT supports two primary training approaches for ensemble generation:
 
 The fine-tuning approach is currently preferred due to computational efficiency and resource requirements.
```
kevinyang-cky: Suggest highlighting this as:

> The fine-tuning approach is currently preferred due to computational efficiency and resource requirements.

kevinyang-cky, line 12: Suggest highlighting as:
> The **fine-tuning approach** is currently preferred due to computational efficiency and resource requirements.

docs/source/Ensembles.md
kevinyang-cky, line 9: Need some clarification: in [config/gen_2/examples/wxformer_sdl.yml](https://github.com/NCAR/miles-credit/blob/main/config/gen_2/examples/wxformer_sdl.yml), it is just using MSE.

docs/source/Ensembles.md
```diff
@@ -11,97 +11,83 @@ CREDIT supports two primary training approaches for ensemble generation:
...
-- Noise factors are learnable parameters that adapt during training
-- Separate noise factors for encoder and decoder stages
+   ```bash
+   mkdir -p /glade/derecho/scratch/$USER/CREDIT_runs/wxformer_sdl
+   cp /glade/derecho/scratch/$USER/CREDIT_runs/starter_gen2/checkpoint.pt \
+      /glade/derecho/scratch/$USER/CREDIT_runs/wxformer_sdl/checkpoint.pt
+   ```
 
-#### Scaling Strategies
+2. **Copy the base model's architecture into the SDL config exactly** (`image_height`/`image_width`, `levels`, `channels`, `surface_channels`, `dim`/`depth`, window sizes, `padding_conf`, everything under `model:` the base config had). SDL loads the checkpoint by matching parameter names and shapes; a mismatch either fails to load or loads weights into the wrong place.
```
kevinyang-cky, line 44: Suggest removing this part and just state that the SDL config should copy everything in the config that trained the deterministic model, plus add SDL-related parameters, change the model type, and turn `load_weights` on.

kevinyang-cky (commented):
Thanks for adding an example config and updating the Docs. I thought the Docs were very easy to follow for setting up the SDL training (I **tested with my already-trained satellite nowcasting model, and it ran fine**). Several general suggestions:

- Which example config is `wxformer_sdl.yml` inherited from?
- Add a CRPS-related example/instruction?
- For the inference part in the Docs, more explicit? e.g., which parameter to set to what.
- I would suggest including an example config for the diffusion-based ensemble part, or maybe remove it from the Docs completely (IMO: unclear to the user how to set it up just from reading the Docs,s and I assume SDL should be the go-to for now)

